PR #30, opened 2015-01-20
Speed boot time and enable GRO.

iuliur:


YanVugenfirer:
Hi Julius,

Can you split the two patches please?

We intentionally removed  RSC (GRO) support from the driver, because currently we cannot pass WHQL with it. So if we will add it, then either as a flag during build or only to master and not to stable branch.

Thanks.

YanVugenfirer:
For now we can add GRO (RSC) support only to master and not to stable. 
The other way around is to enable the feature according to build parameter.

Best regards,
Yan.

YanVugenfirer:
* RSC (GRO) is enabled by default in current drivers
* I prefer to keep the current memory allocation scheme. In HCK scenario driver can be unloaded several times and memory will be fragmented. Allocation of large chunks might not be possible.

PR #35, opened 2015-05-13
Prevent inadvertent set of wce guest feature

asadpanda:
This change prevents VIRTIO_BLK_F_WCACHE guest feature from being set when VIRTIO_BLK_F_SEG_MAX was supposed to be set.

This was causing migrates to fail on RHEL 6.6 with error.
Features 0x20000250 unsupported. Allowed features: 0x71000454

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -538,7 +538,7 @@ VirtIoHwInitialize(
     }
 
     if (CHECKBIT(adaptExt->features, VIRTIO_BLK_F_SEG_MAX)) {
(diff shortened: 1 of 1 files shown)


vrozenfe:
Thank you,
I will apply your patch shortly.
Best regards,
Vadim.

jnpkrn:
Note that this PR can be closed:
https://github.com/YanVugenfirer/kvm-guest-drivers-windows/commit/c6c0158

P.S. I came around this when looking at
http://oss.clusterlabs.org/pipermail/users/2016-October/004209.html

Thanks for getting this addressed.

PR #38, opened 2015-07-08
https://chromium.googlesource.com/external/webrtc

freewaf:


YanVugenfirer:
Sorry for late reply. I was on vacation.

Can you rebase and send only the new patches?

Thanks,
Yan.

PR #49, opened 2016-01-02
balloon: fix memory statistics reporting

locker:
According to libvirt API reference, SWAP_IN, SWAP_OUT, MAJOR_FAULT, and
MINOR_FAULT must be reported as total numbers, not per second, as they
are now. Besides, SWAP_IN and SWAP_OUT are supposed to be in bytes, not
in pages. Virtio balloon driver for Linux follows these conventions,
while balloon service for Windows does not, which makes it inconvenient
to use these statistics when Windows and Linux VMs are running on the
same host.

This patch fixes this issue.

Signed-off-by: Vladimir Davydov <vdavydov@virtuozzo.com>

### DIFF
--- a/Balloon/app/memstat.cpp
+++ b/Balloon/app/memstat.cpp
@@ -93,6 +93,7 @@ BOOL CMemStat::Init()
 
 BOOL CMemStat::Update()
 {
+    SYSTEM_INFO sysinfo;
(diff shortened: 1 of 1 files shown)


vrozenfe:
Applied.
Thank you,
Vadim.

PR #51, opened 2016-01-16
In the ParaNdis_DeviceFiltersUpdateAddresses,SendControlMessage   arg…

grass-lu:
… 7 is not right, it should be size that valid MACS occupied size plus MAC number of field occupation space length


grass-lu:
Windows xp virtual machine  which use virtio net driver cannot receive Multicast network packet 。Qemu -kvm receive_filter function drop Multicast network packet,because VirtIONet  mac_table.first_multi all is zero.IN  Netkvm driver ParaNdis_DeviceFiltersUpdateAddresses ,SendControlMessage arg  7 is not right, it should be size that valid MACS occupied size plus MAC number of field occupation space length.

YanVugenfirer:
Thanks!

YanVugenfirer:
Fixed: https://github.com/YanVugenfirer/kvm-guest-drivers-windows/commit/78ca9c6551bdc468325dcaebcd947ae9b25e462a

grass-lu:
ok ，thanks！

PR #53, opened 2016-02-02
Balloon: gather stats in kernel-mode

dlunev:
This set of patches gathers memory statistics from kernel-mode counters,
thus eliminating the need of user-mode service. It is ugly and excessive.
Moreover some of kernel-mode counters are not available in userspace.
Also relying on WMI counter description is not enough to be sure what
counters measures really, and this is critical to successive ballooning.

Patch [1/4] adds undocumented, but reasonably stable API call and its structures
and proves that obtained kernel mode counters are identical to those from WMI.

Patch [2/4] replaces file queue for app service with workitem and ensures
that every ISR from host will had corresponded virtio push.

Patch [3/4] removes obsolete app service.

Patch [4/4] fixes possible counter overflows due large guest uptime.

Alexey V. Kostyushko (4):
  Balloon: gather stats in kernel-mode
  Balloon: replace filequeue with workitem
  Balloon: user-mode service app removed
  Balloon: counters overflow fix


YanVugenfirer:
Thanks, going to review the patches.

vrozenfe:
Using Native API was the first thing that came to my mind when I started working on memory statistics.
Unfortunately, most of that stuff is not documented, even though that it is good known. 
Thank you guys  for your efforts, but  we will not accept undocumented code in out production drivers.

Hope you understand our concern.

Best regards,
Vadim Rozenfeld.

dlunev:
From our point of view this stuff is not documented for public and is documented for HAL vendors. This API in our experience is really stable and working. These counters are reliable source of the information for all present released Windows systems.

Moreover, keeping this code in guest is problematic. If we are using this statistics for balloon management, we are in trouble. We should react for overloaded guest and this reaction is happens on balloon statistics which could be not available or be obsolete. This thing is really important.

dlunev:
Vadim,

am I correct that production drivers are based on a different repo? These ones are community drivers while commercial ones are based on https://github.com/vrozenfe/virtio-win ?

If the code will be accepted for community drivers than this would be OK. The stability of them will be known with time and we will be able to re-consider this question later.

Does this sounds reasonable to you?

Den

vrozenfe:
No https://github.com/vrozenfe/virtio-win is a kind of my public repository, which I use for a quick code 
sharing with other people. Technically our internal repository is the production repository at the same time. It is not accessible from outside of RH. The repository maintained by Yan is a clone of our internal repository. As you can see, we have two main branches - master and stable. We use stable branch for building our production drivers. What we can do is to accept you patches in master
 branch, but keep them away  from stable (at least until MS will start treating us a HAL vendor and start signing virtio-win drivers without asking for WHQL results :) ).

Best regards,
Vadim.

dlunev:
that would be ABSOLUTELY fine to me. Being mainstream is enough for now. We will ship our own version of the drivers based on this.

Speaking about WHQL I am not sure that this API is a stopper. As far as I know the test does not check the API used. Do you really think that this contradicts WHQL?

dlunev:
ladipro,

thank you very much for this review. We will definitely fix this in next submission.

vrozenfe:
Of cause not, they do check API usage for some stacks (miniport drivers) , but definitely not for generic drivers. But you newer know what will happen in next WHQL/HCK release .

dlunev:
should I open new pull request for changes or this forced push is good enough?

vrozenfe:
You can open a pull request, but push will work as well. Just give us a day a too for reviewing, befor merging to master branch.

Thanks,
Vadim.

dlunev:
ping

YanVugenfirer:
In master: https://github.com/YanVugenfirer/kvm-guest-drivers-windows/commits/master

Thanks!

PR #54, opened 2016-02-11
Balloon: some tweaks for guest statistics

dlunev:
Fix values for memory stats:
    * SWAP_OUT counts as Dirty+Mapped to match with SWAP_IN
    * improve MINFLT stat (as SoftFaults = CopyOnWriteCount + TransitionCount +
                             DemandZeroCount + CacheTransitionCount)
    * improve MAJFLT stat

Some additional information, which may be usefull for Windows guest ballooning:
*https://blogs.technet.microsoft.com/markrussinovich/2009/07/05/pushing-the-limits-of-windows-processes-and-threads/
newly MmResidentAvailablePages is a critical memory counter, its low values prevents apps to run
*MmSharedCommit value is the only source about memory-mapped size
*Filecache stats are vital on Windows for good ballooning, especially Cc...Miss ones.

There is no clean 'pagefile read' counter on Windows.
SWAP_IN uses PageReadCount counter, which includes reads from pagefile,
reads from memory-mapped files (both cached and non-cached) and
prefetching reads (including superfetch).

So SWAP_OUT should include mapped writes as well as dirty writes.
Then these two counters are 'matched by sources'.

This is not mention at MSDN, so here is a brief research report
as a proof (w2k12r2):
PageReadIoCout/PageReadCount are updated in MiWaitForInPageComplete,
which is called both from MiIssueHardFault and MiPfCompletePrefetchIos.
Last one is called from MiPrefetchVirtualMemory,
MmWaitForCacheManagerPrefetch, MiPrefetchControlArea and MmPrefetchPages.

PageFaultCount is incremented in MiAllocateWsle,
i.e. every time a physical page is mapped to a virtual address, using
working set. Actually this is invplg count, not #PF count.

Following formulas show what we can count:
* SoftFaults = CopyOnWriteCount + TransitionCount +
               DemandZeroCount + CacheTransitionCount
* PageReadCount = HardFaults + Prefetch, and we cannot separate them. But we
    can point, that without prefetching most of theese pages will trigger hard
    fault (based on assumption that prefetching doesn't make too much
    unnessessary reads)
* PageFaultCount - SoftFaults = PageReadCount + WsleAllocations.
    at least WsleAllocations includes some PageTables allocations
    and some kernel stack allocations.

That's why we should use PageReadCount as MAJFLT, not PageReadIoCount.
First value can be compared with similar fault counters, while the second
measures number of times the disk was read to resolve hard page faults and
during prefetching. Due to async processing and coalescing this value
doesn't correspond with anything.

Alexey V. Kostyushko (2):
  Balloon: force SWAP_OUT to match SWAP_IN
  Balloon: improve MINFLT and MAJFLT stats values

 Balloon/sys/memstat.c | 16 ++++++++++++----
 1 file changed, 12 insertions(+), 4 deletions(-)

### DIFF
--- a/Balloon/sys/memstat.c
+++ b/Balloon/sys/memstat.c
@@ -54,6 +54,7 @@ UpdateOverflowFreeCounter(PULARGE_INTEGER ofc, ULONG value)
 typedef enum _ULONG_COUNTER {
     _PageReadCount = 0,
     _DirtyPagesWriteCount,
+    _MappedPagesWriteCount,
     _PageReadIoCount,
     _PageFaultCount,
     _LastCounter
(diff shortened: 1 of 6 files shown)


dlunev:
ladipro, thank you. Pushed cleaned version.

dlunev:
guys, any opinion on this except nits?

vrozenfe:
Merged to master.
https://github.com/YanVugenfirer/kvm-guest-drivers-windows/commit/b078c39852e01d7fa6cad9281583790cbbfa3e3d

Thanks,
Vadim.

dlunev:
pls ignore this. This commit comes with a different case. I'll open new pull request for it.

PR #66, opened 2016-05-26
balloon: fix crash on balloon stop/unload

dlunev:
StatWorkItem flush should be perform after interrupts are disabled,
but before deletion of virtio queues in BalloonTerm

Signed-off-by: Alexey V. Kostyushko <aleksko@virtuozzo.com>


ladipro:
Looks good!

vrozenfe:
applied on master.
Thanks,
Vadim.

PR #69, opened 2016-06-27
Balloon: Dpc/D0Entry/D0Exit sync for StatWorkItem

dlunev:
Previous patch moves WdfWorkItemFlush(devCtx->StatWorkItem) calling
BalloonEvtDeviceD0Exit because StatWorkItem flush should be
performed with interrupts disabled, but before calling BalloonTerm which
will delete virtio queues.

Now PNP can sequentually call BalloonEvtDeviceD0Exit and
BalloonEvtDeviceD0Entry thus we should NOT set StatWorkItem to NULL
otherwither the guest will BSOD after the first DPC. PNP performs this
sequence during cpu hotplug.

There is also a race between BalloonInterruptDpc and
BalloonEvtDeviceD0Exit:

BalloonEvtDeviceD0Exit()
    if (devCtx->StatWorkItem)
    {
        WdfWorkItemFlush(devCtx->StatWorkItem);
    }

```
BalloonTerm(Device)
```

-->

BalloonInterruptDpc()
    if (1==InterlockedIncrement(&devCtx->WorkCount))
    {
-->     WdfWorkItemEnqueue(devCtx->StatWorkItem);
    }

Nothing blocks dpc on another processor to enqueue new workitem after workitem
queue was flushed and virtio queue was deleted. The patch adds the flag
(bD0Entry) to indicate that the driver is in D0 state. Dpc enqueues workitem
only if the flag is on. Set/clear flag and flag check / enqueue workitem
operations require spinlock synchronization.

Signed-off-by: Alexey V. Kostyushko aleksko@virtuozzo.com
Signed-off-by: Denis V. Lunev den@openvz.org

### DIFF
--- a/Balloon/sys/Device.c
+++ b/Balloon/sys/Device.c
@@ -514,7 +514,14 @@ BalloonInterruptIsr(
 
     if(VirtIODeviceISR(&devCtx->VDevice) > 0)
     {
-        WdfInterruptQueueDpcForIsr( WdfInterrupt );
+        /*
+         * According to MSDN WdfInterruptQueueDpcForIsr
+         * returns TRUE if dpc was inserted and FALSE otherwise
+         */
+        if (!WdfInterruptQueueDpcForIsr( WdfInterrupt ))
+        {
+            InterlockedIncrement(&devCtx->ExtraIsrCount);
+        }
         return TRUE;
     }
     return FALSE;
@@ -548,6 +555,7 @@ BalloonInterruptDpc(
         KeSetEvent (&devCtx->HostAckEvent, IO_NO_INCREMENT, FALSE);
     }
 
+    InterlockedIncrement(&devCtx->ExtraDpcCount);
     if (devCtx->StatVirtQueue &&
         virtqueue_get_buf(devCtx->StatVirtQueue, &len))
     {
@@ -568,6 +576,7 @@ BalloonInterruptDpc(
         WdfSpinLockAcquire(devCtx->D0Lock);
         if (devCtx->bD0Entry)
         {
+            InterlockedDecrement(&devCtx->ExtraDpcCount);
             if (1==InterlockedIncrement(&devCtx->WorkCount))
             {
                 WdfWorkItemEnqueue(devCtx->StatWorkItem);
--- a/Balloon/sys/ProtoTypes.h
+++ b/Balloon/sys/ProtoTypes.h
@@ -48,6 +48,8 @@ typedef struct _DEVICE_CONTEXT {
     WDFINTERRUPT            WdfInterrupt;
     WDFWORKITEM             StatWorkItem;
     LONG                    WorkCount;
+    LONG                    ExtraIsrCount;
+    LONG                    ExtraDpcCount;
     PUCHAR                  PortBase;
     ULONG                   PortCount;
     BOOLEAN                 PortMapped;


ladipro:
Thanks for the patches!
It looks like BalloonInterruptDpc accesses the virtqueues without any synchronization. A pre-existing bug that should also be fixed. So I'm wondering if it wouldn't be better to run the entire DPC under a lock, and perhaps even use one of the KMDF-provided locks via AutomaticSerialization=TRUE.

AlekseyKostyushko:
virtqueues are created/deleted during D0 Entry/Exit callbacks in BalloonInit/BalloonTerm
these callbacks are called with interrupts disabled. I agree that this is not safe for DPCs (because I've failed to find how WDF flush them).
From another side, we currently see no bsods during rather extreme tests (such as cpu hotplug).

ladipro:
Sorry, I should have been clearer. Buffers are added to virtqueues in routines called on the worker thread and then fetched here in the DPC handler. These two should be synchronized. It may happen to work fine today as a result of the balloon protocol but it's not pretty. So just wondering, if it wouldn't be worth fixing as well, potentially with a simpler fix - using the KMDF device lock instead of adding a new spinlock.

AlekseyKostyushko:
Let talk exactly about potential races between DPC and BalloonTerm:
a) virtqueue_get_buf vs DeleteQueue
b) there is no races with WakeUpThread event
c) HostAckEvent event in DPC => BalloonTellHost vs DeleteQueue

do you see more?

ladipro:
Thanks, yes these would be potential races if DPC could actually run in parallel with D0Exit / BalloonTerm. But I don't believe that's the case.

I'm looking at the Windows-Driver-Frameworks repo here at GitHub and the framework appears to be flushing DPCs as part of disabling interrupts *before* it calls the D0Exit callback.

FxPkgPnp::PowerGotoD3Stopped calls FxPkgPnp::NotifyResourceObjectsDx which calls FxInterrupt::Disconnect on each interrupt which calls FlushQueuedDpcs.

So I'd say that unless I'm not looking at the right source code (please confirm), it is not necessary to synchronize DPC and the power callbacks. Was issue discovered only by code inspection or has there been an evidence of an actual race?

What I believe is a bug is the concurrent access to virtqueues (i.e. virtqueue_add_buf and virtqueue_get_buf) from the DPC, the worker thread, and the workitem worker (BalloonMemStats).

AlekseyKostyushko:
FYI: at least FlushQueuedDpcs is available starting from WinXP SP2

my first attempt to fix bsod was to keep a copy of StatWorkItem pointer and restore it in BalloonInit (while keeping in D0Exit "StatWorkItem = NULL"), but then we've found that during cpu hotplug DPC is still executed with StatWorkItem equal to NULL => bsod

I've read virtqueue_add_buf  / virtqueue_get_buf in VirtIO - and actually I don't understand how it works without sync?! Also we call virtqueue_get_buf in DPC only to check its existance, may be we need to have smth like virtqueue_is_empty?

ladipro:
> but then we've found that during cpu hotplug DPC is still executed with StatWorkItem equal to NULL => bsod

That's odd, the D0Entry callback runs with interrupts disabled so there shouldn't be any DPCs until it returns. Is it possible that DPC didn't see the non-NULL StatWorkItem due to a missing memory barrier?

I would simplify the first patch to just this:

-      if (devCtx->StatWorkItem)
-      {
          WdfWorkItemFlush(devCtx->StatWorkItem);
-        devCtx->StatWorkItem = NULL;
-      }

If KMDF didn't wait for DPC rundown for us, things would be broken in pretty much all virtio-win drivers.

> I've read virtqueue_add_buf / virtqueue_get_buf in VirtIO - and actually I don't understand how it works without sync?! Also we call virtqueue_get_buf in DPC only to check its existance, may be we need to have smth like virtqueue_is_empty?

Yeah, maybe it happens to work by accident because of the usage pattern. Feel free to ignore this for now - I mentioned it because you were adding a lock so it would have been nice to fix this too but if we agree that no new lock is needed, it can be handled separately.

AlekseyKostyushko:
It will take some time to check the exact behaviour if  DPC required sync beyond WDF or not. If not - you're right about simplified patch.
Let me take a pause in conversation and return when I'll obtain results.

ladipro:
Closing based on the commit message in https://github.com/virtio-win/kvm-guest-drivers-windows/pull/146. Thanks!

PR #86, opened 2016-12-10
Return correct info for OID_TCP_OFFLOAD_PARAMETERS

fengmin75:
Docker or some network filter driver could hang if netkvm return RSC as enabled while OID is requested it to be off.

### DIFF
--- a/NetKVM/wlh/ParaNdis6-Oid.cpp
+++ b/NetKVM/wlh/ParaNdis6-Oid.cpp
@@ -703,8 +703,8 @@ void ParaNdis6_FillOffloadConfiguration(PARANDIS_ADAPTER *pContext)
     NDIS_OFFLOAD *po = &pContext->ReportedOffloadConfiguration;
     FillOffloadStructure(po, pContext->Offload.flags);
 #if PARANDIS_SUPPORT_RSC
-    po->Rsc.IPv4.Enabled = (pContext->RSC.bIPv4SupportedSW && pContext->RSC.bIPv4SupportedHW);
-    po->Rsc.IPv6.Enabled = (pContext->RSC.bIPv6SupportedSW && pContext->RSC.bIPv6SupportedHW);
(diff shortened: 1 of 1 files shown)


NetKVM/wlh/ParaNdis6-Oid.cpp
```diff
@@ -703,8 +703,8 @@ void ParaNdis6_FillOffloadConfiguration(PARANDIS_ADAPTER *pContext)
     NDIS_OFFLOAD *po = &pContext->ReportedOffloadConfiguration;
     FillOffloadStructure(po, pContext->Offload.flags);
 #if PARANDIS_SUPPORT_RSC
-    po->Rsc.IPv4.Enabled = (pContext->RSC.bIPv4SupportedSW && pContext->RSC.bIPv4SupportedHW);
```
YanVugenfirer: Hi,

Thank you for the patch.

In general - if backend doesn't support RSC and we set this feature, the driver will still function.
But I don't think it is a good idea. Also this patch might hide several actual issues (OnSetRSCParameters overrides initial setting of RSC.bIPv4Enabled and  RSC.bIPv6Enabled).

Best regards,
Yan.

fengmin75: HI Yan,

If the backend doesn't support RSC, protocol stack/overlay driver won't turn it ON. The problem is that the system might request to turn if off if the netkvm/backend set it as ON. One of the reason is that network filter driver doesn't support RSC and then need to turn it off by sending a OID_TCP_OFFLOAD_PARAMETERS request. In current code, we return NDIS_SUCCESS while still indicating RSC as enabled. The several filter driver registration/startup will just hang in there forever.

You can observe it by installing docker on a VM with RSC enabled. By default, docker will create NAT vswitch. winnat.sys will attach a filter driver which in turn will try to turn off RSC. (docker won't start).

YanVugenfirer:
Hi,

Thank you for the explanation.

PR #91, opened 2017-01-10
Fix verifier and debug spew due to worker exiting with affinity set.

iuliur:
By default the workers have no affinity. We set it to the specific cpu, but KeSetSystemAffinityThreadEx returns 0, as there was no affinity. You still have to KeSetSystemAffinityThreadEx before you exit and reset it back to 0


vrozenfe:
Thank you. 
Applied on our internal repository.
Vadim.

PR #95, opened 2017-02-10
vioserial: Don't set interrupt vectors if MSI is disabled

ladipro:
VirtIOWdfInitQueuesCB introduced in commit 48db3727 attempts to
set config and queue vectors even if they are VIRTIO_MSI_NO_VECTOR.
This may fail because the IO region in QEMU does not necessarily
include the VIRTIO_MSI_CONFIG_VECTOR and VIRTIO_MSI_QUEUE_VECTOR
fields (i.e. it is smaller) so the writes are ignored and reading
the fields back yields 0. This is interpreted as a failure to set
the vector (0 != VIRTIO_MSI_NO_VECTOR).

This fixes https://github.com/YanVugenfirer/kvm-guest-drivers-windows/issues/94


ladipro:
Already merged as https://github.com/YanVugenfirer/kvm-guest-drivers-windows/commit/a528eb6e6790675a33f3612cca31462264d19a58

PR #96, opened 2017-02-13
[vioserial] fix VIRTIO_CONSOLE_PORT_OPEN fails

up2wing:
Handler of VIRTIO_CONSOLE_PORT_OPEN event may fail when guest OS reboot, because VIOSerialHandleCtrlMsg may be invoked concurrently, but  the procedure of this event is not protected by any lock.

For example:
[vioserial] [level-3] --> VIOSerialHandleCtrlMsg::123  Device:FFFFE00114549820 Devid:0 id:1 event:6 value:0 iodevice:FFFFE001145615F0
[vioserial] [level-3] --> VIOSerialHandleCtrlMsg::123  Device:FFFFE00114549820 Devid:0 id:1 event:6 value:1 iodevice:FFFFE001145615F0
[vioserial] [level-2] VIOSerialHandleCtrlMsg::179 port:org.qemu.guest_agent.0 VIRTIO_CONSOLE_PORT_OPEN id = 1, HostConnected = 0 current HostConnected:1
[vioserial] [level-2] VIOSerialHandleCtrlMsg::179 port:org.qemu.guest_agent.0 VIRTIO_CONSOLE_PORT_OPEN id = 1, HostConnected = 1 current HostConnected:1
[vioserial] [level-3] --> VIOSerialPortPnpNotify
[vioserial] [level-3] <-- VIOSerialHandleCtrlMsg::217
[vioserial] [level-3] <-- VIOSerialHandleCtrlMsg::217

### DIFF
--- a/vioserial/sys/Control.c
+++ b/vioserial/sys/Control.c
@@ -159,6 +159,7 @@ VIOSerialHandleCtrlMsg(
         break;
 
         case VIRTIO_CONSOLE_PORT_OPEN:
+           WdfSpinLockAcquire(pContext->CCtrlLock);
            if (port)
            {
               BOOLEAN  Connected = (BOOLEAN)cpkt->value;
@@ -189,6 +190,7 @@ VIOSerialHandleCtrlMsg(
            {
               TraceEvents(TRACE_LEVEL_ERROR, DBG_PNP, "VIRTIO_CONSOLE_PORT_OPEN invalid id = %d\n", cpkt->id);
            }
+           WdfSpinLockRelease(pContext->CCtrlLock);
         break;
 
         case VIRTIO_CONSOLE_PORT_NAME:
--- a/vioserial/sys/Device.c
+++ b/vioserial/sys/Device.c
@@ -192,6 +192,7 @@ VIOSerialEvtDevicePrepareHardware(
     UINT nr_ports;
     u64 u64HostFeatures;
     u64 u64GuestFeatures = 0;
+    WDF_OBJECT_ATTRIBUTES  attributes;
 
     UNREFERENCED_PARAMETER(ResourcesRaw);
     PAGED_CODE();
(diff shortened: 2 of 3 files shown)


YanVugenfirer:
Hi,

Thank you very much for the report and the fix!

I have a small request in order to merge the fix into the code:

1. Please send it as one commit (in you repository you can merge several
commits into one by using "git rebase -i HEAD~3" (3 in this case the number
of commits) and then squashing the commits.

2. Also please add "Sign Off" message. For example: "Signed-off-by: Yi
Wang <Yi.Wang@yourdomain.com>"
The easies way to do it is to use "git commit -s" or you can change the
commit message when you will be doing the interactive rebase.

Thanks,
Yan.

ladipro:
Even with this fix, the messages can still be processed in the wrong order. For example, the ctrl queue receives connected=0 followed by connected=1, but connected=1 may acquire CCtrlLock first, so the port incorrectly ends up with port->HostConnected=0.

I think that we should instead increase the scope of CVqLock to keep (some of) the operations in VIOSerialHandleCtrlMsg serialized.

up2wing:
@YanVugenfirer Thanks for your reply! Should I open a new Pull Request?

@ladipro Thanks for your review. But I suppose increasing the scope of CVqLock to control VIOSerialHandleCtrlMsg is not a good idea, 'cause VIOSerialHandleCtrlMsg may call some functions such like VIOSerialInitPortConsole, which will acquire CVqLock. How about using CCtrlLock in my commit to do that?

ladipro:
Of course, we would need to be careful to not hold the lock when it's not supposed to be held.

CCtrlLock would work but you'd have to acquire it before you release CVqLock so it would probably be just as complex as refactoring CVqLock, if not more.

Do you agree with the high-level point that most of the operations in VIOSerialHandleCtrlMsg are order-sensitive and should be processed in the same order in which they were popped from the virtqueue? VIRTIO_CONSOLE_PORT_ADD and VIRTIO_CONSOLE_PORT_REMOVE have the same problem as VIRTIO_CONSOLE_PORT_OPEN. Thanks!

ladipro:
Actually - what if we set AutomaticSerialization=TRUE for VIOSerialInterruptDpc? This is separate from VIOSerialQueuesInterruptDpc so it's fine if it takes a bit of a perf hit. Maybe we could then eliminate CVqLock altogether.

up2wing:
Set AutomaticSerialization=TRUE sounds good, and CVqLock can be eliminated altogether. I will test this and will open a new RP if it works.
Thanks! @ladipro

PR #97, opened 2017-02-14
support for multiple vendors

dlunev:
(v4) - tools and all drivers except NetKVM and viorng
(Balloon, pvpanic, vioinput, vioserial, vioscsi, viostor)

Introduce $(_VENDOR_) property (rhel by default).

There is no way to make substitutions inside inf file
except custom MSBuild target imported from .vcxproj

WDK/LegacyDDK scenarios are slightly different:

In case of WDK all that is needed is to execute inf substitution task
just after StampInf target had stamped .inx/.inf and moved it into
$(OutDir) and before InfVerif target.

In case of LegacyDDK we need to do things in the middle of PostBuild event
that copyes inf files, calls stampinf and calls packone.bat
That's why this logic was translated into MSBuild targets

Driver.Common.targets contains inf substituion and copying/stampinf as it
was in PostBuildEvent
Driver.PackOne.targets replaces packone.bat

Features (all feature are false by default):
1. $(Feature_UsingWDF) - do we use WDF?
2. $(Feature_AdjustInf) and $(Feature_AdjustInfLegacy)
   executes SubstituteFileContent task
   to substitute vendor specific strings in .inf/.inx files
   with @(Substitution) from Driver.$(_VENDOR_).props
   $(SourceInfFile) is required for Legacy case
3. $(Feature_LegacyStampInf) - executes
   stampinf tool for Legacy DDK
4. $(Feature_PackOne) - Import Driver.PackOne.targets

Removed rhel versioning from Driver.Common.props,
added import of Driver.Vendor.props,
and simplified InfArch definitions

Driver.Vendor.props introduces schema to support different vendors
Driver.RHEL.props contains versioning and inf substitutions for Red Hat
Driver.VZ.props contains same for Virtuozzo

Due to the fact that version info
in .rc files for all drivers is almost the same this patch introduces
a schema for simplifying them.
.rc files should only declare VENDOR_VIRTIO_PRODUCT,
VER_FILEDESCRIPTION_STR, VER_INTERNALNAME_STR

### DIFF
--- a/Balloon/sys/balloon.vcxproj
+++ b/Balloon/sys/balloon.vcxproj
@@ -160,6 +160,7 @@
     <GenerateManifest>false</GenerateManifest>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
+  <Import Project="$(MSBuildProjectDirectory)\balloon.props" />
   <Import Project="$(MSBuildProjectDirectory)\..\..\Tools\Driver.Common.props" />
   <ImportGroup Label="ExtensionSettings">
   </ImportGroup>
@@ -298,6 +299,7 @@
     <ClCompile Include="utils.c" />
   </ItemGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
+  <Import Project="$(MSBuildProjectDirectory)\..\..\Tools\Driver.Common.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
 </Project>
--- a/Balloon/sys/balloon.vcxproj
+++ b/Balloon/sys/balloon.vcxproj
@@ -259,20 +259,8 @@
     <ClCompile>
       <DisableSpecificWarnings>4267;%(DisableSpecificWarnings)</DisableSpecificWarnings>
     </ClCompile>
-    <PostBuildEvent>
-      <Command>
-        copy balloon.inx $(OutDir)\balloon.inf
-        stampinf -f $(OutDir)\balloon.inf -a $(InfArch) -k 1.9 -v $(STAMPINF_VERSION)
-        packOne.bat $(TargetOS) $(TargetArch) balloon "$(LegacyDDKDir)"
-      </Command>
-    </PostBuildEvent>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(PlatformToolset)'!='v140_xp'">
-    <PostBuildEvent>
-      <Command>
-        packOne.bat $(TargetOS) $(TargetArch) balloon "$(WDKContentRoot)"
-      </Command>
-    </PostBuildEvent>
   </ItemDefinitionGroup>
   <ItemGroup>
     <FilesToPackage Include="$(TargetPath)" />
--- a/Tools/Driver.Common.targets
+++ b/Tools/Driver.Common.targets
@@ -1,0 +1,113 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+***********************************************************************************************
+Driver.Common.targets
+Common targets used by all drivers.
+
+Features (all feature are false by default):
+  1. $(Feature_UsingWDF) - do we use WDF?
+  2. $(Feature_AdjustInf) and $(Feature_AdjustInfLegacy)
+      executes SubstituteFileContent task
+      to substitute vendor specific strings in .inf/.inx files
+      with @(Substitution) from Driver.$(_VENDOR_).props
+     $(SourceInfFile) is required for Legacy case
+  3. $(Feature_LegacyStampInf) - executes stampinf tool for Legacy DDK
+  4. $(Feature_PackOne) - Import Driver.PackOne.targets
+***********************************************************************************************
+-->
+<Project ToolVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+<!--
+look at https://msdn.microsoft.com/ru-ru/library/dd722601.aspx?f=255&MSPPError=-2147217396 for an example of TokenReplace task
+look at http://blogs.clariusconsulting.net/kzu/how-to-perform-regular-expression-based-replacements-on-files-with-msbuild/ for an example of RegexTransform task
+============================================================
+            SubstituteFileContent
+Replaces text in sources files with multiple substitutions.
+required Items: SourceFiles and Substituions (with Identity/ReplaceWith metadata)
+
+Example item:
+        <Substition Include="MACRO_VENDOR">
+            <ReplaceWith>Vendor Corp.</ReplaceWith>
+        </Substition>
+
+Invoking the task:
+    <SubstituteFileContent SourceFiles="@(SourceFile)" Substituions="@(Substition)" />
+============================================================
+-->
+  <UsingTask TaskName="SubstituteFileContent"
+             TaskFactory="CodeTaskFactory"
+             AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.v4.0.dll">
+    <ParameterGroup>
+      <SourceFiles ParameterType="Microsoft.Build.Framework.ITaskItem[]" Required="true" />
+      <Substituions ParameterType="Microsoft.Build.Framework.ITaskItem[]" Required="true" />
+    </ParameterGroup>
+    <Task>
+      <Using Namespace="System.IO" />
+      <Using Namespace="System.Linq" />
+      <Using Namespace="System.Collections.Generic" />
+      <Using Namespace="System.Text.RegularExpressions" />
+      <Using Namespace="Microsoft.Build.Framework" />
+      <Code Type="Fragment" Language="cs">
+        <![CDATA[
+        var substitutions = new Dictionary<string,string>();
+        foreach (var item in Substituions) {
+            var replaceWhat = item.GetMetadata("Identity");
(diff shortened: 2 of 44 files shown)


YanVugenfirer:
Thank you for the patches! We will review them soon.

Balloon/sys/balloon.vcxproj
```diff
@@ -259,20 +259,8 @@
     <ClCompile>
       <DisableSpecificWarnings>4267;%(DisableSpecificWarnings)</DisableSpecificWarnings>
     </ClCompile>
-    <PostBuildEvent>
-      <Command>
-        copy balloon.inx $(OutDir)\balloon.inf
-        stampinf -f $(OutDir)\balloon.inf -a $(InfArch) -k 1.9 -v $(STAMPINF_VERSION)
-        packOne.bat $(TargetOS) $(TargetArch) balloon "$(LegacyDDKDir)"
```
ladipro: Should deleting all packOne.bat files be part of this commit?

Tools/Driver.PackOne.targets
```diff
@@ -0,0 +1,203 @@
...
+    <PackOne_WdfArch>x86</PackOne_WdfArch>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetArch)'=='amd64'">
+    <PackOne_SourcePathArch>amd64</PackOne_SourcePathArch>
+    <PackOne_WdfArch>x64</PackOne_WdfArch>
+  </PropertyGroup>
+
+  <!-- $(PackOne_SourcePath) and $(PackOne_DestinationPath) -->
+  <PropertyGroup>
+    <PackOne_SourcePath>objfre_$(TargetOS)_$(TargetArch)\$(PackOne_SourcePathArch)</PackOne_SourcePath>
```
ladipro: Ideally, this would also handle CHK builds. Not a big deal though, we can add it later.

vioscsi/vioscsi.vcxproj
```diff
@@ -204,20 +205,8 @@
...
-    <PostBuildEvent>
-      <Command>
-        copy vioscsi.inx $(OutDir)\vioscsi.inf
-        stampinf -f $(OutDir)\vioscsi.inf -a $(InfArch) -v $(STAMPINF_VERSION)
-        packOne.bat $(TargetOS) $(TargetArch) vioscsi
-      </Command>
-    </PostBuildEvent>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(PlatformToolset)'!='v140_xp'">
-    <PostBuildEvent>
```
ladipro: nit: Can remove the now-empty ItemDefinitionGroup as well.

vioserial/sys/vioser.vcxproj
```diff
@@ -250,20 +251,8 @@
...
-    <PostBuildEvent>
-      <Command>
-        copy vioser.inx $(OutDir)\vioser.inf
-        stampinf -f $(OutDir)\vioser.inf -a $(InfArch) -k 1.9 -v $(STAMPINF_VERSION)
-        packOne.bat $(TargetOS) $(TargetArch) vioser "$(LegacyDDKDir)"
-      </Command>
-    </PostBuildEvent>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(PlatformToolset)'!='v140_xp'">
-    <PostBuildEvent>
```
ladipro: nit: Can remove the now-empty ItemDefinitionGroup as well.

viostor/viostor.vcxproj
```diff
@@ -204,20 +205,8 @@
...
-    <PostBuildEvent>
-      <Command>
-        copy $(SourceInfFile) $(OutDir)\viostor.inf
-        stampinf -f $(OutDir)\viostor.inf -a $(InfArch) -v $(STAMPINF_VERSION)
-        packOne.bat $(TargetOS) $(TargetArch) viostor
-      </Command>
-    </PostBuildEvent>
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(PlatformToolset)'!='v140_xp'">
-    <PostBuildEvent>
```
ladipro: nit: Can remove the now-empty ItemDefinitionGroup as well.

YanVugenfirer:
On Fri, Feb 17, 2017 at 11:12 AM, Ladi Prosek <notifications@github.com>
wrote:

> *@ladipro* requested changes on this pull request.
>
> Overall, this is awesome. I have added a few more comments, please take a
> look. Thanks!
> ------------------------------
>
> In Tools/rhel.ver
> <https://github.com/virtio-win/kvm-guest-drivers-windows/pull/97#discussion_r101700788>
> :
>
> > + * the COPYING file in the top-level directory.
> + *
> +**********************************************************************/
> +
> +/*
> + * rhel versioning
> + */
> +#define VENDOR_VIRTIO_1            _NT_TARGET_MAJ
> +#define VENDOR_VIRTIO_2            _RHEL_RELEASE_VERSION_
> +#define VENDOR_VIRTIO_3            _BUILD_MAJOR_VERSION_
> +#define VENDOR_VIRTIO_4            _BUILD_MINOR_VERSION_
> +
> +/*
> + * rhel strings
> + */
> +#define VENDOR_VIRTIO_COPYRIGHT    "Copyright (C) 2008-2017 Red Hat, Inc."
>
> Each RHEL driver has its legal copyright string reflecting the year when
> it was conceived. With this change, they all become "2008-2017" which we
> likely don't want. Can we parameterize the "from" year?
> CC @YanVugenfirer <https://github.com/YanVugenfirer>, @vrozenfe
> <https://github.com/vrozenfe>
>

We can add definition per project. For that we can use include file with
the same name across the projects, and it will be used when specific binary
is compiled.

Tools/rhel.ver
```diff
@@ -0,0 +1,37 @@
...
+ */
+#define VENDOR_VIRTIO_1            _NT_TARGET_MAJ
+#define VENDOR_VIRTIO_2            _RHEL_RELEASE_VERSION_
+#define VENDOR_VIRTIO_3            _BUILD_MAJOR_VERSION_
+#define VENDOR_VIRTIO_4            _BUILD_MINOR_VERSION_
+
+/*
+ * rhel strings
+ */
+#define VENDOR_VIRTIO_COPYRIGHT    "Copyright (C) 2008-2017 Red Hat, Inc."
```
AlekseyKostyushko: Ok, will be parameterized

Tools/Driver.Common.targets
```diff
@@ -0,0 +1,113 @@
...
+            if(!File.Exists(fileName))
+            {
+                Log.LogMessage(MessageImportance.High, "Could not find file: {0}", fileName);
+                continue;
+            }
+
+            Log.LogMessage(MessageImportance.High, "Saving original context to: {0}", fileNameOrig);
+            File.Copy(fileName, fileNameOrig, true);
+
+            string content = File.ReadAllText(fileName);
```
AlekseyKostyushko: ReadLine will bring possible issues with the last empty line, /n/r format et.c.
As it recommended, for small files <1Mb it is better not to deal with that at all using ReadAllText

Tools/Driver.PackOne.targets
```diff
@@ -0,0 +1,203 @@
...
+    <ParameterGroup>
+      <Value Required="true" />
+      <Item ParameterType="Microsoft.Build.Framework.ITaskItem[]" Required="true" />
+      <Result ParameterType="System.Boolean" Output="true" />
+    </ParameterGroup>
+    <Task>
+      <Using Namespace="System" />
+      <Code Type="Fragment" Language="cs">
+        <![CDATA[
+        Result = false;
```
AlekseyKostyushko: Thanks!

AlekseyKostyushko:
I'm preparing next version of this patchset that will incorporate all mentioned changes

dlunev:
we are going to re-submit today with all changes made. Thank you for notification

Tools/Driver.PackOne.targets
```diff
@@ -0,0 +1,203 @@
...
+  <!-- from packone.bat create_xp variant
+       both Wxp and Wnet drivers have mask for both Windows XP and Server2003 OSes -->
+  <PropertyGroup Condition="'$(TargetOS)'=='Wxp' AND '$(TargetArch)'=='x86'">
+    <PackOne_OsMask>XP_X86,Server2003_X86</PackOne_OsMask>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetOS)'=='Wxp' AND '$(TargetArch)'=='amd64'">
+    <PackOne_OsMask>XP_X64,Server2003_X64</PackOne_OsMask>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetOS)'=='Wnet' AND '$(TargetArch)'=='x86'">
+    <PackOne_OsMask>XP_X86,Server2003_X86</PackOne_OsMask>
```
AlekseyKostyushko: This is how packone.bat is currently written.
Let change current behaviour later?

ladipro: I found it by diffing the .cat files with and without your changes:

```
Diff C:\Users\lprosek\Desktop\baseline\Wnet\x86\viostor.cat vs C:\Users\lprosek\Desktop\vendor1\Wnet\x86\viostor.cat:
3c3
<     OS: Server2003X86
---
>     OS: XPX86,Server2003X86
```

It looks like we're currently not consistent across drivers, but I think that viostor is more correct than the rest.

dlunev:
created new pull request https://github.com/virtio-win/kvm-guest-drivers-windows/pull/108

PR #103, opened 2017-02-22
Revert "Fix queue locking."

ladipro:
This reverts commit ed1d377efdd04223bdfbebc23ace1c230aa6a0eb.

By not acquiring MSI spin locks anymore, the commit breaks
synchronization in the driver. Both VioScsiMSInterruptWorker
and the code that calls VioScsiVQLock with isr=FALSE, like
SendSRB, access the request virtqueues and these accesses must
be serialized.

See comments in VirtioLib, for example:

 * virtqueue_add_buf - expose buffer to other end
 * ...
 * Caller must ensure we don't call this with other virtqueue operations
 * at the same time (except where noted).

 * virtqueue_get_buf - get the next used buffer
 * ...
 * Caller must ensure we don't call this with other virtqueue
 * operations at the same time (except where noted).

The issue manifests as the VM hanging under I/O stress, looping
in vioscsi!ProcessQueue indefinitely.

3: kd> k
 # Call Site
00 vioscsi!virtqueue_get_buf+0x4c [c:\kvm-guest-drivers-windows\virtio\virtioring.c @ 536]
01 vioscsi!ProcessQueue+0xee [c:\kvm-guest-drivers-windows\vioscsi\vioscsi.c @ 1287]
02 vioscsi!DispatchQueue+0x4f [c:\kvm-guest-drivers-windows\vioscsi\vioscsi.c @ 1258]
03 vioscsi!VioScsiMSInterruptWorker+0x63 [c:\kvm-guest-drivers-windows\vioscsi\vioscsi.c @ 986]
04 vioscsi!VioScsiMSInterrupt+0x23 [c:\kvm-guest-drivers-windows\vioscsi\vioscsi.c @ 1052]
05 storport!RaidpAdapterMSIInterruptRoutine+0x61
06 nt!KiCallInterruptServiceRoutine+0x87

### DIFF
--- a/vioscsi/helper.c
+++ b/vioscsi/helper.c
@@ -431,17 +431,30 @@ VioScsiVQLock(
 ENTER_FN();
     adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
 
-    if (!isr) {
-        if (adaptExt->msix_enabled) {
-            if (!CHECKFLAG(adaptExt->perfFlags, STOR_PERF_ADV_CONFIG_LOCALITY)) {
-                // Queue numbers start at 0, message ids at 1.
-                NT_ASSERT(MessageID > VIRTIO_SCSI_REQUEST_QUEUE_0);
-                NT_ASSERT(MessageID <= VIRTIO_SCSI_REQUEST_QUEUE_0 + adaptExt->num_queues);
-                StorPortAcquireSpinLock(DeviceExtension, DpcLock, &adaptExt->dpc[MessageID - VIRTIO_SCSI_REQUEST_QUEUE_0 - 1], LockHandle);
+    if (!adaptExt->msix_enabled) {
+        if (!isr) {
+            StorPortAcquireSpinLock(DeviceExtension, InterruptLock, NULL, LockHandle);
+        }
+    }
+    else {
+        if (adaptExt->num_queues == 1) {
+            if (!isr) {
+                ULONG oldIrql = 0;
+                StorPortAcquireMSISpinLock(DeviceExtension, (adaptExt->msix_one_vector ? 0 : MessageID), &oldIrql);
+                LockHandle->Context.OldIrql = (KIRQL)oldIrql;
             }
         }
         else {
-            StorPortAcquireSpinLock(DeviceExtension, InterruptLock, NULL, LockHandle);
+            NT_ASSERT(MessageID > VIRTIO_SCSI_REQUEST_QUEUE_0);
+            NT_ASSERT(MessageID <= VIRTIO_SCSI_REQUEST_QUEUE_0 + adaptExt->num_queues);
+            if (CHECKFLAG(adaptExt->perfFlags, STOR_PERF_CONCURRENT_CHANNELS)) {
+                if (CHECKFLAG(adaptExt->perfFlags, STOR_PERF_ADV_CONFIG_LOCALITY)) {
+                    StorPortAcquireSpinLock(DeviceExtension, StartIoLock, &adaptExt->dpc[MessageID - VIRTIO_SCSI_REQUEST_QUEUE_0 - 1], LockHandle);
+                }
+                else {
+                    RhelDbgPrint(TRACE_LEVEL_FATAL, ("%s STOR_PERF_CONCURRENT_CHANNELS yes, STOR_PERF_ADV_CONFIG_LOCALITY no\n", __FUNCTION__));
+                }
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


ladipro:
CC @iuliur who authored ed1d377 and @vrozenfe who pushed it.

vrozenfe:
oh yeah,
I saw similar issue as well. It's why I didn't apply in on stable :)

Thanks,
Vadim.

iuliur:
Sorry we took out the adv_config_locality if last week but I got caught up
in some other stuff and forgot to push

ladipro:
On Thu, Feb 23, 2017 at 1:36 AM, iuliur <notifications@github.com> wrote:

> Sorry we took out the adv_config_locality if last week but I got caught up
> in some other stuff and forgot to push


​Are you planning to submit a fix on top of ed1d377
<https://github.com/virtio-win/kvm-guest-drivers-windows/commit/ed1d377efdd04223bdfbebc23ace1c230aa6a0eb>
or
should we merge this revert for now? Thanks!​

iuliur:
https://github.com/iuliur/kvm-guest-drivers-windows/commit/9e7485de31a2300ba6a78112dc066d40d908a51d

ladipro:
I'm afraid iuliur@9e7485d won't help, you're still not synchronizing with the MSI handler.

iuliur:
I don't think I have the contexts then - what are you seeing?

ladipro:
I tried to capture that in the commit message: https://github.com/virtio-win/kvm-guest-drivers-windows/pull/103/commits/0cac38030f5f2a286ca1de58fd117b177828e8f3

Before your change, VioScsiVQLock acquired the MSI spinlock if adaptExt->num_queues == 1. This was important because the MSI handler accesses the request virtqueue and it must not do it concurrently with the code that called VioScsiVQLock.

VioScsiMSInterrupt -> VioScsiMSInterruptWorker -> DispatchQueue -> ProcessQueue -> virtqueue_get_buf

iuliur:
Ah, i see. Our code for DispathQueue is different. We fire the dpc for cpu
0 in case num_queues == 1. That's why we didnt' see that.
I don't know how the code diverged, i thought we pushed this at some point
in the past.

iuliur:
Or  DIspatchQueue looks like this:
https://github.com/iuliur/kvm-guest-drivers-windows/commit/bcdfd85ac79566cb66341f7ae5417fd49f68a386

ladipro:
Gotcha, thanks. Does https://github.com/iuliur/kvm-guest-drivers-windows have all differences between virtio-win/master and your code or is there another public repo you use?

iuliur:
No, there were changes you didn't accept. There are some other differences
such as you and us adding wmi at the same time, and our code is in a
different wmi.c file. We also added trace to ETW in windows 8.1 to make it
easier for customers to get support.
I can either push everything in a different repro or try to split the
changes that you want.

ladipro:
Pushing everything to a different repo and opening a pull request would be awesome. I can't promise that everything will be accepted but at least the differences would be kind of documented.

For this particular difference of completing requests in the ISR vs. DPC, do you have performance numbers to back the always-use-DPC approach? Thank you!

iuliur:
https://github.com/YanVugenfirer/kvm-guest-drivers-windows/pull/2

Couple of notes:
 - we build with wdk lab kit (
https://msdn.microsoft.com/en-us/windows/hardware/drivers/develop/installing-the-wdk-build-environment-in-a-lab?f=255&MSPPError=-2147217396)
not with visual studio so i can't guarantee if you copy everything your
build scripts will work. We do this because our build server don't have
visual studio installed and you can just copy the wkd lab build around
without having to install anything.
- we have unit tests that run on linux (they compile the drivers as shared
libraries) so you'll see some changes from unsigned long to ULONG and some
general portability fixes.
- our Netkvm version is behind yours.

As for the perf numbers difference for one queue, back when we did the
change the difference was insignificant. Also, we don't really care about
the one queue perf case. All the cloud customers who want perf are on 4
cpus or higher.

vrozenfe:
Have you thought about Enterprise WDK? Unless you need SDV or Code Analysis, it should also work for your case.  
Regarding to single queue configuration. For some load scenarios  single queue performs better than multi-queue, even on SMP VMs. I would still take single queue into account when we speak about performance.

Cheers,
Vadim.

iuliur:
Enterprise wdk is the same thing, it's the renamed wdk10 lab install.
That's what we have. We made it built XP too but unlike you we made the
changes straight to the wdk files

iuliur:
As for the one queue, that might be true but we don't give customers the
option to configure number of queues.

ladipro:
So to move this forward and unbreak master - what I'm hearing is that there is a small performance benefit in completing requests in the ISR in single-queue configuration. Thus the more complex VioScsiVQLock / VioScsiVQUnlock is justified and ed1d377 should be reverted and/or redone.

@vrozenfe if you agree, can you please approve this pull request.

@iuliur thanks for sharing your sources. I'll go over them and report back. Going forward it would be nice to bring the code bases closer together so the same fixes and improvements apply to both and we don't run into issues like this one. Of course, we each support a different product so there will be instances where it's not easy to tell what's right and wrong because it depends on the hypervisor, use-case, etc.. Thanks!

iuliur:
I don't think there is a perfectly benefit from completing in isr. From
what I remember iops go up a bit but latency also goes up a bit. I would
have to rerun though.

ladipro:
A relevant thread on netdev to watch:
http://www.osronline.com/showthread.cfm?link=282412

ladipro:
Thanks, so we don't need this revert then. We want https://github.com/iuliur/kvm-guest-drivers-windows/commit/bcdfd85ac79566cb66341f7ae5417fd49f68a386 and https://github.com/iuliur/kvm-guest-drivers-windows/commit/9e7485de31a2300ba6a78112dc066d40d908a51d, is that correct?

vrozenfe:
Wait a little.
I applied both these patches on upstream's master and tried to re-test it again.
Unfortunately it crashes my test VM with DPC_WATCHDOG_VIOLATION bugcheck code 
every time I'm trying to start IoMeter.

All the best,
Vadim.

iuliur:
Must be something new or some subtle difference cause we don't see any dpc
watchdog bugchecks.

ladipro:
Vadim, did you also fix the line in VioScsiHwInitialize where we call StorPortEnablePassiveInitialization? We want to do passive initialization regardless of the number of queues.

ladipro:
https://github.com/virtio-win/kvm-guest-drivers-windows/pull/125 seems to work fine for me, please review. Thanks!

ladipro:
Abandoned in lieu of https://github.com/virtio-win/kvm-guest-drivers-windows/pull/125.

PR #104, opened 2017-02-22
Merge stable to master: balloon

ladipro:
This is part of an effort to merge the stable branch (where Fedora and RHEL virtio-win drivers come from) into master.

The balloon driver diverged in early 2016 when the user-mode service app was eliminated from the master branch. Due to concerns with calling an undocumented API (ZwQuerySystemInformation with SystemPerformanceInformation), the work was not ported to the stable branch at that time.

This series brings the user-mode service and all its supporting code from stable to master, but keeps it under a newly introduced macro / environment variable USE_BALLOON_SERVICE. Once https://github.com/virtio-win/kvm-guest-drivers-windows/pull/97 is merged, the variable may be set from vendor-specific .props files.

CC @dlunev , @AlekseyKostyushko


Balloon/sys/public.h
```diff
@@ -12,7 +12,8 @@ DEFINE_GUID(GUID_DEVINTERFACE_BALLOON,
 #define VIRTIO_BALLOON_S_MINFLT   3   /* Number of minor faults */
 #define VIRTIO_BALLOON_S_MEMFREE  4   /* Total amount of free memory */
 #define VIRTIO_BALLOON_S_MEMTOT   5   /* Total amount of memory */
-#define VIRTIO_BALLOON_S_NR       6
+#define VIRTIO_BALLOON_S_AVAIL    6   /* Available memory */
```
AlekseyKostyushko: This new counter requires initialization in case without USE_BALLOON_SERVICE.
At least with 0, but see my comment in thread.

AlekseyKostyushko:
Regarding VIRTIO_BALLOON_S_AVAIL:
ullAvailVirtual
The amount of unreserved and uncommitted memory currently in the user-mode portion of the virtual address space of the calling process, in bytes.

I don't believe this is a proper value for this counter.
Let split introduction of this new counter as separate patch (within this patchset or not) and discuss the way to obtain it in Windows both in user-mode for app and kernel-mode.

AlekseyKostyushko:
**of the calling process**

ladipro:
On Mon, Feb 27, 2017 at 3:59 PM, AlekseyKostyushko <notifications@github.com
> wrote:

> Regarding VIRTIO_BALLOON_S_AVAIL:
> ullAvailVirtual
> The amount of unreserved and uncommitted memory currently in the user-mode
> portion of the virtual address space of the calling process, in bytes.
>
> I don't believe this is a proper value for this counter.
> Let split introduction of this new counter as separate patch (within this
> patchset or not) and discuss the way to obtain it in Windows both in
> user-mode for app and kernel-mode.
>

​Yes, this is definitely not the right value, @vrozenfe is already aware​.

dlunev:
AFAIR we have switched to the kernel mode approach as we were unable to find suitable user space counter.

Balloon/sys/public.h
ladipro: I think I had convinced myself that there's no need to do anything for counters that are available only with or only without USE_BALLOON_SERVICE.

The driver initializes the entire stats with -1's

  `RtlFillMemory (devCtx->MemStats, sizeof (BALLOON_STAT) * VIRTIO_BALLOON_S_NR, -1);`

and -1 means 'unset'. See reset_stats() in hw/virtio/virtio-balloon.c and the way s->stats is filled in virtio_balloon_receive_stats.

ladipro:
I have updated the pull request.

v1->v2:
* rebased
* added patch 4

I have verified that if the no-service balloon driver is installed, QEMU receives (-1, -1) in the last stats entry and the entry is ignored.

PR #106, opened 2017-02-28
balloon: Boost worker thread priority

ladipro:
The balloon worker thread may get starved and inflate/delate may
take an unreasonably long time if the system is under CPU stress.
This commit increases the base priority of the thread to
LOW_REALTIME_PRIORITY and adds the EVENT_INCREMENT dynamic priority
boost to all KeSetEvent calls.

Of course, there is no guarantee that the thread will be scheduled
and increasing the priority may be perceived as a peeing contest.
In this case though, given that the thread is idle and not scheduled
at all most of the time, and, in fact, logically an extension of the
DPC routine, higher priority is justified.


ladipro:
Still looking for review / approval, cc @vrozenfe @YanVugenfirer 
Thanks!

PR #107, opened 2017-03-08
vioinput: Switch from filter to bus/child architecture

ladipro:
The canonical KMDF HID driver uses mshidkmdf.sys (or its equivalent
viohidkmdf.sys) as the FDO for the device and the actual KMDF driver
as its lower filter to work around PnP KMDF limitations. This works,
except for the case where the same device, in terms of USB or PCI or
any other bus location, may expose different HID collections every time
it's plugged in. Because the instance path of the HID PDO (like for
example "HID-compliant mouse") is constructed only from the instance
path of the parent - likely just by replacing the bus ID with "HID" -
the system will reuse pre-existing devnodes, even if the HID collection
and the report descriptor are completely different and a keyboard,
for example, should be created instead of a mouse.

It's not hard to run into this issue with the vioinput driver. E.g.
switching from
-device virtio-mouse-pci
to
-device virtio-keyboard-pci
for the same PCI slot will result in the system loading the mouse HID
driver and the keyboard not working.

This commit works around it by making the KMDF driver a bus driver
which enumerates one and only child PDO, functionally driven by
viohidkmdf.sys. IOCTL IRPs are forwarded from viohidkmdf to vioinput
and processed just like before. The key difference is that vioinput
now generates the instance ID of the child, thus having the ability
to alter the instance path depending on the shape of the input device.
Specifically, vioinput computes a simple hash out of the HID report
descriptor and uses it as the child instance ID.

The child device is hidden from Device Manager so as not to confuse
users. It's not meant to be manipulated except through its parent.

### DIFF
--- a/vioinput/hidpassthrough/hidpassthrough.c
+++ b/vioinput/hidpassthrough/hidpassthrough.c
@@ -24,6 +24,7 @@
  */
 #include <wdm.h>
 #include <hidport.h>
+#include "pdo.h"
 
 static DRIVER_ADD_DEVICE HidPassthroughAddDevice;
 static DRIVER_UNLOAD     HidPassthroughUnload;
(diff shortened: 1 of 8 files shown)


ladipro:
I've run a few. Will run all of them before merging. Thanks!

ladipro:
And of course, the "PNP Rebalance" tests got me again. I have added a flag to make sure that the child PDO is created only the first time VIOInputEvtDevicePrepareHardware runs and updated the pull request. Tests are passing now, thanks!

PR #108, opened 2017-03-13
[V2] support for multiple vendors

dlunev:
(v6) - tools and all drivers except NetKVM
(Balloon, pvpanic, vioinput, vioserial, vioscsi, viostor, viorng)

Introduce $(VENDOR) property (rhel by default).

There is no way to make substitutions inside inf file
except custom MSBuild target imported from .vcxproj

WDK/LegacyDDK scenarios are slightly different:

In case of WDK all that is needed is to execute inf substitution task
just after StampInf target had stamped .inx/.inf and moved it into
$(OutDir) and before InfVerif target.

In case of LegacyDDK we need to do things in the middle of PostBuild event
that copyes inf files, calls stampinf and calls packone.bat
That's why this logic was translated into MSBuild targets

Driver.Common.targets contains inf substituion and copying/stampinf as it
was in PostBuildEvent
Driver.PackOne.targets replaces packone.bat

Features (all feature are false by default):

$(Feature_UsingWDF) - do we use WDF?
$(Feature_AdjustInf) and $(Feature_AdjustInfLegacy)
executes SubstituteFileContent task
to substitute vendor specific strings in .inf/.inx files
with @(Substitution) from Driver.$(VENDOR).props
$(SourceInfFile) is required for Legacy case
$(Feature_LegacyStampInf) - executes
stampinf tool for Legacy DDK
$(Feature_PackOne) - Import Driver.PackOne.targets
Removed rhel versioning from Driver.Common.props,
added import of Driver.Vendor.props,
and simplified InfArch definitions

Driver.Vendor.props introduces schema to support different vendors
Driver.RHEL.props contains versioning and inf substitutions for Red Hat
Driver.VZ.props contains same for Virtuozzo

Due to the fact that version info
in .rc files for all drivers is almost the same this patch introduces
a schema for simplifying them.
.rc files should only declare VENDOR_VIRTIO_PRODUCT,
VER_FILEDESCRIPTION_STR, VER_INTERNALNAME_STR

each .props contains RHEL_COPYRIGHT_STARTING_YEAR
and VZ_COPYRIGHT_STARTING_YEAR as well as vendor.props
contains COPYRIGHT_CURRENT_YEAR. this values applies both
to inx/inf and .rc files

Alexey V. Kostyushko (16):
  tools: added support for multiple vendors
  tools: added support for multiple vendors for .rc
  tools: fix for VS internal rc viewer
  tools: support for multiple vendors for .inx/.inf
  tools: support for multiple vendors (substitutions)
  tools: support multiple vendors (c) years
  balloon: fix vcxproj.filters with missing files
  balloon: support for multiple vendors
  vioinput: support for multiple vendors
  vioscsi: support for multiple vendors
  vioserial: add support for multiple vendors
  viostor: add support for multiple vendors
  pvpanic: add multiple vendor support
  added support for multiple vendors
  viorng: add support for multiple vendors
  all: added support for multiple vendors (c) years


pvpanic/PVPanic Package/pvpanic_pack.props
```diff
@@ -0,0 +1,15 @@
...
+pvpanic_pack.props
+Enabling and customizing virtio build features
+***********************************************************************************************
+-->
+
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" TreatAsLocalProperty="Platform">
+  <PropertyGroup>
+    <Feature_AlwaysDefaultVendor>false</Feature_AlwaysDefaultVendor>
+    <Feature_AdjustInf>true</Feature_AdjustInf>
+    <RHEL_COPYRIGHT_STARTING_YEAR>2015</RHEL_COPYRIGHT_STARTING_YEAR>
```
ladipro: This is unfortunate. Now we have these per-driver .props files with vendor entries in them. Makes it harder to separate vendor-specific properties from the general build properties. Ideally everything vendor specific would stay in its dedicated files.

Just brainstorming here. What if put the copyright starting years (plus anything else that might be different - like for example the balloon service/no-service feature) in per-driver per-vendor files and then:

  <Import Project="pvpanic_pack.$(_VENDOR_).props" />

or

  <Import Project="pvpanic_pack.$(_VENDOR_).props" Condition="exists(pvpanic_pack.$(_VENDOR_).props)" />


Or another option. To avoid adding many new files, this file would have something like:

  <CopyrightYears>%(PVPanicCopyrightYears)</CopyrightYears>

(similar for other drivers)

and PVPanicCopyrightYears would be defined in Driver.RHEL.props and Driver.VZ.props with the right years.

I'll think about it some more but I really don't like the current solution because it breaks the otherwise elegant separation of the vendor-specific props.

Also not super-excited about the for (;;) loop in Driver.Common.targets. I would prefer if the copyright string was concatenated by MSBuild and only one regex-replace pass was needed to massage the .inf files into their final form. Thanks!

ladipro: What do you think about https://github.com/ladipro/kvm-guest-drivers-windows/commit/e0d3cf4df058cb9909a5704e37c3b474912034df ?

We could also put all those RHEL_COPYRIGHT_STARTING_YEAR definitions in a separate file so it's not duplicated in Driver.RHEL.props and Driver.VZ.props.

This is the only issue I have with this patchset, everything else looks good.

AlekseyKostyushko:
Your patch elegantly fix inx copyright issue, but should be extended to propagate macros in *.ver to fix .rc too

AlekseyKostyushko:
I've figured that .ver doesn't requires any modifications. So if all is ok we'll add your patch to patchset and it can be applied?

pvpanic/PVPanic Package/pvpanic_pack.props
dlunev: looks good. I think that the best way is to get our patchset and apply your patch on top. This would be the simplest way.

ladipro: Yes, that's fine. Going to apply the patches now.

PR #110, opened 2017-03-15
vioscsi: Allocate only one virtqueue in dump mode

ladipro:
The driver currently computes max_queues using the configured
number of queues and maximum CPUs supported by the system, even
in dump mode. This results in attempts to allocate excessive
amount of memory and VioScsiFindAdapter failing.

Note that StorPortGetUncachedExtension can only allocate a
maximum of 64 kB on some Windows versions.

We know that we always use only one queue in dump mode. This
commit fixes the code accordingly.


ladipro:
Tracked in https://bugzilla.redhat.com/show_bug.cgi?id=1432567

PR #111, opened 2017-03-16
Rename NetKVMTemporaryCert to VirtIOTestCert

ladipro:
The certificate is useful for testing all drivers, for example when
deploying them using VS. There is no reason it should be NetKVM
specific. Also note the generic cert name (with a typo):
"Virto test certificate"

This commit renames NetKVMTemporaryCert.* to VirtIOTestCert.*
in Tools and deletes identical copies in NetKVM/tools.


YanVugenfirer:
One thing to check - is the certificate used by NDIS5 code?

ladipro:
NetKVM/NDIS5/tools/signing.cmd has these commented lines:

```
:create
rem This section should not be used
rem set _CERTSTORE_=NetKVMTemporaryCertStore
rem certmgr -del -all -s %_CERTSTORE_%
rem Makecert -r -pe -ss %_CERTSTORE_% -n "CN=NetKVMTemporaryCert" NetKVMTemporaryCert.cer
goto :eof

```

but the cert files are not used anywhere. I have built NetKVM for all targets with this change to be sure.

PR #112, opened 2017-03-16
Move {RHEL,VZ}_COPYRIGHT_STARTING_YEAR out of driver props files

ladipro:
Defines a new CopyrightStrings property which is then used in
vendor specific files to compose copyright strings. This way
a new vendor FOO will define their copyright string policy in
Driver.FOO.props
foo.ver
etc.

without touching driver props files.

Keeping vendor specific properties in their separate files makes
introducing a new vendor a purely additive process, without the
need to edit existing files. The vendor may even keep their files
"downstream only" if they choose so.


ladipro:
Reviewed by @dlunev in https://github.com/virtio-win/kvm-guest-drivers-windows/pull/108#discussion_r106418239

PR #123, opened 2017-04-03
virtiolib: Remember interrupt message numbers

ladipro:
This is a workaround for a WDF bug, last seen in version 1.9 and
confirmed fixed in 1.11. On resource rebalance WDF does not
preserve the MessageNumber field of its internal data structures
describing interrupts. As a result, we fail to report the right
MSI message number to the virtio device when re-initializing it
and it may stop working.

This manifests as a hang while trying to write to a virtual serial
port after a CPU is hotplugged to Windows 2008 R2, for example.

The fix is to keep the message number in a separate context
structure, attached to the interrupt object the first time it is
passed to us. Getting the message number can be subsequently
satisfied from this context, rather than from the buggy
WdfInterruptGetInfo call.

---

The first commit is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1435778

The second commit fixes a minor issue found when working on the bug.
It is not related to BZ #1435778 other than by also touching interrupt context.

### DIFF
--- a/VirtIO/WDF/PCI.c
+++ b/VirtIO/WDF/PCI.c
@@ -120,18 +120,53 @@ int PCIReadConfig(PVIRTIO_WDF_DRIVER pWdfDriver,
     return (read == length ? 0 : -1);
 }
 
+NTSTATUS PCIRegisterInterrupt(WDFINTERRUPT Interrupt)
+{
+    PVIRTIO_WDF_INTERRUPT_CONTEXT context;
+    WDF_OBJECT_ATTRIBUTES attributes;
+    NTSTATUS status;
+
+    if (Interrupt == NULL) {
+        status = STATUS_SUCCESS;
+    } else {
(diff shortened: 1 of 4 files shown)


ladipro:
@hammerg, can you please take a look?

hammerg:
Other than the Hungarian notation naming I'm okay with these changes.

@ladipro Do you have a reference to the WDF bug? Thanks.

ladipro:
Hungarian is used in all of VirtIO/WDF. Is there a good reason for not using it? For what it's worth, VS2015 project templates still use the Hungarian notation also:

```
int APIENTRY wWinMain(_In_ HINSTANCE hInstance,
                     _In_opt_ HINSTANCE hPrevInstance,
                     _In_ LPWSTR    lpCmdLine,
                     _In_ int       nCmdShow)
{
  ...
```

The WDF bug has not been officially documented, as far as I know. The best we have is the binary diff between version 1.9 and 1.11 of FxInterrupt::RevokeResources where the code to preserve m_InterruptInfo.MessageNumber was added. Thanks!

PR #137, opened 2017-05-15
BZ#1393772: Add null driver for SM Bus Controller

sameehj:
The device 8086:2930 has no built in driver in Windows 2008 and Win7 on
both x86 and x64 architectures, this commit adds a suitable null driver
for this device.

Signed-off-by: Sameeh Jubran <sjubran@redhat.com>
Signed-off-by: Daniel Rempel <daniel@daynix.com>

### DIFF
--- a/Q35/SMBus/smbus.inf
+++ b/Q35/SMBus/smbus.inf
@@ -1,0 +1,47 @@
+;-------------------------------------------------------------------------------
+; smbusq35.INF
+;
+; Red Hat SM Bus Q35
+;
+; Copyright (c) 2017 Red Hat, Inc.  All rights reserved.
+;
+;-------------------------------------------------------------------------------
+
+[Version]
+Signature = "$Windows NT$"
+Class = System
+ClassGuid = {4D36E97D-E325-11CE-BFC1-08002BE10318}
+Provider = %RedHat%
+DriverVer=04/27/2017,100.0.0.0
+DriverPackageType = PlugAndPlay
+DriverPackageDisplayName = %smbus.DeviceDesc%
+CatalogFile  = smbus.cat
+
+; ======
+; Following two lines are questionable works without them
+DriverPackageType = PlugAndPlay
+DriverPackageDisplayName = %smbus.DeviceDesc%
+
+[Manufacturer]
+%smbus.DeviceDesc% = Models, NTamd64
+
+[Models]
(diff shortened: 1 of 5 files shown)


Q35/SMBus/smbus.inf
```diff
@@ -0,0 +1,47 @@
...
+DriverPackageDisplayName = %smbus.DeviceDesc%
+CatalogFile  = smbus.cat
+
+; ======
+; Following two lines are questionable works without them
+DriverPackageType = PlugAndPlay
+DriverPackageDisplayName = %smbus.DeviceDesc%
+
+[Manufacturer]
+%smbus.DeviceDesc% = Models, NTamd64
```
ladipro: Does this .inf really work on 32-bit Windows? I would have expected

`%smbus.DeviceDesc% = Models,NTx86,NTamd64`

here and a

`[Models.NTx86]`

section below. Thanks!

sameehj: Yes it does, according to [this](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/inf-models-section):
> These platform extensions are not required in INF files for x86-based target operating system versions or non-PnP driver INF files.

However I can add it if you believe it is necessary.

sameehj: I have retested it again and it works on x86.

ladipro: Thank you! I didn't know that the x86 section is not required.

Q35/SMBus/smbus.inf
```diff
@@ -0,0 +1,47 @@
...
+
+[Models]
+%smbus.DeviceDesc% = NullInstallSection, PCI\VEN_8086&DEV_2930&SUBSYS_11001AF4&REV_02
```
ladipro: Hmm, does this line do anything if we also bind to `PCI\VEN_8086&DEV_2930&SUBSYS_11001AF4`, without the REV?

Q35/SMBus/smbus.inf
```diff
@@ -0,0 +1,47 @@
+;-------------------------------------------------------------------------------
+; smbusq35.INF
```
ladipro: nit: Doesn't match the file name.

sameehj: Thanks for noticing that I'll update it =)

Q35/SMBus/smbus.inf
sameehj: Nope, we can drop it.

PR #139, opened 2017-05-16
fix for BZ#1442322 - Cannot enlarge/shrink disk with virtio-blk-pci d…

vrozenfe:
…evice.

### DIFF
--- a/Tools/Driver.Common.targets
+++ b/Tools/Driver.Common.targets
@@ -103,8 +103,8 @@ Invoking the task:
   <!-- Stampinf (for Legacy DDK) -->
   <Target Condition="'$(Feature_LegacyStampInf)'=='true' AND '$(UseLegacyDDK)'=='true'" Name="StampInf_LegacyDDK" AfterTargets="AdjustInf_LegacyDDK" BeforeTargets="PackOneTarget">
     <Message Text="StampInf_LegacyDDK: for $(OutDir)$(TargetName).inf with Feature_UsingWDF=$(Feature_UsingWDF) and STAMPINF_VERSION=$(STAMPINF_VERSION)" Importance="high"/>
-    <Exec Condition="'$(Feature_UsingWDF)'=='false'" Command="stampinf -f $(OutDir)$(TargetName).inf -a $(InfArch) -v $(STAMPINF_VERSION)"/>
-    <Exec Condition="'$(Feature_UsingWDF)'=='true'" Command="stampinf -f $(OutDir)$(TargetName).inf -a $(InfArch) -k 1.9 -v $(STAMPINF_VERSION)"/>
+    <Exec Condition="'$(Feature_UsingWDF)'=='false'" Command="stampinf -f $(OutDir)$(TargetName).inf -a $(InfArch) -v $(STAMPINF_VERSION) -d *"/>
+    <Exec Condition="'$(Feature_UsingWDF)'=='true'" Command="stampinf -f $(OutDir)$(TargetName).inf -a $(InfArch) -k 1.9 -v $(STAMPINF_VERSION) -d *"/>
   </Target>
 
   <!-- Imports -->


Balloon/app/memstat.cpp
```diff
@@ -202,8 +205,25 @@ BOOL CMemStat::Update()
...
+                             L"CacheBytes",
+                             0,
+                             &var_val,
+                             NULL,
+                             NULL
+                             );
+
+        if (FAILED(status) || (var_val.vt == VT_NULL)) {
+            PrintMessage("Cannot get CacheBytes");
+            var_val.vt = 0;
```
ladipro: nit: It would be cleaner to call Clear() on the variant.

Assigning -1 to vt is not correct though -- pre-existing issue in several places in this function.

```
    _variant_t var_val(123ULL);
    var_val.vt = -1;
    printf("%llu\n", (ULONGLONG)var_val);

Unhandled exception at 0x7767DBE8 in ConsoleApplication1.exe: Microsoft C++ exception: _com_error at memory location 0x008FF744.
```

viostor/virtio_pci.c
```diff
@@ -197,12 +197,12 @@ static u16 vdev_get_msix_vector(void *context, int queue)
             } else {
                 vector = queue + 1;
             }
-        } else {
-            vector = VIRTIO_MSI_NO_VECTOR;
         }
     } else {
         /* on-device-config-change interrupt */
-        vector = VIRTIO_MSI_NO_VECTOR;
+        if (!adaptExt->msix_one_vector) {
```
vrozenfe: please seem my comment at https://bugzilla.redhat.com/show_bug.cgi?id=1442322#c13
However, I can be completely wrong, and if you think that you know how to deliver two different messages using one vector, please share your thoughts.

Balloon/app/memstat.cpp
vrozenfe: Why?
Did you miss CWebObject::Get call in the middle, which calls _VariantInit internally?

viostor/virtio_pci.c
ladipro: If the device has only one vector and the driver requires more, then the driver should fail to load.

The current state of viostor is that it loads, pretends to support one vector (the msix_one_vector flag which you added recently), but doesn't function properly. That's certainly a bug.

The fix is either to have the driver fail to load if it doesn't have enough vectors, or fix the code to be able to drive a one-vector device. I don't see why it wouldn't be possible - you just have to look at all sources of interrupts when you receive the one message.

vrozenfe: Do you think that failing driver to load (especially when it provides 99.99% of its functionality but fails to handle configuration changed request which probably never happens in its life) will be a good solution?

In the real life, failing to report the correct number of vectors  vector is a HW firmware bug, which makes a driver to deal with a broken hardware. Keeping a driver functional in this situation is a big win, but the HW has to be fixed ASAP. Do you have any particular scenario in your mind when a user will prefer using one vector, rater that specifying number of vectors equal to number of queues + 1, or switching to IRQ mode?

As I  asked already, do you really think that failing a boot-time driver because it will not be able to handle one minor feature is what we should to do?  Regarding to looking at source of interrupt, you probably mean reading the ISR status register. Am I right? If so, who can guarantee that ISR status register will provide us with the correct information now and in the future? And even more, IO operations in MSI service routine looks like a bit strange solution to me.

Balloon/app/memstat.cpp
ladipro: We do var_val.vt = -1 if Get fails, don't we? This is the problematic pattern:

```
         if (FAILED(status) || (var_val.vt == VT_NULL)) {
             PrintMessage("Cannot get PageReadsPerSec");
             var_val.vt =  -1;
         }
         m_Stats[idx].tag = VIRTIO_BALLOON_S_MAJFLT;
         m_Stats[idx].val = (long)var_val;
         idx++;
```

viostor/virtio_pci.c
ladipro: My preferred solution is to fix the code to be fully functional with any number of vectors that the QEMU device supports. Look at what Linux virtio driver do in vp_find_vqs. They first try to allocate one vector per queue + one for config. Failing that, they try one for all queues and one for config (2 total). And if even that doesn't work they fall back to IRQ. Not sure if we can do the IRQ fallback on Windows but at least the fallback to 2 vectors is doable.

With the code proposed in this patch you're going to disable config interrupts if the number of vectors is less than number of queues + 1. That's not necessary.

This condition in VirtIoHwInitialize should be refined:

```
    if (adaptExt->msix_vectors >= (adaptExt->num_queues + 1U)) {
        /* initialize queues with a MSI vector per queue */
        RhelDbgPrint(TRACE_LEVEL_INFORMATION, ("Using a unique MSI vector per queue\n"));
        adaptExt->msix_one_vector = FALSE;
    } else {
        /* if we don't have enough vectors, use one for all queues */
        RhelDbgPrint(TRACE_LEVEL_INFORMATION, ("Using one MSI vector for all queues\n"));
        adaptExt->msix_one_vector = TRUE;
    }

```
To distinguish between "at least 2 but less than num_queues + 1" and "1" and disable config interrupts only in the latter case.

Also, looking at the code you're changing in VirtIoMSInterruptRoutine, how are we supposed to handle all queues with one vector? You just assign 1 to MessageID. There needs to be a loop somewhere. See how VioScsiMSInterrupt does it.

I agree that trying to handle everything with one vector wouldn't work well with viostor. We can't even use the ISR status register in MSI mode so we would have to read the entire configuration on each interrupt to see if anything changed - that's way too suboptimal, even for "limp mode". Running with no config interrupts if there is no other choice is reasonable. We could poll the config periodically but I'd leave that to the over-engineering department :)

If you believe that reporting one vector is a HW bug, then we should submit a QEMU patch to fix it.

ladipro: Ah, so viostor requires adequate number of vectors for multiqueue operation. This is what I was missing:

```
        ((adaptExt->num_queues + 1U) > adaptExt->msix_vectors)) {
        //FIXME
        adaptExt->num_queues = 1;
    }

```

Please ignore my last comment then, the change is good. Thanks!

vrozenfe: Thanks,
speaking about fallback  from MSI to ISR, it is not supported in Windows as far as I remember.
What we can try doing in the future is intercepting StorQueryResourceRequirements/StorFilterResourceReqirements calls and see if we can ask for allocating one vector more.

Balloon/app/memstat.cpp
vrozenfe: Ah, I see your point now.
You mean trying to use varint after marking vt as illegal. 
You  are right, we can get there if someone decided to call Get
with wrong parameters, like incorrect name of property, when GetSystemPropertyByName
call will fail, or any other random crap instead of third parameter . Fix to this potential problem can (and probably should) be submitted as a separate patch.

ladipro: Yes, definitely a separate patch. Can you please update commit messages in this pull request (short description + Signed-off-by) so we can merge this. Thanks!

ybendito:
As far as I remember, there is a guideline of supporting MSI(X) that the
device shall function when it receives expected number of vectors and also
when it receives just one vector.
As far as I remember, again, this happens on one of x86 operating systems
and some time ago we fixed the NetKVM for that.
Need to dig in BZs to find exactly where it happened, but this was a case.

vrozenfe:
Do you mean Dynamically Configuring MSI-X https://msdn.microsoft.com/en-us/library/windows/hardware/ff544219(v=vs.85).aspx ?
If yes, then  as I said before, proper handling of  StorQueryResourceRequirements/StorFilterResourceReqirements  request is probably what needs to be added in one of the following releases.

ybendito:
On Sun, May 21, 2017 at 3:05 PM, Vadim Rozenfeld <notifications@github.com>
wrote:

> Do you mean Dynamically Configuring MSI-X https://msdn.microsoft.com/en-
> us/library/windows/hardware/ff544219(v=vs.85).aspx ?
> If yes, then as I said before, proper handling of
> StorQueryResourceRequirements/StorFilterResourceReqirements request is
> probably what needs to be added in one of the following releases.
>
> No, I did not mean that.
When Windows does not succeed to provide all the MSI vectors as the
hardware supports (for example, due to limitations of IRQ vectors), it
configures the device to work with just one MSI vector.

PR #145, opened 2017-06-08
Integration with Appveyor CI

dmitryfleytman:
This pull request introduces integration with Appveyor CI.

On each push and PR (including this one) the code will be build on Appveyor, build state notifications will be sent to virt-win-devel@redhat.com.

### DIFF
--- a/Tools/build.bat
+++ b/Tools/build.bat
@@ -92,7 +92,7 @@ if "%BUILD_SPEC%"=="" (
       )
       if not "!BUILD_SPEC!"=="" (
         call :build_target !BUILD_SPEC!
-        if not "!BUILD_FAILED!"=="" goto :eof
+        if not "!BUILD_FAILED!"=="" goto :fail
       )
     )
   )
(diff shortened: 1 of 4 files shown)


Tools/build.bat
```diff
@@ -76,8 +76,8 @@ for %%N in (%SUPPORTED_BUILD_SPECS%) do (
   )
 )
 
-rem Silently exit if the build target could not be matched
-goto :eof
+rem Return error if the build target could not be matched
+goto :fail
```
ladipro: The reason for ignoring build target mismatch are projects like NetKVM, viostor, and vioscsi, which build different sln/vcxproj for different targets. NetKVM for example:

```
@echo off
call ..\tools\build.bat netkvm-vs2015.sln "Wlh Win7 Win8 Win8.1 Win10" %*
if errorlevel 1 goto :eof
call ..\tools\build.bat NetKVM-VS2015.vcxproj "Win10_SDV" %*
if errorlevel 1 goto :eof
call ..\tools\build.bat NDIS5\NetKVM-NDIS5.sln "WXp Wnet" %*

```

With this change, I can no longer build NetKVM for Wnet by typing "buildall Wnet" in the NetKVM directory because the first build.bat invocation fails. And same with the top-level buildAll actually. Any chance we can still keep this working?

buildAll.bat
```diff
@@ -1,16 +1,22 @@
...
-  if errorlevel 1 goto :eof
+  if errorlevel 1 goto :fail
   popd
 )
+
+exit /B 0
+
+:fail
+
+exit /B 1
```
ladipro: Is the goal here to normalize all errorlevel >=1 to 1 or did it not propagate the error at all?

Tools/build.bat
dmitryfleytman: Hi Ladi, thanks for review!

It was not clear for me what was that code for. With your explanations this makes sense. Would omitting this chunk fully resolve the issue?

buildAll.bat
dmitryfleytman: It did not propagate at all.

Tools/build.bat
ladipro: Yes, omitting would fix it. Thanks!

dmitryfleytman: Cool! I will update the PR. I would also recommend adding a comment there to make it clear.

dmitryfleytman:
@ladipro PR updated. Please take a look.

PR #146, opened 2017-06-29
Balloon: Remove devCtx->StatWorkItem=NULL from D0Exit

dlunev:
When the device power state is changed, or during the hardware resources
rebalancing (occurs after cpu hotplug, for example), BalloonEvtDeviceD0Entry
can be called again after BalloonEvtDeviceD0Exit. In this case, guest will
BSOD after the first Isr Dpc. In addition StatWorkItem is created in
BalloonDeviceAdd and will be automatically deleted by the framework along
with the device, so setting StatWorkItem to NULL in BalloonEvtDeviceD0Entry
is simply incorrect.

About a year ago AlekseyKostyushko already made a similar patch, which also
synchronize Dpc activity with D0Entry / D0Exit. Our tests and framework code
analysis showed that wdf really waits the Dpc rundown before D0Exit call so
additional synchronization is not required.


ladipro:
Thank you!

PR #149, opened 2017-07-17
Fix incorrect condition for send led status

xuxi0513:
uPendingLedCode initial value is 0xFF, which means invalid code, and uShortPendingLedValue initial value is 0.
Using "uShortPendingLedValue != 0xFF" as condition to send kbd led status, it will send invalid code 0xFF and value 0 to backend.


ladipro:
I have squashed the two commits and pushed it to master as https://github.com/virtio-win/kvm-guest-drivers-windows/commit/a582ed09aa29138f6bd85028b7f4ff0865c0bf26
Thanks!

PR #166, opened 2017-09-18
viostor: Fix geometry updates with vectors=1

ladipro:
If the virtio-blk-pci device is configured with only one MSI vector,
the driver didn't refresh its geometry information on disk resize
because it's not easy to tell that the interrupt was triggered by
a config change.

Possible solutions include:
* checking the config generation field (virtio 1.0 only) to detect
  config changes
* checking the ISR status field (QEMU seems to always set bit 1 even
  if the device operates in MSI mode)
* re-reading the config in a rate-limited fashion (on every n-th
  interrupt, or every second, or something similar)

This commit fixes it by polling the config when we're asked for the
geometry info by the OS. RhelGetDiskGeometry just copies the info to
our internal structures anyway, so we're not losing much by not
doing it before Windows asks.

This fixes https://bugzilla.redhat.com/show_bug.cgi?id=1451572

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -803,7 +803,14 @@ VirtIoStartIo(
 
         case SCSIOP_READ_CAPACITY16:
         case SCSIOP_READ_CAPACITY: {
-            UCHAR SrbStatus = RhelScsiGetCapacity(DeviceExtension, (PSRB_TYPE)Srb);
+            UCHAR SrbStatus;
+            if (adaptExt->msix_one_vector) {
+                /* We don't have a separate MSI vector for VirtIO config changes so
+                 * we simply poll the disk geometry here. See VirtIoMSInterruptRoutine.
+                 */
+                RhelGetDiskGeometry(DeviceExtension);
+            }
+            SrbStatus = RhelScsiGetCapacity(DeviceExtension, (PSRB_TYPE)Srb);
             CompleteRequestWithStatus(DeviceExtension, (PSRB_TYPE)Srb, SrbStatus);
             return TRUE;
         }


vrozenfe:
Hi Ladi,
While your patch looks quite harmless, I would like to ask you keeping it on hold for 
awhile. The problem  is that entire rescan logic is not yet fully implemented in viostor 
driver. Upon triggering configuration change event the driver should notify port driver 
with BusChangeDetected message and initiate UNIT ATTENTION condition. The driver
should be able to initiate and handle the entire rescan sequence automatically. We used 
to have  a bug in bugzilla Bug 1009314 - [RFE][QEMU-GA][virtio-win] Add "scan-disks" capability 
following a disk resize. Unfortunately this problem failed under qemu-ga category which was not correct. So, refreshing internal structure is just a small part of the code that needs to be (re)implemented. Another big question is how configuration change event handler will be 
implemented in case of one MSI vector configuration. In Win8/Win10 we might be able to claim insufficient PCI resources during driver loading. For the previous versions we can add timer to check
configuration periodically (quite close to what Windows used to do for removable media long time ago,
and considered not being good for VM). However, taking into account that single MSI vector should be a quite rare case and complicity of workaround for handling this case properly,  I would leave it as is,
without fixing.

Best regards,
Vadim.
resources

ladipro:
Hi Vadim, no problem, we don't have to do this now. Let's reimplement the config change logic properly.

Another issue in the current code is the lack of synchronization, btw. RhelGetDiskGeometry doesn't do anything to ensure that the virtio_blk_config fields are consistent. We should use the virtio 1.0 config generation field when reading the config from the device and a spinlock to protect our internal structures.

vrozenfe:
Good 
I'm going to to add automatic resize logic in next two weeks or so, and then we can respin  this 
issue again.
Best,
Vadim.

PR #173, opened 2017-10-13
serial: remove RETRY_THRESHOLD limit from c_ovq poll loop

dlunev:
This patch is a fix for very annoying problem in the production. It was
very hard to understand what is going on and why. From the user point
of view the situation looks like

	dev: virtserialport, id "channel0"
	chardev = "charchannel0"
	nr = 1 (0x1)
	name = "org.qemu.guest_agent.0"
	port 1, guest off, host off, throttle off

and the device comes into this state under unknown circumstances.

We've found out that this problem is caused by a breakdown in the
control channel. The host has stopped responding to the control messages
at some moment. Sometimes channel could be fixed after qemu-qa service
restart, sometimes not. From the technical point of view there is the
following problem:

The amount of virtqueue_get_buf calls from VIOSerialSendCtrlMsg is
limited to RETRY_THRESHOLD times(400). Usually this is more than enough.
The amount of retries in the common case in 1-2. But unfortunately
sometimes host needs MUCH more time to handle message. Thus there is
a probability that the polling counter will exceeds RETRY_THRESHOLD.
In our production and load testing this probability fires 1-2 times
a week. Thus here loop ends, but the queue still contains the element
with the address on VIOSerialSendCtrlMsg stack. This buffer will become
invalid after the return from the function. Obviously QEMU will see the
garbage once it will get a chance to process such orphaned message.
It ignores it without guest notification and thus the vioserial is
stuck in some intermediate state.

We can't just remove RETRY_THRESHOLD limit. In this case there is a
risk of getting DPC_WATCHDOG bugcheck due to long spin wait on
DPC_LEVEL.

This patch splits in/out queues lock CVqLock on two, with WaitLock
COutVqLock being used for c_ovq. In this case virtqueue_get_buf poll
loop is executed on PASSIVE_LEVEL. The restriction to the amount of
retries is removed accordingly. WorkItem is used to low the current
IRQL where necessary.

Spinlock is better solution to synchronize such a small and fast code
and was rightfully used to lock c_ivq and c_ovq access. Thus it is kept
for CInVqLock to lock c_ivq. Using WaitLock for c_ovq leads to
insignificant decrease of the control message sending rate due to
walking through dispatcher every time on lock acquire/free and a possible
thread preemtion while running locked code, but reliability is more
important than control channel speed.

### DIFF
--- a/vioserial/sys/Control.c
+++ b/vioserial/sys/Control.c
@@ -43,23 +43,18 @@ VIOSerialSendCtrlMsg(
     sg.physAddr = MmGetPhysicalAddress(&cpkt);
     sg.length = sizeof(cpkt);
 
-    WdfSpinLockAcquire(pContext->CVqLock);
+    WdfWaitLockAcquire(pContext->COutVqLock, NULL);
     if(0 <= virtqueue_add_buf(vq, &sg, 1, 0, &cpkt, NULL, 0))
     {
         virtqueue_kick(vq);
         while(!virtqueue_get_buf(vq, &len))
         {
            KeStallExecutionProcessor(50);
-           if(++cnt > RETRY_THRESHOLD)
-           {
-              TraceEvents(TRACE_LEVEL_FATAL, DBG_PNP, "<-> %s retries = %d\n", __FUNCTION__, cnt);
-              break;
-           }
         }
     }
-    WdfSpinLockRelease(pContext->CVqLock);
+    WdfWaitLockRelease(pContext->COutVqLock);
 
-    TraceEvents(TRACE_LEVEL_INFORMATION, DBG_PNP, "<-- %s cnt = %d\n", __FUNCTION__, cnt);
+    TraceEvents(TRACE_LEVEL_INFORMATION, DBG_PNP, "<-- %s\n", __FUNCTION__);
 }
 
 VOID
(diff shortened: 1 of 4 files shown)


vioserial/sys/Control.c
```diff
@@ -43,23 +43,18 @@ VIOSerialSendCtrlMsg(
...
     sg.length = sizeof(cpkt);
 
-    WdfSpinLockAcquire(pContext->CVqLock);
+    WdfWaitLockAcquire(pContext->COutVqLock, NULL);
     if(0 <= virtqueue_add_buf(vq, &sg, 1, 0, &cpkt, NULL, 0))
     {
         virtqueue_kick(vq);
         while(!virtqueue_get_buf(vq, &len))
         {
            KeStallExecutionProcessor(50);
```
ladipro: I wonder if we can use KeDelayExecutionThread now that the loop executes at passive IRQL.

vioserial/sys/Control.c
```diff
@@ -43,23 +43,18 @@ VIOSerialSendCtrlMsg(
...
-           {
-              TraceEvents(TRACE_LEVEL_FATAL, DBG_PNP, "<-> %s retries = %d\n", __FUNCTION__, cnt);
-              break;
-           }
         }
     }
-    WdfSpinLockRelease(pContext->CVqLock);
+    WdfWaitLockRelease(pContext->COutVqLock);
 
-    TraceEvents(TRACE_LEVEL_INFORMATION, DBG_PNP, "<-- %s cnt = %d\n", __FUNCTION__, cnt);
```
ladipro: nit: cnt is now unused

`int cnt = 0;`

at the top of the function can be deleted.

irudakov77: My fault, forgot to delete

vioserial/sys/Control.c
irudakov77: It's not a crime to call KeStallExecutionProcessor on PASSIVE_LEVEL for small delay. But I left it there not for delay but for calling pause instruction, because Windows compiler has no intrinsic for it. Now I think using KeDelayExecutionThread with 0 time interval would be a better solution. Thanks, Ladi!

PR #175, opened 2017-10-23
[ivshmem] correct various issues and nits pointed out by ladipro

gnif:
This corrects various issues with the driver in it's existing state, but as @ladipro has suggested there is a better way to handle the interrupt events which will be implemented when I have more time.


ivshmem/Device.c
```diff
@@ -90,8 +90,8 @@ NTSTATUS IVSHMEMEvtDevicePrepareHardware(_In_ WDFDEVICE Device, _In_ WDFCMRESLIS
 
 	if (deviceContext->interruptCount > 0)
 	{
-		deviceContext->interrupts = (WDFINTERRUPT*)MmAllocateNonCachedMemory(
-			sizeof(WDFINTERRUPT) * deviceContext->interruptCount);
+		deviceContext->interrupts = (WDFINTERRUPT*)ExAllocatePoolWithTag(NonPagedPool,
```
ladipro: Also need to fix the free side: ExFreePoolWithTag instead of MmFreeNonCachedMemory.

gnif: thanks, done.

gnif:
@ladipro I am in rather urgent need of this driver being built and signed even if just for private usage at the moment, is there any chance of this being possible? I am able to offer some (very limited) funds if that would help get this pushed through sooner. I need to be able to run my development environment outside of test-signing mode as the application I am interfacing with contains anti-piracy measures which includes refusal to operate if windows is in test signing mode.

vrozenfe:
at the moment we don't use the upstream repository to build our publicly
available drivers.
however we might be interested in starting this process in nearest future.
Give us a couple
of days to see what can be done toward solving this issue.

Best regards,
Vadim.

gnif:
@vrozenfe Thanks for the information. I removed my comment when I realized I could purchase a cheap enough certificate to sign a private copy with. This is no longer an urgent matter.

vrozenfe:
just make sure that you are buying one that can be cross certified ( https://docs.microsoft.com/en-us/windows-hardware/drivers/install/cross-certificates-for-kernel-mode-code-signing ) otherwice try checking reactos project, iirc they used to provide kernel binary sighing for free.

YanVugenfirer:
Signing with Visual Studio: https://docs.microsoft.com/en-us/windows-hardware/drivers/develop/signing-a-driver-for-public-release

Signing with Visual Studio will be an easier option if you are new to Windows drivers development. Pay attention that the changes are saved in .user file of the project and *.user is in .gitignore - so don't miss it if you are committing changes to your local git repo. 

You can achieve the same using signtool.
Signing with signtool: https://msdn.microsoft.com/en-us/library/windows/desktop/aa387764(v=vs.85).aspx

PR #176, opened 2017-10-26
serial: Minor corrections according to ladipro comments

irudakov77:
1. Counter variable (int cnt) is now unused and should be deleted.

2. Using KeDelayExecutionThread with zero timeout interval cause current thread
to relinquish the remainder of its time slice to other thread in most cases.
This allows to make small delay on PASSIVE_LEVEL and optimize CPU usage.
It looks slightly more efficient then using KeDelayExecutionThread which is
actually a busy loop.

### DIFF
--- a/vioserial/sys/Control.c
+++ b/vioserial/sys/Control.c
@@ -49,7 +49,8 @@ VIOSerialSendCtrlMsg(
         virtqueue_kick(vq);
         while(!virtqueue_get_buf(vq, &len))
         {
-           KeStallExecutionProcessor(50);
+            LARGE_INTEGER interval = {0};
+            KeDelayExecutionThread(KernelMode, FALSE, &interval);
         }
     }
     WdfWaitLockRelease(pContext->COutVqLock);
--- a/vioserial/sys/Control.c
+++ b/vioserial/sys/Control.c
@@ -26,7 +26,7 @@ VIOSerialSendCtrlMsg(
     UINT len;
     PPORTS_DEVICE pContext = GetPortsDevice(Device);
     VIRTIO_CONSOLE_CONTROL cpkt;
-    int cnt = 0;
+
     if (!pContext->isHostMultiport)
     {
         return;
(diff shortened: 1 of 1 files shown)


vioserial/sys/Control.c
```diff
@@ -49,7 +49,8 @@ VIOSerialSendCtrlMsg(
         virtqueue_kick(vq);
         while(!virtqueue_get_buf(vq, &len))
         {
-           KeStallExecutionProcessor(50);
+            LARGE_INTEGER interval = {0};
```
ladipro: It looks like passing 0 for interval is not really documented as it's neither positive nor negative. Would it be maybe cleaner to use -1?

YanVugenfirer:
I found the original BZ that triggered the addition of the problematic code: https://bugzilla.redhat.com/show_bug.cgi?id=626333

It is worth running the "Sleep Stress with IO" and "Common scenario stress with IO" tests to check that the solution is complete.

vrozenfe:
I would suggest testing on both, SMP and UP setups as well.
Best,
Vadim.

vioserial/sys/Control.c
irudakov77: I agree. Zero interval just makes KeDelayExecutionThread to call NtYieldExecution with minimal additional actions, but it not obvious, and not documented.

ladipro:
Thanks, looks good. Is it easy for you to run the tests as suggested by @YanVugenfirer and @vrozenfe ?

irudakov77:
It is not easy right now, we'll do it at the middle of the next week, I think.

20lives:
We ran HCK tests on Windows 2008R2 for virtio-serail device for both single CPU and SMP HCK clients and didn't find any issues.
Logs can be found at: 
https://www.dropbox.com/s/osykjs3nhxxvo1f/virtioserial-multicpu.hckx?dl=0
https://www.dropbox.com/s/femesfi48hrn7u1/virtioserial-singlecpu.hckx?dl=0

PR #178, opened 2017-10-31
[ivshmem] fixed infiniate loop bug

gnif:
If a second application opens and then closes the device handle the cleanup will enter an infinite loop. This patch fixes this.


gnif:
Ah, sorry about that, done :)

PR #179, opened 2017-10-31
SDV ivshmem fixes

ladipro:
A few fixes to try to make the SDV build succeed. It has found real code issues.

I still haven't been able to make SDV generate the DVL file though. Running

`tools\build.bat ivshmem\ivshmem.vcxproj "Win10_SDV"`

fails with

```
"C:\kvm-guest-drivers-windows\ivshmem\ivshmem.vcxproj" (sdv target) (1) ->
(sdv target) ->
  C:\Program Files (x86)\Windows Kits\10\build\windowsdriver.Sdv.targets(132,9): error MSB3073: The command "staticdv /devenv /check" exited with code -1. [C:\kvm-guest-drivers-windows\ivshmem\ivshmem.vcxproj]

    0 Warning(s)
    1 Error(s)

```

and it seems to be caused by one of the tools looking for sdv-harness.c.rawcfgf in ivshmem\sdv\check\purekmdfdriver instead of ivshmem\sdv where another tool puts it. So very likely a WDK bug. @gnif, do you see the same thing? Thanks!

### DIFF
--- a/buildAll.bat
+++ b/buildAll.bat
@@ -9,8 +9,6 @@ call tools\build.bat viostor\viostor.vcxproj "Win8_SDV Win10_SDV" %*
 if errorlevel 1 goto :fail
 call tools\build.bat ivshmem\ivshmem.vcxproj "Win10_SDV" %*
 if errorlevel 1 goto :fail
-call tools\build.bat ivshmem\test\ivshmem-test.vcxproj "Win10_SDV" %*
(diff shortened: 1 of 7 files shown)


gnif:
@ladipro I have been unsuccessful in getting SDV to work at all on my dev system, I used the output from AppVeyor to identify the first set of SDV issues I corrected.

PR #180, opened 2017-11-01
Kick ring

sameehj:


sameehj:
https://github.com/virtio-win/kvm-guest-drivers-windows/blob/e735fd25e6ef138e343264772b62676d55d67cc1/NetKVM/Common/ParaNdis-TX.cpp#L532

vbusireddy:
Got it. Thanks!

PR #184, opened 2017-11-05
Inf2cat: Use /uselocaltime when compiling

sameehj:
The inf2cat tool uses the UTC timezone by default for timestamp comparison,
this can cause the following error when there is a mismatch between time on
the build machine is and the one in UTC timezone. Using /uselocaltime
parameter solves this issue.

Inf2cat error:

* DriverVer set to incorrect date (postdated DriverVer not allowed) in
\netkvm.inf. The current date (UTC) is 01/01/2017.

### DIFF
--- a/Balloon/sys/balloon.vcxproj
+++ b/Balloon/sys/balloon.vcxproj
@@ -177,20 +177,24 @@
     <TargetName>balloon</TargetName>
     <OutDir>objfre_win8_x86\i386\</OutDir>
     <IntDir>objfre_win8_x86\i386\</IntDir>
+    <Inf2CatUseLocalTime>true</Inf2CatUseLocalTime>
   </PropertyGroup>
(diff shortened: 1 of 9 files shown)


YanVugenfirer:
There are also several batch files that build INF for platform devices without drivers or with MS supplied drivers that need to be changed as well:

1. https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/Q35/buildAll.bat
2. https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/fwcfg/buildAll.bat
3. https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/pciserial/buildAll.bat

ladipro:
Adding

`  <Inf2CatUseLocalTime>true</Inf2CatUseLocalTime>
`
to Tools/Driver.Common.props should have the same effect as adding it to every project file.

PR #189, opened 2017-11-08
Update README and parameterize PCI subsystem vendor ID

ladipro:
It is important that all parties shipping virtio-win drivers use their
own vendor-specific HW ID to avoid driving somebody else's device.
This is especially critical for WHQL'ed drivers distributed with Windows
Update, as these could be installed without any explicit user action.

A new build warning is introduced to make this requirement more
discoverable (suggested by @dlunev, thanks!)

Note that NetKVM has not been plugged into the new build infrastructure
yet so it is not handled by these commits.

### DIFF
--- a/Balloon/sys/balloon.inx
+++ b/Balloon/sys/balloon.inx
@@ -9,8 +9,8 @@
 ;Abstract:
 ;
 ;Installation Notes: 
-;    Using Devcon: Type "devcon install BALLOON.inf PCI\VEN_1AF4&DEV_1002&SUBSYS_00051AF4&REV_00" or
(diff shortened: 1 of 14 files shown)


YanVugenfirer:
>>> Note that NetKVM has not been plugged into the new build infrastructure
yet so it is not handled by these commits.

Also, ivshmem is missing. Currently, it uses Red Hat vendor and device ID.

ladipro:
> Also, ivshmem is missing. Currently, it uses Red Hat vendor and device ID.

Good point. Ivshmem is new and experimental but that makes it even more important for the HW ID to be as specific as possible. I'll add a commit to address this, thanks!

PR #194, opened 2017-11-10
NetKVM: NDIS5 driver support for multiple vendors

irudakov77:
Support for SUBSYSTEM_VENDOR_ID, vendor and copyright strings substitution for .inf and resources.

Drivers files packing was performed by additional tools, being executed during PostBuildStep. Now build doesn't run PostBuildStep, but uses PackOneTarget. Because this driver supports only Wnet/Wxp common.inf.h is not required and should be merged with netkvm.inx. Packing tools become unnecessary and are removed.


YanVugenfirer:
Hi Ilya,

It looks like the commits broke debug build for NDIS5 drivers.

Please take a look what happens when you run "buildall.bat debug"

Thanks,
Yan.

irudakov77:
Hi,
> Please take a look what happens when you run "buildall.bat debug"
Do you mean PackOne task errors? The same happens for ivshmem and every other project having Debug configs and using PackOne. When we make clean debug build, we get the errors, but we have successful debug build after release one. PackOne uses hardcoded path to release output directory regardless of current project config, thus task fails on debug build if it can't find release modules. I can fix PackOne task.
By the way, some projects don't have Debug configs, I can fix it too.

YanVugenfirer:
Hi,

In this case, at the end of the build - XP and Wnet directories are missing from Intall_Debug directory.

Best regards,
Yan.

irudakov77:
It's a packone issue too, Install directory name is hardcoded, I can change packone, or rollback latest commit.

Regards,
Ilya

YanVugenfirer:
Please change packone.

Thanks!

PR #196, opened 2017-11-19
NetKVM: Use NdisGetSharedDataAlignment to align miniport's adapter context

sameehj:
Aligning shared data can affect performance as it might reduce cache
misses which result in performance gain.

According to msdn documentation NdisGetSharedDataAlignment returns boundary
value, in bytes, on which drivers should align structures that can be shared
by more than one processor.


ladipro:
In a similar vein, pContext->pPathBundles should be padded so that no two bundles share a cache line.

PR #198, opened 2017-11-19
Fix bugs in netkvm and Virtio headers

sameehj:


VirtIO/VirtIORing.c
```diff
@@ -168,7 +168,7 @@ void *virtqueue_get_buf(
     put_unused_desc_chain(vq, idx);
 
     vq->last_used++;
-    if (virtqueue_is_interrupt_enabled(vq)) {
+    if (vq->event_suppression_enabled) {
```
ladipro: Hmm.. `event_suppression_enabled` is about notifying the device (i.e. our kicks), not the device notifying us (i.e. interrupts). It's not clear how kick suppression is related to the used ring.

The old code had the same condition with a comment saying "If we expect an interrupt for the next entry, tell host by writing event index".

sameehj: According to vrtio spec:
> 2.4.7.1 Driver Requirements: Virtqueue Interrupt Suppression
If the VIRTIO_F_EVENT_IDX feature bit is not negotiated:
• The driver MUST set flags to 0 or 1.
• The driver MAY set flags to 1 to advise the device that interrupts are not needed.
Otherwise, if the VIRTIO_F_EVENT_IDX feature bit is negotiated:
• The driver MUST set flags to 0.
• The driver MAY use used_event to advise the device that interrupts are unnecessary until the device
writes entry with an index specified by used_event into the used ring (equivalently, until idx in the used
ring will reach the value used_event + 1).
The driver MUST handle spurious interrupts from the device.

VirtIO/VirtIORing.c
```diff
@@ -233,7 +236,10 @@ bool virtqueue_enable_cb_delayed(struct virtqueue *vq)
 
     if (!virtqueue_is_interrupt_enabled(vq)) {
         vq->master_vring_avail.flags &= ~VIRTQ_AVAIL_F_NO_INTERRUPT;
-        vq->vring.avail->flags &= ~VIRTQ_AVAIL_F_NO_INTERRUPT;
+        if (!vq->event_suppression_enabled)
```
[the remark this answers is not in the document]
sameehj: if the VIRTIO_F_EVENT_IDX feature bit is negotiated:
• The driver MUST set flags to 0.

I kept the shadow copy for reference in case we need to debug this. However thinking about it again maybe I should update both of them.

VirtIO/VirtIORing.c
ladipro: > As fas as I understand this is for both ways Check out section 2.4.9 - Virtqueue Notification Suppression - which discusses the kicks that the driver send to the device.
>
> When VIRTIO_F_EVENT_IDX both of these mechanisms should be working and the flags shouldn't be used at all.

Ah, I understand now, thanks!

So I think this if should have both conditions now:

```if (vq->event_suppression_enabled && virtqueue_is_interrupt_enabled(vq)) {```

Otherwise we wouldn't really be able to disable device->driver notifications with VIRTIO_F_EVENT_IDX.

File: VirtIO/VirtIORing.c (hunk 2 above)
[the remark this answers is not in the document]
Comment (ladipro) [reply VirtIO/VirtIORing.c]: The idea with keeping the shadow copy for virtqueue_is_interrupt_enabled() is fine. I just failed to understand the intent of the code.

Maybe something like:

```
static inline set_avail_flags(struct virtqueue *vq, u16 flags)
{
    if (!vq->event_suppression_enabled) {
        vq->vring.avail->flags = flags;
    }
}
```

and call it from all three places? Thanks!

VirtIO/VirtIORing.c
sameehj: >if (vq->event_suppression_enabled && virtqueue_is_interrupt_enabled(vq)) {
Otherwise we wouldn't really be able to disable device->driver notifications with VIRTIO_F_EVENT_IDX.

Seems reasonable and should do the job, thanks!

sameehj:
In case you haven't notice, I have updated the pull request.

ladipro:
I wonder if VirtioLib shouldn't be handling VIRTIO_RING_F_EVENT_IDX internally. It's a transport feature flag so drivers shouldn't have to worry about it.

But that's a separate piece of work, this looks good, thank you!

PR #199, opened 2017-11-20
ivshmem: Don't delete WDFINTERRUPTs

ladipro:
WDFINTERRUPT objects are deleted by the framework. Explicitly calling
WdfObjectDelete results in an assert/BSOD when running with Driver
Verifier enabled.

---

Note that the **Creating an Interrupt Object** MSDN article at https://docs.microsoft.com/en-us/windows-hardware/drivers/wdf/creating-an-interrupt-object is wrong. WdfObjectDelete must not be called on interrupts:

https://github.com/Microsoft/Windows-Driver-Frameworks/blob/master/src/framework/shared/irphandlers/pnp/interruptobject.cpp#L234

cc @gnif

### DIFF
--- a/ivshmem/Device.c
+++ b/ivshmem/Device.c
@@ -220,8 +220,7 @@ NTSTATUS IVSHMEMEvtDeviceReleaseHardware(_In_ WDFDEVICE Device, _In_ WDFCMRESLIS
 
     if (deviceContext->interrupts)
     {
-        for (int i = 0; i < deviceContext->interruptsUsed; ++i)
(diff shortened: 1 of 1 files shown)


ladipro:
I've run a few HLK tests on WS2016 with this fix and all have passed.

```
  DF - PNP (disable and enable) with IO Before and After (Bring Up)
  DF - PNP Cancel Remove Device Test (Development and Integration)
  DF - PNP Cancel Stop Device Test (Development and Integration)
  DF - PNP DIF Remove Device Test (Development and Integration)
  DF - PNP Disable And Enable Device Test (Development and Integration)

```

PR #200, opened 2017-11-26
Work

vrozenfe:


vrozenfe:
Thanks for reviewing.
Vadim.

ladipro:
RhelGetDiskGeometry is still prone to reading inconsistent config values because of the lack of synchronization with the host side. More of a hypothetical issue but probably worth tracking in an issue/BZ.

Virtio 1.0 provides a "config generation" concept to address this. Legacy virtio can work around it by reading the config repeatedly until the reads match.

Issue #195, opened 2017-11-19
[virtio-serial] qemu exit after hot plug vcpu on Win2016


zyb521:
Hi everyone,

We found that in Windows server 2016, after the hot plug vcpu，qemu process directly exit some times， I'm not sure if it affects other Windows systems.

By preliminary analysis, is determined to be caused by the virtio-serial driver, our current code is based on the 0.1.130 version and incorporated into the part of the patch, by gradually backing out the patch, it is determined to "virtiolib: Rewrite VirtIORing.c, 9b89619f7787088dd026da01e3fb566ccb9b9397" commit.
we are throwing the specific code location of the problem.

There was also a problem with the community's last master code, and the Fedora 0.1.141 did not include the commit, so it did not find this problems.

Please check this issue.
tks all.

some information:
Hostos: 3.10.0-327.61.59.57_15.x86_64 is based on CentOS 7.2
qemu: 2.6
qemu exit log:
![image](https://user-images.githubusercontent.com/5163063/32988302-2dd02d3c-cd3c-11e7-8a57-2a8b16bb9c7d.png)

zyb521:
qemu cmd:
[...]

PR #203, opened 2017-11-28
vioserial: Send VIRTIO_CONSOLE_PORT_OPEN (0) on port D0 exit

ladipro:
Virtio serial ports receive the D0 exit callback before the parent driver.
This is expected in this driver architecture of a parent bus driver and
ports modeled as children. It however causes issues because ports own their
respective queues and try to clean them up while VirtIO as a whole, to be
shut down by the master D0 exit callback, is still running.

As a result, D0 exit (caused by resource rebalance on CPU hot-plug for
example), may crash QEMU with errors similar to:

  Looped descriptor

or

  Guest moved used index from 253 to 176

because port D0 exits race with QEMU still pushing data to the queues.

This patch fixes it by sending the VIRTIO_CONSOLE_PORT_OPEN control message
with open=0, i.e. signaling to the host that the port is closed and should
not be receiving more traffic.

It would be nicer to address this on the VirtIO level but VirtIO 1.0
explicitly forbids the driver from shutting down individual queues:

  4.1.4.3.2 Driver Requirements: Common configuration structure layout
  [...]
  The driver MUST NOT write a 0 to queue_enable.

Alternatively, the driver might be refactored so that all queues are owned
by the parent bus driver which would then be in charge of cleaning them up
following the correct protocol, i.e. only after the VirtIO device has been
reset. This would be a bigger and riskier change.

Link: https://github.com/virtio-win/kvm-guest-drivers-windows/issues/195

### DIFF
--- a/vioserial/sys/Port.c
+++ b/vioserial/sys/Port.c
@@ -1450,6 +1450,12 @@ VIOSerialPortEvtDeviceD0Exit(
 
     VIOSerialDisableInterruptQueue(GetInQueue(Port));
 
+    if (Port->GuestConnected)
+    {
+        VIOSerialSendCtrlMsg(Port->BusDevice,
+            Port->PortId, VIRTIO_CONSOLE_PORT_OPEN, 0);
+    }
+
     WdfSpinLockAcquire(Port->InBufLock);
     VIOSerialDiscardPortDataLocked(Port);
     Port->InBuf = NULL;


ladipro:
No, I want the port to be reopened in VIOSerialPortEvtDeviceD0Entry so it needs to remember that it was open before the D0 exit.

If an actual file close comes in between, it will not even check the flag because Port->Removed will be true. So there should be no double-reporting of VIRTIO_CONSOLE_PORT_OPEN (0) either.

PR #204, opened 2017-11-28
build: Debug config support for PackOne

irudakov77:
Now PackOne takes files from objchk_* and copy them to Install_Debug when project configuration name ends with "Debug". It's required for NDIS5 driver build.
Also, PackOne source and dest can be overrided in the project configuration.

### DIFF
--- a/Tools/Driver.PackOne.targets
+++ b/Tools/Driver.PackOne.targets
@@ -143,9 +143,14 @@ Related features:
   </PropertyGroup>
 
   <!-- $(PackOne_SourcePath) and $(PackOne_DestinationPath) -->
-  <PropertyGroup>
-    <PackOne_SourcePath>objfre_$(TargetOS)_$(TargetArch)\$(PackOne_SourcePathArch)</PackOne_SourcePath>
-    <PackOne_DestinationPath>$(PackOne_DestinationPrefix)Install\$(TargetOS)\$(TargetArch)</PackOne_DestinationPath>
+  <PropertyGroup Condition="$(Configuration.EndsWith('Debug'))"> 
+    <PackOne_SourcePath Condition="$(PackOne_SourcePath)==''">objchk_$(TargetOS)_$(TargetArch)\$(PackOne_SourcePathArch)</PackOne_SourcePath>
+    <PackOne_DestinationPath Condition="$(PackOne_DestinationPath)==''">$(PackOne_DestinationPrefix)Install_Debug\$(TargetOS)\$(TargetArch)</PackOne_DestinationPath>
+  </PropertyGroup>
+  
+  <PropertyGroup Condition="!$(Configuration.EndsWith('Debug'))"> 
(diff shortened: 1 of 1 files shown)


ladipro:
virtio-win/master was in a bad state for a few minutes and it made the pull request look ugly.

Can you please kick it, so GitHub updates the PR? `git commit --amend` followed by `git push -f` should do it. Thanks!

irudakov77:
Done.

YanVugenfirer:
It looks like as a side effect empty Install directory with OS\platform subdirectories created as well.

Tools/Driver.PackOne.targets
```diff
@@ -143,9 +143,14 @@ Related features:
   </PropertyGroup>
 
   <!-- $(PackOne_SourcePath) and $(PackOne_DestinationPath) -->
-  <PropertyGroup>
-    <PackOne_SourcePath>objfre_$(TargetOS)_$(TargetArch)\$(PackOne_SourcePathArch)</PackOne_SourcePath>
-    <PackOne_DestinationPath>$(PackOne_DestinationPrefix)Install\$(TargetOS)\$(TargetArch)</PackOne_DestinationPath>
+  <PropertyGroup Condition="$(Configuration.EndsWith('Debug'))"> 
+    <PackOne_SourcePath Condition="$(PackOne_SourcePath)==''">objchk_$(TargetOS)_$(TargetArch)\$(PackOne_SourcePathArch)</PackOne_SourcePath>
```
ladipro: Is this missing single quotes around the variable? I don't think it's optional, both side are expected to expand to two single quotes if the variable is empty.

`  Condition="'$(PackOne_SourcePath)'==''"`

irudakov77: Yes it's missing and is really optional, how mentioned in https://msdn.microsoft.com/en-us/library/7szfhaft.aspx: "Single quotes are not required for simple alphanumeric strings or boolean values. However, single quotes are required for empty values." And it works for variables too, but actually is not matched to code style, thus I'll add quotes.

irudakov77:
> It looks like as a side effect empty Install directory with OS\platform subdirectories created as well.

Actually, it's not a PackOne side effect or any other degradation, because you get the same result without my commits to NetKVM. I can fix it, but prefer to make a separate pull request.

Tools/Driver.PackOne.targets
ladipro: Got it, thanks!

YanVugenfirer:
Ilya, thanks for the pull request - approving. Wating for CI to finish the build to merge.

ladipro:
> It looks like Install is created by the in NetKVM-VS2015.vcxproj.

by the `<CustomBuildStep>`, GitHub ate the markup :)

irudakov77:
Thank you. I plan to replace CustomBuildStep with PackOne if possible.

YanVugenfirer:
Thanks!

PR #205, opened 2017-11-28
virtiolib: fix compilation error

hammerg:
Search for the Linux imported header files in the local directory and not in the global Include files one.

Signed-off-by: Gal Hammer <ghammer@redhat.com>


YanVugenfirer:
Gal, is this some non-standard setup? NetKVM compiles without the issues for me.

PR #208, opened 2017-11-30
CParaNdisCX: SendControlMessage: Make sure the buffer is returned

sameehj:
The following assumption in CParaNdisCX::SendControlMessage:

"Control messages are processed synchronously in QEMU, so upon kick_buf
return, the response message has been already inserted in the queue"

Is not true, so we need to wait until the buffer is
returned by the virtqueue.

Prior to this patch the debug build of NetKVM would flood the logs
on startup of the device with the message:

"CParaNdisCX::SendControlMessage - ERROR: add_buf failed"

The driver would fill the virtqueue by not waiting for the buffers to
return due to the misleading assumption above which lead the queue to
become overflawed that caused failure in adding buffers to the queue.

This assumption seems to be true with vhost=on as this doesn't reproduce
with vhost=on.

### DIFF
--- a/NetKVM/Common/ParaNdis-CX.cpp
+++ b/NetKVM/Common/ParaNdis-CX.cpp
@@ -28,6 +28,8 @@ bool CParaNdisCX::Create(PPARANDIS_ADAPTER Context, UINT DeviceQueueIndex)
         return false;
     }
 
+    m_Context->m_CxStateMachine.Start();
+
     return m_VirtQueue.Create(DeviceQueueIndex,
         &m_Context->IODevice,
         m_Context->MiniportHandle,
@@ -89,6 +91,7 @@ BOOLEAN CParaNdisCX::SendControlMessage(
         if (0 <= m_VirtQueue.AddBuf(sg, nOut, 1, (PVOID)1, NULL, 0))
         {
             UINT len;
+            m_Context->m_CxStateMachine.RegisterOutstandingItem();
 
             m_VirtQueue.Kick();
             while (!m_VirtQueue.GetBuf(&len))
@@ -96,6 +99,8 @@ BOOLEAN CParaNdisCX::SendControlMessage(
                 UINT interval = 1;
                 NdisStallExecution(interval);
             }
+            m_Context->m_CxStateMachine.UnregisterOutstandingItem();
+
             if (len != sizeof(virtio_net_ctrl_ack))
             {
                 DPrintf(0, ("%s - ERROR: wrong len %d\n", __FUNCTION__, len));
--- a/NetKVM/Common/ParaNdis-CX.cpp
+++ b/NetKVM/Common/ParaNdis-CX.cpp
@@ -89,16 +89,14 @@ BOOLEAN CParaNdisCX::SendControlMessage(
         if (0 <= m_VirtQueue.AddBuf(sg, nOut, 1, (PVOID)1, NULL, 0))
         {
             UINT len;
-            void *p;
-            /* Control messages are processed synchronously in QEMU, so upon kick_buf return, the response message
-              has been already inserted in the queue */
+
             m_VirtQueue.Kick();
-            p = m_VirtQueue.GetBuf(&len);
-            if (!p)
+            while (!m_VirtQueue.GetBuf(&len))
             {
-                DPrintf(0, ("%s - ERROR: get_buf failed\n", __FUNCTION__));
+                UINT interval = 1;
+                NdisStallExecution(interval);
             }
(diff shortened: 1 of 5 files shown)


NetKVM/Common/ParaNdis-CX.cpp
```diff
@@ -89,16 +89,15 @@ BOOLEAN CParaNdisCX::SendControlMessage(
...
+
             m_VirtQueue.Kick();
-            p = m_VirtQueue.GetBuf(&len);
-            if (!p)
+            while (!m_VirtQueue.GetBuf(&len))
             {
-                DPrintf(0, ("%s - ERROR: get_buf failed\n", __FUNCTION__));
+                LARGE_INTEGER interval;
+                interval.QuadPart = -1;
+                KeDelayExecutionThread(KernelMode, FALSE, &interval);
```
ladipro: I'm afraid that we are at IRQL = DISPATCH_LEVEL at this point and KeDelayExecutionThread is illegal to call.

Firstly, we're holding a spinlock (`CLockedContext<CNdisSpinLock>` at the top of the function). And even if we found a way around this and were able to release it while waiting, we still might be running in the DPC.

NdisMSleep should work here. It's already used to implement the 'sleep' VirtioLib callback:
https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/NetKVM/Common/ParaNdis-VirtIO.cpp#L463

YanVugenfirer: First of all, NdisMSleep should be used on IRQL < DISPATCH_LEVEL (https://msdn.microsoft.com/en-us/library/windows/hardware/ff563677(v=vs.85).aspx).  NdisStallExecution call work here (https://msdn.microsoft.com/en-us/library/windows/hardware/ff564568(v=vs.85).aspx).

My second concern is that having a loop like this can hang the driver on pause or PM operation. Those can arrive in parallel and the driver might stop the queue. So I think we should sync with state machine as well.

ladipro: Oops, so the 'sleep' VirtioLib callback should be fixed too.

Interestingly, none of the callers of CParaNbuffersdisCX::SendControlMessage do anything with the return value other than debug printout (likely a bug), which makes me wonder if we can make this fully asynchronous, "fire and forget" style. Unless the driver really needs to wait for the host to process any these control messages. Does it?

sameehj: Apparently CX path has no integration with the state machine while the RX [1] and TX [2] both have. While taking into consideration that with vhost=on the loop won't be executed I do think that reading from the device may affect performance a bit while waiting for a reply as commands on the CX path get executed several times a second especially on cores where CX path share the core with TX and RX.

> Unless the driver really needs to wait for the host to process any these control messages. Does it?
From the quick look I made it seems that the return doesn't seem helpful at all as no one is taking it into account beside a debug print. So this might be an option too but it should be taken into consideration that this might change in the future and thus a proper handling mechanism should be introduced as well.

[1] https://github.com/virtio-win/kvm-guest-drivers-windows/blob/0b8bb1abae7132518b052707c03ab9ebbbba47c6/NetKVM/Common/ndis56common.h#L399

[2] https://github.com/virtio-win/kvm-guest-drivers-windows/blob/0b8bb1abae7132518b052707c03ab9ebbbba47c6/NetKVM/Common/ParaNdis-TX.h#L264

YanVugenfirer: Regarding async handling - need to take a close look at current configuration changes.

sameehj:
I have updated the PR as requested (now using NdisStallExecution instead) and added integration for CX path into the state machine.

NetKVM/Common/ParaNdis-SM.h
```diff
@@ -2,18 +2,21 @@
 
 #include "ParaNdis-Util.h"
 
-class CDataFlowStateMachine : public CPlacementAllocatable
+class CFlowStateMachine : public CPlacementAllocatable
 {
 public:
-    void Start()
+    virtual void Start()
```
ladipro: I usually don't argue against virtual methods if they make the code more elegant and readable. But in this case I don't follow why you made all of these virtual. They don't seem to be overriden anywhere. Is it in preparation for further changes? Thanks!

sameehj: Well no there is no upcoming change soon, but I though it would be more elegant this way as it emphasis that this methods can be overridden.

If you think this isn't good this can be reverted by a suitable patch =)

ladipro: I'm fine with it either way :) Thanks!

PR #209, opened 2017-12-01
Simplify event suppression

ladipro:
There's no point in telling VirtioLib about the event suppression setting using  virtio_set_queue_event_suppression() or the bEnableInterruptSuppression field. VirtioLib can detect the presence of the VIRTIO_RING_F_EVENT_IDX feature (as acked by the driver) internally.

This series removes the API and all its uses, which simplifies all drivers.

### DIFF
--- a/Balloon/sys/balloon.c
+++ b/Balloon/sys/balloon.c
@@ -54,15 +54,12 @@ BalloonInit(
     WdfObjectAcquireLock(WdfDevice);
 
     // inflate
-    params[0].bEnableInterruptSuppression = false;
     params[0].Interrupt = devCtx->WdfInterrupt;
 
     // deflate
-    params[1].bEnableInterruptSuppression = false;
     params[1].Interrupt = devCtx->WdfInterrupt;
 
     // stats
-    params[2].bEnableInterruptSuppression = false;
     params[2].Interrupt = devCtx->WdfInterrupt;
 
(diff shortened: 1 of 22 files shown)


ladipro:
The NDIS5 PublishIndices configuration is actually used though. Unlike NDIS6 where it's dead code.

I'll remove it if you confirm this is OK. Thanks!

YanVugenfirer:
Please remove the code and INF entry from NDIS5 as well.

Regarding the debug params that don't have entries in INF - at some point, it was decided that it is better not to expose them to the user but keep them for debugging purposes. And for each new virtio feature we used to have a flag to turn it off (several years ago it made sense). Because NDIS5 code was frozen - it still has the parameters in INF.

ladipro:
Added "netkvm/NDIS5: Remove the "PublishIndices" config entry" as requested.

ladipro:
The "// Will be removed" lines are actually removed in "virtiolib: Delete unused API" https://github.com/virtio-win/kvm-guest-drivers-windows/pull/209/commits/86123bb7374c320327375a8e2baa3ff434ef38bc

I organized the changes this way to keep the commits small / easy to review / easy to bisect.

sameehj:
> The "// Will be removed" lines are actually removed in "virtiolib: Delete unused API" 86123bb
I organized the changes this way to keep the commits small / easy to review / easy to bisect.

Oh, I missed that.

PR #222, opened 2017-12-13
balloon: Static analysis fixes and tweaks

ladipro:
A bunch of minor tweaks to make static analysis happy and a few actual bug fixes, mostly around error handling.

### DIFF
--- a/Balloon/app/main.cpp
+++ b/Balloon/app/main.cpp
@@ -29,7 +29,7 @@ ULONG
 _cdecl
 wmain(
     __in              ULONG argc,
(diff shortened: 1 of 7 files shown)


ladipro:
Oops, fixed by adding ifdefs for downlevel builds:

```
#if !defined(NTDDI_WIN8) || (NTDDI_VERSION < NTDDI_WIN8)
                       0,
#else
                      POOL_NX_ALLOCATION,
#endif

```
and

```
#ifndef _IRQL_requires_
#define _IRQL_requires_(level)
#endif

```
Thanks!

PR #223, opened 2017-12-14
vioserial: Static analysis fixes and tweaks

ladipro:
Minor tweaks to make static analysis happy.


sameehj:
Just wondering, did you ran SDV checks other than the ones that run with the "BuildAll" script?

ladipro:
> Just wondering, did you ran SDV checks other than the ones that run with the "BuildAll" script?

I used the "Run Code Analysis" command in VS. It's much faster than what BuildAll does for NetKVM, vioscsi, and viostor so it doesn't look like the same thing.

right click a project -> Project Only -> Run Code Analysis on Only <project name>

vs.

Driver -> Launch Static Driver Verifier...

PR #225, opened 2017-12-28
Cleanup of IVSHMEM and add ability to specify cache type

gnif:
This PR performs two things:

* moves ioctl code from the large switch statement into local static functions
* Adds `IVSHMEM_MMAP_CONFIG` to the `IVSHMEM_REQUEST_MMAP` IOCTL which allows specifying the cache type rather then always assuming write combining.


YanVugenfirer:
Can you please post the patch of the test application as well?

gnif:
Sure but this will need to wait as I am away on vacation and do not have access to my build environment.

YanVugenfirer:
Hi Geoffrey,

Are you back from vacation? Can you add the test app patch?

Thanks,
Yan.

PR #230, opened 2018-01-09
netkvm: If the driver doesn't support RSS, it should not ACK virtio multi-queue feature

jackli577:
…rt RSS

This bug caused by when frontend notify qemu support multi queue, but in face not support, just one hardqueue, qemu alloc max queue(large then one), but just enable one queue;
dpdk do hot upgrade, save all the queue value , then recover,  crash.
this change to when frontend not support multi queue, not to notify qemu support, avoid qemu to alloc more than one queue.

Signed-off-by: jackli <1744764178@qq.com>

### DIFF
--- a/NetKVM/Common/ParaNdis-Common.cpp
+++ b/NetKVM/Common/ParaNdis-Common.cpp
@@ -752,16 +752,20 @@ NDIS_STATUS ParaNdis_InitializeContext(
     if (pContext->ulPriorityVlanSetting)
         pContext->MaxPacketSize.nMaxFullSizeHwTx = pContext->MaxPacketSize.nMaxFullSizeOS + ETH_PRIORITY_HEADER_SIZE;
 
+#if PARANDIS_SUPPORT_RSS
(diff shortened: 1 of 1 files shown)


jackli577:
hi, sameehj
i did it.

jackli577:
@sameehj 
 Is this ok?

sameehj:
> Is this ok?

The * is still in the title of the commit and there are still tabs instead of spaces, actually you didn't change the commit itself , only the pull request title. You should apply these changes in the commit itself on your local branch and then force push the changes to your own repository. Once you do that, the pull request will be updated with the new commit.

jackli577:
@sameehj 
sorry, is ok?

sameehj:
Hi @jackli577 ,

I have rephrased the commit title/message a bit, please take a look and tell me if this is okay from your side.

https://github.com/sameehj/kvm-guest-drivers-windows/commit/a2f5549cf9378e6e047afb59f9a7deaee9b5f332

jackli577:
@sameehj  ofcourse，thanks,.

Issue #171, opened 2017-10-05
Disk conflicts


sundarchi1:
I have a standard Windows Server 2016 boot disk and also a block storage disk. Restarting the server, I get these events logged. seems like an issue with the drivers, and the info in that KB was not useful. Reformatting the disks and such aren't helping.

![diskerr1](https://user-images.githubusercontent.com/31782486/31241670-a38e19cc-a9d2-11e7-939a-5d19ac95bada.PNG)
![diskerr2](https://user-images.githubusercontent.com/31782486/31241671-a38e2ba6-a9d2-11e7-91a7-f1dcdced422c.PNG)

vrozenfe:
Hi sundarchi1,
Did you specify a unique serial name for each drive?
If you can, please post the qemu command line.

Thanks,
Vadim.

sundarchi1:
This is from a 100% clean install. Shouldn't all serials be generated and be unique the first time?

vrozenfe:
not sure about libvirt, but if you create qemu command line by yourself, the serial number needs to be explicitly specified like this "-drive file=.........,serial=XXXXXXXXXXXX,..."

vrozenfe:
Hi sundarchi1,

Any update on this case? 
Did "serial" help to solve your problem?
If it did, let's close the issue.

Thanks,
Vadim

sundarchi1:
I do not really know what qemu command line is. I just install my ISO through a custom VPS provider and they tell me to also add these drivers, and these errors show up.

vrozenfe:
can you please do the following things:
- start command prompt as administrator and run diskpart,
- type "list disk" to see all the disk attached (you should see at least two),
- then go to the first one "select disk 0" and ask for details "detail disk",
- repeat the above step for all other disks in your system,
- post the results to this thread.

Apart from that, you can try converting your data disk into GPT (if it is not GPT yet),
and see if it solved the problem. But don't forget to backup your data on that disk first.

sundarchi1:
Two disk results:

![0](https://user-images.githubusercontent.com/31782486/34910643-b2cd4bca-f887-11e7-9510-2453403a3e7f.PNG)
![1](https://user-images.githubusercontent.com/31782486/34910644-b2dad7a4-f887-11e7-9cb7-10bba6780e4f.PNG)

sundarchi1:
@vrozenfe ^^^

vrozenfe:
Hi sundarchi1.

[...]

PR #250, opened 2018-02-27
[viostor] Fix for bug 1549455 - Event ID: 158 is logged on Win10/WS2016 when …

vrozenfe:
Fix for https://github.com/virtio-win/kvm-guest-drivers-windows/issues/171
(Bug 1549455 - Event ID: 158 is logged on Win10/WS2016 whentwo or more virtio-blk disks attached to VM). 
Win10/WS2016 are logging Event Id 158 if  viostor driver fails to report unique identifier through
Device Identification VPD page (page number: 0x83)


YanVugenfirer:
Looks good.

One thought that I have - is there a chance that this patch might trigger re-activation of the already installed guest?

vrozenfe:
You mean due to the driver update? 
No, it shouldn't lead to Windows guest reactivation.

PR #255, opened 2018-03-14
Remove unused configuration

YanVugenfirer:
* Removing Debug configuration. Most of them are not compiled anyway, and they also not used

* For some of the solutions removed irrelevant OS configurations


YanVugenfirer:
Sending new PR

PR #270, opened 2018-04-11
Add crashdumping through VMCoreInfo

viktor-prutyanov:
These pathes add driver which exposes Windows crashdump header to QEMU at startup. Then QEMU can use the header to produce full crashdump even if guest OS can't do it.
At the moment, this driver can only be used with 64-bit guest, also memory hotplug isn't supported.


viktor-prutyanov:
Hello,

> 1. Please rename the driver to fwcfg or something similar

Should I replace dummy fwcfg driver with this driver or keep it separate and just rename?

Regards,
Viktor Prutyanov

YanVugenfirer:
Hello,

If you are going to add 32bit support soon, than replace now. But if it will take some time, keep it separate.

Thanks,
Yan.

viktor-prutyanov:
Hello,

All 6 issues have been fixed.
Driver has been renamed to fwcfg64.

Regards,
Viktor Prutyanov

PR #277, opened 2018-04-26
[vioscsi]. Fix for  Bug 1568739 

vrozenfe:
Win7/win2008 guest BSOD after the first reboot when installing os with virtio_scsi
https://bugzilla.redhat.com/show_bug.cgi?id=1568739

### DIFF
--- a/vioscsi/vioscsi.c
+++ b/vioscsi/vioscsi.c
@@ -706,17 +706,13 @@ ENTER_FN();
             return FALSE;
         }
 
-        for (index = 0; index < adaptExt->num_queues; ++index) {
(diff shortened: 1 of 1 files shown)


vrozenfe:
Not initialized spin lock cased the problem.
The bug description is not completely correct here, because BSOD happened on every reboot, which didn't allow Windows to complete installation. If you wonder why it it happens on reboot, than explanation is quite simple. Witout going to deep into storage stack internals, miniport driver needs to 
know how to operate in two different modes - normal and crash dump. During the first stage of boot time installation phase, storport miniport driver works in crash dump mode, then reboot, switching to normal mode. Win7 driver failed into BSOD when tried to operate in normal mode.

PR #278, opened 2018-04-26
WPP Tracing: Set wppmodulename as project name

sameehj:
The wppmodulename is the friendly name of the message GUID. The
friendly name of the message GUID appears, by default, in the trace
message prefix.

By default, the friendly name of the message GUID is the name of
the directory in which the trace provider was built, this causes
some drivers as "sys". This commit overrides the default value
by setting the wppmodulename to be project name.

### DIFF
--- a/Tools/Driver.Common.props
+++ b/Tools/Driver.Common.props
@@ -153,6 +153,13 @@ Common property definitions used by all drivers:
     </Link>
   </ItemDefinitionGroup>
 
+  <!-- WPP settings -->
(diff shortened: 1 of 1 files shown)


YanVugenfirer:
Can the change here https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/Tools/Driver.Common.props solve it for all the projects at once?

sameehj:
@YanVugenfirer 
I have updated the PR as requested, this is indeed a better way to do this.

PR #279, opened 2018-05-07
fwcfg64: add decoded KdDebuggerDataBlock exposure

viktor-prutyanov:
After this patch driver can expose crashdump header with pointer to
decoded KdDebuggerDataBlock. In addition to producing crashdump at BSOD time,
QEMU using this structure can create valid live system dump.

Signed-off-by: Viktor Prutyanov <viktor.prutyanov@virtuozzo.com>


YanVugenfirer:
Thanks!

PR #281, opened 2018-05-11
netkvm: Driver should serialize process control msi

jackli577:
The bug was caused by a bug in the backend notify control msi to frontend indicate link status.
the problem is ISR route and ISRDPC route to process one global flag.

eg:
   1: the first isr come, set m_interruptReported true, the dpc process and clear, set m_interruptReported to false.
   2: the second isr come soon before the first isr clear m_interruptReported flag.
   3: the second isr dpc will not be processed, drop.

Signed-off-by: jackli <1744764178@qq.com>

### DIFF
--- a/NetKVM/wlh/ParaNdis6-Impl.cpp
+++ b/NetKVM/wlh/ParaNdis6-Impl.cpp
@@ -248,7 +248,6 @@ static BOOLEAN MiniportInterrupt(
     if (pContext->bCXPathCreated)
     {
         pContext->CXPath.DisableInterrupts();
-        pContext->CXPath.ReportInterrupt();
     }
     
     *QueueDefaultInterruptDpc = TRUE;
@@ -309,9 +308,6 @@ static BOOLEAN MiniportMSIInterrupt(
     CParaNdisAbstractPath *path = GetPathByMessageId(pContext, MessageId);
 
     path->DisableInterrupts();
-    path->ReportInterrupt();
-
-
 #if NDIS_SUPPORT_NDIS620
     if (path->DPCAffinity.Mask)
     {
@@ -361,6 +357,10 @@ static VOID MiniportInterruptDPC(
     PARANDIS_ADAPTER *pContext = (PARANDIS_ADAPTER *)MiniportInterruptContext;
     bool requiresDPCRescheduling;
 
+	if (pContext->bCXPathCreated)
+	{
+		pContext->CXPath.ReportInterrupt();
+	}
 #if NDIS_SUPPORT_NDIS620
     PNDIS_RECEIVE_THROTTLE_PARAMETERS RxThrottleParameters = (PNDIS_RECEIVE_THROTTLE_PARAMETERS)ReceiveThrottleParameters;
     DEBUG_ENTRY(5);
@@ -410,6 +410,10 @@ static VOID MiniportMSIInterruptDpc(
     PARANDIS_ADAPTER *pContext = (PARANDIS_ADAPTER *)MiniportInterruptContext;
     bool requireDPCRescheduling;
 
+	if (pContext->CXPath.getMessageIndex() == MessageId)
+	{
+		pContext->CXPath.ReportInterrupt();
+	}
 #if NDIS_SUPPORT_NDIS620
     PNDIS_RECEIVE_THROTTLE_PARAMETERS RxThrottleParameters = (PNDIS_RECEIVE_THROTTLE_PARAMETERS)ReceiveThrottleParameters;
 


sameehj:
Hi @jackli577 ,

Thanks for the patch. If I understand correctly I don't think this actually solves the root cause of the problem but reduces the odds of it happening. What would actually solve it is a counter that is atomically incremented when the interrupt is reported and decremented when it should be cleared. Please tell me if I am missing something...

YanVugenfirer:
Hi @jackli577 ,

Can you please describe what was the scenario where you reproduce this issue?

I think better solution to have a loop here:
```
if (pContext->CXPath.WasInterruptReported())
        {
            ReadLinkState(pContext);
            if (pContext->bLinkDetectSupported)
            {
                ParaNdis_SynchronizeLinkState(pContext);
            }
            if (pContext->bGuestAnnounceSupported && pContext->bGuestAnnounced)
            {
                ParaNdis_SendGratuitousArpPacket(pContext);
                pContext->CXPath.SendControlMessage(VIRTIO_NET_CTRL_ANNOUNCE, VIRTIO_NET_CTRL_ANNOUNCE_ACK, NULL, 0, NULL, 0, 0);
                pContext->bGuestAnnounced = FALSE;
            }
            pContext->CXPath.ClearInterruptReport();
        }
```
while instead of if. And also use atomic operation for setting and clearing the report.

Setting the report from DPC, will not solve the problem - if we get an additional interrupt, the report will be missed anyway.

Thanks,
Yan.

jackli577:
@YanVugenfirer 
the scenario is qemu notify two control interrupt quick,:
1:  the first control interrupt come,  set "report" true, queue in dpc, dpc find report is true,
     indicate to ndis.
2:  NOW,  THE  KEY  when indicate to nids, the second control interrupt come,  high IRQL
    set "report" to true again. queue dpc.
3:  then the first control  interrupt dpc complete. call ClearInterruptReport() set "report " false.

4: the last the second control interrupt dpc run, find "report" false,  do nothing with this dpc.

the key is interrupt route and the dpc  all to access  the "report" .

if loop that, can add one time check, but alse have a risk. here just check when control interrupt,
set report true, let os dpc on the same cpu will sequentialization  to do this.

" Setting the report from DPC, will not solve the problem - if we get an additional interrupt, the report will be missed anyway"

can u please describe what was the scenario, let me try.

sameehj:
@jackli577 Hi,

After checking the PR again I think we should do the following:
* For the legacy interrupts we should use atomic counter.
* For the MSI interrupts we should separate the CX path handling from the main DPC function "ParaNdis_DPCWorkBody" to a different DPC function, this would make the ReportInterrupt logic superfluous  as currently we are only using it with CX path. I believe that this check is very redundant with MSI support.

jackli577:
@sameehj 
the problem is not make sure atomic.
for msi cx use a singe dpc is better way, as u said, can make less to check this msi whether cx (in tx/rx)。  but also need to keep atomic,  and u also need to judge msi is cx or tx/rx in somewhere. interrup route or dpc.

sameehj:
@jackli577 
We have to do the check "pContext->CXPath.getMessageIndex() == MessageId" either way but we con't have to update the atomic counters at all, only schedule the dpcs for the CX path if the condition evaluates to true and execute RX and TX otherwise. Maybe, we should separate them too, but that needs to be done carefully along with performance testing as this might impact performance negatively.

In regards to why we need atomics, consider this scenario:

* The first Control interrupt arrives and schedules DPC
* The DPC of the first interrupt starts execution
* An interrupt arrives and sets that the interrupt was reported and schedules DPC, since the ISR runs in a higher IRQL, it can interrupt the DPC during it's execution.
* The first DPC continues to execute and clears the report interrupt flag and finishes
* The second DPC starts execution and the report flag check fails and this interrupt is missed.

jackli577:
@sameehj 
yes, u are right,  the change will result to check cx msi too much,  might impact performance.

this is also i  through,  let us to think more. because i through that, can not block msi， so do this change.  or  like @YanVugenfirer  said,  when process cx msi, just loop one more, like queue dpc again,

just like, plaster..

20lives:

jackli577:
@20lives  i done， it can fix that problem。

sameehj:
Hi @jackli577 ,

I have posted a pull request which should resolve all the issues that you're seeing neatly. Can you please test it and share your results?

https://github.com/virtio-win/kvm-guest-drivers-windows/pull/324

Thanks!

jackli577:
@sameehj  i am sorry to see this comment now,  
i can not understand why miss Control interrupts?

separates the cx and rx DPC, the logic see good,.

but i guess use a dpc to do this may may for rss situation, rss will queue dpc to the process same as
cx, so can use one dpc to process rx and cx.  just guess.

sameehj:
> @sameehj i am sorry to see this comment now,
> i can not understand why miss Control interrupts?
What do you mean?
> 
> separates the cx and rx DPC, the logic see good,.
> 
> but i guess use a dpc to do this may may for rss situation, rss will queue dpc to the process same as
> cx, so can use one dpc to process rx and cx. just guess.
We can, but should we? We are already doing so today. Separating the CX and TX/RX path to different DPCs makes a lot of sense since it does the following:

* Resolves the issue that was originally reported by you, that we miss control interrupts.
* We don't do unnecessary check of CX path when we know for sure that CX path has no work to do, this is an optimization.
* We don't do RX/TX work when we only need to handle CX interrupts.

jackli577:
Resolves the issue that was originally reported by you, that we miss control interrupts.
  ok. i misunderstand u said my pull request. 

We don't do unnecessary check of CX path when we know for sure that CX path has no work to do, this is an optimization.
We don't do RX/TX work when we only need to handle CX interrupts.

good.

YanVugenfirer:
Fixed by PR #324

PR #282, opened 2018-05-15
fwcfg64: Using PackOne instead of PostBuildEvent

irudakov77:
I think, PackOne is a better way to place driver files into Install dir.


YanVugenfirer:
Thanks!

PR #324, opened 2018-08-13
Fix possible race conditions with CX path.

sameehj:
This patch series address mainly the issues with CX interrupts that was originally reported in the following pull request:

https://github.com/virtio-win/kvm-guest-drivers-windows/pull/281


YanVugenfirer:
Closed according to sameehj commented on Aug 13

PR #329, opened 2018-08-31
netkvm and viostor: Proper do with enable msi.

jackli577:
1: netkvm:  when qemu enable 8 queue (msi count is 17), nPathBundles is 2, qemu send msi large than 6, will result os crash.
2: viostor: same to netkvm,  vcpu 4, qemu hard queue 2, viostor  num_queues will be 1,  send msi 3 will result adpater->dpc null, os crash

### DIFF
--- a/NetKVM/wlh/ParaNdis6-Impl.cpp
+++ b/NetKVM/wlh/ParaNdis6-Impl.cpp
@@ -262,7 +262,11 @@ static CParaNdisAbstractPath *GetPathByMessageId(PARANDIS_ADAPTER *pContext, ULO
     CParaNdisAbstractPath *path = NULL;
 
     UINT bundleId = MessageId / 2;
-    if (pContext->CXPath.getMessageIndex() == MessageId)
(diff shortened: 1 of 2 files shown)


sameehj:
The commit looks good to me, if I'm not mistaken this is a known issue in Qemu/KVM where less vectors get assigned than requested. @YanVugenfirer should be able to tell.

I am not familiar with the viostorage code, vadim can you confirm that this commit resolves the issue? @vrozenfe

One comment though, I think it is better too separate the commits and have one to each driver.

Thanks!

vrozenfe:
Hi  jackli577,
Thanks a lot for your patch

Is it reproducible on upstream qemu or can you share the relevant crashdump?
Can you pass "Verify driver support for Hot-Add CPU" test if the above patch applied?

Thanks,
Vadim.

jackli577:
@sameehj
if I'm not mistaken this is a known issue in Qemu/KVM where less vectors get assigned than requested

yes, because netkvm will ues hardqueue, vcpu and vector to get npathbundle.

if need i will to separate the commits and have one to each driver.

PR #332, opened 2018-09-17
netkvm: fix compilation problem without WPP

ybendito:
Use FIELD_OFFSET instead of offsetof


sameehj:
Just for the sake of writing history, the compiler gave the following error when compiling without WPP enable:

Severity	Code	Description	Project	File	Line	Suppression State
Error C3861 'offsetof': identifier not found	NetKVM-VS2015	E:\temp\sameeh\kvm-guest-drivers-windows\NetKVM\Common\ParaNdis-Common.cpp	573	

So the offset macro wasn't defined when WPP is disabled.

PR #338, opened 2018-10-12
netkvm: Calculate number of coalesced segments correctly

ybendito:
When the driver receives GRO coalesced packet from the device
it provides number of coalesced segments to the OS.
The estimation was not correct:
There was an assumption that segments always have MTU size,
  when this happens only by default. Certification tests and
  networking tools can control that (example is -M option of iperf)
Even in this case the calculation was wrong, it did not take
  in account that default MSS is smaller than default MTU.
Current commit corrects the calculation.

### DIFF
--- a/NetKVM/wlh/ParaNdis6-Impl.cpp
+++ b/NetKVM/wlh/ParaNdis6-Impl.cpp
@@ -846,11 +846,29 @@ VOID NBLSet8021QInfo(PPARANDIS_ADAPTER pContext, PNET_BUFFER_LIST pNBL, PNET_PAC
 
 #if PARANDIS_SUPPORT_RSC
 static __inline
-UINT PktGetTCPCoalescedSegmentsCount(PNET_PACKET_INFO PacketInfo, UINT nMaxTCPPayloadSize)
+UINT PktGetTCPCoalescedSegmentsCount(PPARANDIS_ADAPTER pContext,
+                                     PNET_PACKET_INFO PacketInfo,
+                                     UINT mss)
 {
-    // We have no corresponding data, following is a simulation
-    return PacketInfo->L2PayloadLen / nMaxTCPPayloadSize +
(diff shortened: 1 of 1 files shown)


NetKVM/wlh/ParaNdis6-Impl.cpp
```diff
@@ -923,9 +941,10 @@ tPacketIndicationType ParaNdis_PrepareReceivedPacket(
 #if PARANDIS_SUPPORT_RSC
             if (!(pContext->RSC.bIPv4SupportedQEMU || pContext->RSC.bIPv6SupportedQEMU) && (pHeader->hdr.gso_type != VIRTIO_NET_HDR_GSO_NONE))
             {
-                *pnCoalescedSegmentsCount = PktGetTCPCoalescedSegmentsCount(pPacketInfo, pContext->MaxPacketSize.nMaxDataSize);
+                *pnCoalescedSegmentsCount = PktGetTCPCoalescedSegmentsCount(pContext, pPacketInfo, pHeader->hdr.gso_size);
                 NBLSetRSCInfo(pContext, pNBL, pPacketInfo, *pnCoalescedSegmentsCount, 0);
-                DPrintf(1, "RSC host packet, datalen %d, GSO type %d\n", pPacketInfo->dataLength, pHeader->hdr.gso_type);
+                DPrintf(1, "RSC host packet, datalen %d, GSO type %d, mss %d, %d segments\n",
```
YanVugenfirer: Maybe it is a good idea to have an additional patch that raises the level for this print. After it will be printed for each packet.

ybendito: Currently the concept of logging is that in default logging level we print mainly logging of initialization and shutdown, during the connection there are no printouts. If we raise this print, we'll have a lot of such printouts on each file download. This is what I did during debugging and  this created a log of 500 MB in a minute.

PR #343, opened 2018-10-28
Unplug

ybendito:
This series of commits solves https://bugzilla.redhat.com/show_bug.cgi?id=1517107
In case of device hot unplug some packets may be in TX queue and must be returned to NDIS, otherwise device removal will not be completed.


ybendito:
The comment is correct, such a way it will be one patch less. More inputs?

YanVugenfirer:
Please proceed with Sameeh comment.

ybendito:
The PR updated

YanVugenfirer:
Thanks!

Waiting for CI to Run.

PR #351, opened 2018-11-30
NetKVM: Prevent hanging in SendControlMessage

npes87184:
Windows with NetKVM has a little chance hang during shutdown.

If the NetKVM has not initialized yet (or reset), QEMU will ignore
the control message. It will cause driver hanging in infinite while loop
for waiting response (which leads to windows hang in shutdown). In our
test, there is a little chance to receive an OID for setting multicast list
after device reset. To make symptom more clear, I have added some log.
Below is the log and call stack when this situation happens:

Logs:

[ParaNdis_OnShutdown]=>
Common [ParaNdis_ResetVirtIONetDevice] Done
[ParaNdis_OnOidSetMulticastList] Go
[ParaNdis_UpdateDeviceFilters] Do
[CParaNdisCX::SendControlMessage] enter: 4277 - cls: 0, cmd: 0
CParaNdisCX::SendControlMessage - ERROR: in while
Assertion: *** DPC watchdog timeout
    This is NOT a break in update time
    This is most likely a BUG in an ISR
    Perform a stack trace to find the culprit
    The period will be doubled on continuation
    Use gh to continue!!

Call stack:

kd> k
ChildEBP RetAddr
8b582530 83ab50be nt!KeAccumulateTicks+0x3c5
8b582570 83ab4f6b nt!KeUpdateRunTime+0x145
8b5825c8 83ab9c17 nt!KeUpdateSystemTime+0x613
8b5825c8 83a10122 nt!KeUpdateSystemTimeAssist+0x13
8b582680 8e16db70 hal!HalpTscStallExecutionProcessor+0xc0
WARNING: Stack unwind information not available. Following frames may be
wrong.
8b582700 8e16b216 netkvm+0x4b70
8b582730 8e16c418 netkvm+0x2216
8b582740 8e16f219 netkvm+0x3418
8b582750 8e17740f netkvm+0x6219
8b5827a8 880fc32b netkvm+0xe40f
8b5827f0 880a8ce3 ndis!ndisMDoOidRequest+0x3a1
8b582818 880a87ff ndis!ndisFQueueRequestOnNext+0x21e
8b582880 8df21086 ndis!NdisFOidRequest+0x181
8b5828a0 880fc961 pacer!PcFilterRequest+0x78
8b582900 880a8c5c ndis!ndisFDoOidRequest+0x1a2
8b582928 880a87ff ndis!ndisFQueueRequestOnNext+0x197
8b582990 8df15b22 ndis!NdisFOidRequest+0x181
8b5829b0 880fc961 wfplwf!FilterOidRequest+0x50
8b582a10 880a7799 ndis!ndisFDoOidRequest+0x1a2
8b582a3c 8810d602 ndis!ndisQueueRequestOnTop+0x224
8b582ab0 8810d686 ndis!ndisMOidRequest+0x1dd
8b582ac0 882a4646 ndis!NdisOidRequest+0xf
8b582b7c 88254d1d tcpip!FlpNdisRequestUnderReference+0x7d
8b582bb8 8825523b tcpip!FlpMulticastListSet+0xa1
8b582bc4 88251a2e tcpip!FlpSerializedNdisDeleteGroupWorker+0xd
8b582bec 83c4f460 tcpip!FlpSerializedNdisRequestWorkerRoutine+0x41
8b582c00 83ab7aab nt!IopProcessWorkItem+0x23
8b582c50 83c43f5e nt!ExpWorkerThread+0x10d
8b582c90 83aeb219 nt!PspSystemThreadStartup+0x9e
00000000 00000000 nt!KiThreadStartup+0x19

Signed-off-by: yuchenlin <yuchenlin@synology.com>

### DIFF
--- a/NetKVM/Common/ParaNdis-CX.cpp
+++ b/NetKVM/Common/ParaNdis-CX.cpp
@@ -102,17 +102,24 @@ BOOLEAN CParaNdisCX::SendControlMessage(
         if (0 <= m_VirtQueue.AddBuf(sg, nOut, 1, (PVOID)1, NULL, 0))
         {
             UINT len;
+            void *p;
             m_Context->m_CxStateMachine.RegisterOutstandingItem();
 
             m_VirtQueue.Kick();
-            while (!m_VirtQueue.GetBuf(&len))
+            p = m_VirtQueue.GetBuf(&len);
+            for (int i = 0; i < 1000 && !p; ++i)
             {
                 UINT interval = 1;
                 NdisStallExecution(interval);
+                p = m_VirtQueue.GetBuf(&len);
             }
             m_Context->m_CxStateMachine.UnregisterOutstandingItem();
 
-            if (len != sizeof(virtio_net_ctrl_ack))
+            if (!p)
+            {
+                DPrintf(0, "%s - ERROR: get_buf failed\n", __FUNCTION__);
+            }
+            else if (len != sizeof(virtio_net_ctrl_ack))
             {
                 DPrintf(0, "%s - ERROR: wrong len %d\n", __FUNCTION__, len);
             }


npes87184:
## How to reproduce:

Doing the following testing:
```
while (true):
    1. Poweron a VM (I used Win7SP1x32 here)
    2. Wait for guest agent up and we can get IP through agent
    3. Shutdown VM
```
In average, the problem will happen in about 100 rounds. Sometimes, it will occur after 700 rounds.

## Version of driver and QEMU will be affected:

The problem is due to commit [483778149](https://github.com/virtio-win/kvm-guest-drivers-windows/commit/4837781496314580b07824885d9dbeca6152a297), it will affect all driver version after this commit. I think the version is **0.1-146**.

In QEMU side, when the device reset, all the virtqueue's `vring.desc` will be set to zero. The behavior is shown in `virtio_reset` at QEMU `hw/virtio/virtio.c`. After this, if we receive the control message, QEMU will do nothing and the problem happens. The code in QEMU is:
```
void virtio_queue_notify(VirtIODevice *vdev, int n)
{
    VirtQueue *vq = &vdev->vq[n];

    // oops...
    if (unlikely(!vq->vring.desc || vdev->broken)) {
        return;
    }

    // handle the message
    trace_virtio_queue_notify(vdev, vq - vdev->vq, vq);
    if (vq->handle_aio_output) {
        event_notifier_set(&vq->host_notifier);
    } else if (vq->handle_output) {
        vq->handle_output(vdev, vq);
    }
}
```
The behavior of QEMU existed in a long time. At least in 2009-05-18 the code existed...

## Testing result:

After apply the patch:

* I am running above test on four Win7SP1x32 with vhost=on (`-netdev tap,ifname=tap,script=no,id=hostnet0,vhost=on,vhostfd=25 -device virtio-net-pci,netdev=hostnet0`), and all VMs have run 300 rounds now.
* I also tested it on Win7SP1x32 with vhost=off (`-netdev tap,ifname=tap,script=no,id=hostnet0 -device virtio-net-pci,netdev=hostnet0`) and it worked fine.

I don't know how many time of trying is best. What do you think?

sameehj:
I have did the following test:
E:\temp\sameeh\kvm-guest-drivers-windows\NetKVM>git diff
diff --git a/NetKVM/Common/ParaNdis-CX.cpp b/NetKVM/Common/ParaNdis-CX.cpp
index 69516db..e213366 100644
--- a/NetKVM/Common/ParaNdis-CX.cpp
+++ b/NetKVM/Common/ParaNdis-CX.cpp
@@ -105,11 +105,15 @@ BOOLEAN CParaNdisCX::SendControlMessage(
             m_Context->m_CxStateMachine.RegisterOutstandingItem();

             m_VirtQueue.Kick();
+            int count = 0;
             while (!m_VirtQueue.GetBuf(&len))
             {
                 UINT interval = 1;
                 NdisStallExecution(interval);
+                count++;
             }
+            DPrintf(0, "CParaNdisCX::SendControlMessage: The count is: %d", count);
+
             m_Context->m_CxStateMachine.UnregisterOutstandingItem();

             if (len != sizeof(virtio_net_ctrl_ack))

with vhost off:
The count was mostly 1 and a maximum of 4, the host is not heavy loaded so this number can be higher I believe.

with vhost on:
The count is always zero.

This behavior is expected since we know that with vhost on the queue replies synchronously and asynchronously with vhost off.

I think that number of tries around 100 - 1000 should keep the driver bulletproof. Can you try if the delay is tolerable with high number of tries such as 1000?

Thanks!

npes87184:
Hi, @sameehj 

I will start stress test tomorrow.

Thank you for your testing.
Yu-Chen Lin

npes87184:
Hi, @sameehj 

Start to test with the following setting:

### VMs:
- 8 Win7SP1x32

### Number of tries:
- 1000

### Target
- 1000 rounds

It may take 2-3 days to achieve, will update once it finishes or some bad things happen.

Thanks,
Yu-Chen Lin

npes87184:
Hi, @sameehj 

The 8 VMs run more than 1180 rounds, and everything is fine. I believe that setting number of tries to 1,000 is fine. I have updated the commit to retry 1,000 times. If you have other concern, please contact me.

Thanks,
Yu-Chen Lin

npes87184:
ping?

PR #352, opened 2018-12-05
This patch enables the CX path to send and receive more than one buffer in the CX path

sameehj:
The Control Queue can handle as much buffers as we want to, currently we
only support 2 buffers from the driver and both of them are from the
driver to the device.

This patch is not currently needed but should be required in the future for implemented RSS offload feature/ steering mode implementation.


YanVugenfirer:
We are going to pack everything that is needed for RSS\virtio into one message

PR #357, opened 2018-12-16
Test Certificate: Create a new test certificate

sameehj:
The old certificate has expired, this commit drops the old one and adds a
new one.

The new certificate automatically adds itself to the Trusted Root
Certificate Authorities store.

The certificate was created using the
following command line:

makecert -r -pe -ss root -n CN="Virto test certificate" VirtIOTestCert.cer

This commit also drops the pfx file which is not needed.


20lives:
**Tools/signAll.bat** script runs this command `signtool.exe sign /f %~dp0\VirtIOTestCert.pfx file` to sign all the cat and sys files with the signature.

PR #358, opened 2018-12-17
virtio-blk: set slot number to vpd

zhouwjfi:
Since virtio disks' vpds are same in pre-win10 os, different virtio disks have same
unique id. It affects some apps' behavior.

Signed-off-by: Zhou Wenjian <zhouwj.fi@gmail.com>


vrozenfe:
merged into master

Thank you.
Vadim.

PR #373, opened 2019-02-24
netkvm: allocate also inactive queues

ybendito:
https://bugzilla.redhat.com/show_bug.cgi?id=1608226
In case the driver uses less queues than the device supports
and unused queues are not initialized vhost start fails and
the device works without vhost. The root cause of the problem
is that the device touches also queues that the guest will
not use. Fixing in QEMU currently seems more complicated than
workaround in the driver (Linux driver also allocates all the
queues, including inactive ones). See also rejected QEMU fix:
http://lists.nongnu.org/archive/html/qemu-devel/2019-02/msg03374.html


YanVugenfirer:
This is a PR for a case we decide to use a solution in the driver. The best solution will be fixing QEMU.

YanVugenfirer:
This was experimental PR. Closing it for now.

PR #379, opened 2019-03-09
[viostor/vioscsi] use storport ddi function for sleeping (stall execution)

vrozenfe:


JonKohler:
Hey quick question - was looking through latest commits, this PR didn't have any description. What does this change actually do?

vrozenfe:
Nothing special. Just some trivial fix to replace OS routine with equal one, provided by port driver support library.

PR #405, opened 2019-09-23
[virtio-win] report the same TargetOsVersion for both Win8.1 and Win1…

vrozenfe:
Use the same TargetOsVersion (6.3) for both Win8.1 and Win10.

### DIFF
--- a/viosock/lib/precomp.h
+++ b/viosock/lib/precomp.h
@@ -36,6 +36,7 @@
 #include <strsafe.h>
 #include <ws2spi.h>
 #include <winternl.h>
+#include <crtdbg.h>
 
 #include "..\sys\public.h"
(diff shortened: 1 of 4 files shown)


YanVugenfirer:
Not merging for now.
We might need to build drivers explicitly for different Windows 10 versions.

vrozenfe:
Just for the record. 

A very similar patch has been merged already into our internal repository and used in internal build b173. This build was created to help Gal Zaidman checking his new virtio-win drivers installer which needs this change at the moment. I'm not going to revert this change in our internal repository unless Gal will ask to do it. 
That means that there  is a chance that some new builds published in fedoraproject will come with this or similar change inside.   

Cheers,
Vadim.

PR #410, opened 2019-10-31
viorng: fix vendor info substitution

irudakov77:

### DIFF
--- a/viorng/VirtRNG Package/VirtRNG Package.vcxproj
+++ b/viorng/VirtRNG Package/VirtRNG Package.vcxproj
@@ -85,8 +85,8 @@
     <TargetVersion>Windows7</TargetVersion>
     <UseDebugLibraries>false</UseDebugLibraries>
   </PropertyGroup>
-  <Import Project="$(MSBuildProjectDirectory)\..\..\Tools\Driver.Common.props" />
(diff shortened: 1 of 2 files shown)


viorng/VirtRNG Package/VirtRNG Package.vcxproj
```diff
@@ -104,6 +104,7 @@
     <OutDir>$(ProjectDir)..\Install\win8\x86\</OutDir>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Win10 Release|Win32'">
+    <DriverTargetPlatform>Desktop</DriverTargetPlatform>
```
YanVugenfirer: What will it take to move the platform to "Univeral". We would like to be able to build VirtRNG for the ARM64 platform eventually.

irudakov77: In short: we should remove custom coinstaller from inf file.

hammerg: No other way than removing the coinstaller? I tried using an extension inf file without success and still didn't find a way to register the driver as a RNG provider without an installer.

PR #419, opened 2019-12-09
Various fixes and improvements in netkvm

ybendito:


ybendito:
Will be updated later

PR #430, opened 2020-01-14
netkvm: avoid crash on second virtqueue shutdown

ybendito:
The BSOD was reported during engineering tests.
In hot unplug flow the Shutdown() was called when
the m_VirtQueue already zeroed. Avoid NULL pointer
dereferencing.


ybendito:
Changed the commit message, added BZ reference

PR #453, opened 2020-05-04
Work

vrozenfe:


vrozenfe:
bugfix for "BZ#1828658 - Windows guest BSOD when do Scan for hardware..." a bit more

PR #457, opened 2020-05-08
Sdv works

ybendito:
Add netkvm win8.1 configuration to SDV build.
Allow easy switch to modern SDV per-project.


ybendito:
Note that only netkvm project is **actually** affected and the changes related to netkvm are related only to the build (several files are just renamed). IMO, running HCK is not really needed.

PR #459, opened 2020-05-11
Sdv fix

ybendito:
Use modern SDV for Win 10 build of NetKvm
Do not build SDV for 8.1
Make SDV to do real checks on Win 10 (sacrifice crash callback in Win10 configuration)


ybendito:
Only build requires for all the drivers except NetKvm.
Only change in NetKvm is removal of crash callback

PR #463, opened 2020-05-21
WHQL fixes

hammerg:
Several fixes to issues that caused the PCI driver to fail in the WHQL tests.


YanVugenfirer:
There is some problem on CI side.

Basil is fixing it.

YanVugenfirer:
@hammerg There are still 3 failures.

PR #471, opened 2020-06-21
Various fixes to problems found during winfsp-tests execution.

hammerg:


viofs/svc/virtiofs.c
```diff
@@ -390,6 +399,8 @@ static NTSTATUS VirtFsCreateFile(VIRTFS *VirtFs,
         create_in.create.flags |= O_APPEND;
     }
 
+    create_in.create.flags |= O_EXCL;
```
YanVugenfirer: But what if Windows had CREATE_ALWAYS specified to CreateFile? 

According to https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea:
 Creates a new file, always. If the specified file exists and is writable, the function overwrites the file, the function succeeds, and last-error code is set to ERROR_ALREADY_EXISTS (183). If the specified file does not exist and is a valid path, a new file is created, the function succeeds, and the last-error code is set to zero

viofs/svc/virtiofs.c
```diff
@@ -321,7 +321,7 @@ static VOID SetFileInfo(struct fuse_attr *attr, FSP_FSCTL_FILE_INFO *FileInfo)
 {
     FileInfo->FileAttributes = PosixUnixModeToAttributes(attr->mode);
     FileInfo->ReparseTag = 0;
-    FileInfo->AllocationSize = attr->blocks * attr->blksize;
+    FileInfo->AllocationSize = attr->blocks * 512;
```
YanVugenfirer, line 328: Shouldn't we query volume for this info?

hammerg: According to the man page (https://linux.die.net/man/2/stat), no: number of 512B blocks allocated.

viofs/svc/virtiofs.c
hammerg: I'm not 100% sure but I think winfsp-tests cover this case and it is handled currently.

PR #472, opened 2020-06-24
SRIO-v Failover

ybendito:
NETKVM-VF teaming. Added VIOPROT protocol.
Note that the all the changes are in fact only for netkvm, no HCK of all the drivers required, only NETKVM


NetKVM/NetKVM-VS2015.sln
```diff
@@ -194,6 +199,36 @@ Global
 		{92E3B437-C258-47FB-8856-D3FEA56A3BCC}.Win8.1 Release|x64.Build.0 = Win8.1 Release|x64
 		{92E3B437-C258-47FB-8856-D3FEA56A3BCC}.Win8.1 Release|x86.ActiveCfg = Win8.1 Release|Win32
 		{92E3B437-C258-47FB-8856-D3FEA56A3BCC}.Win8.1 Release|x86.Build.0 = Win8.1 Release|Win32
+		{3149C26A-F437-42BD-A40D-D10A695F93EF}.Vista Release|ARM64.ActiveCfg = Release|ARM64
```
YanVugenfirer, line 202: Do we have ARM64 build for Vista? Isn't Vista built with old WDK?

ybendito: The Studio creates such configurations on each edit. So it exists, but not built.

PR #477, opened 2020-07-07
netkvm: Arm Windows support

mediouni-m:
NdisMAllocateSharedMemory maps memory as non cacheable on Arm Windows,
Given that Windows on Arm does not support unaligned accesses on uncached memory, this behavior does result in crashes.

As such, use MmAllocateContiguousMemorySpecifyCache for the Arm64 architecture.
This choice is safe. It's also to be noted that NdisMAllocateSharedMemory ignores the isCached attribute anyway.

With this fix, networking w/ netkvm for Windows on Arm works fine.
(tested with QEMU emulator version 5.0.50 (v5.0.0-1065-g9f1f264edb))

### DIFF
--- a/NetKVM/Common/ParaNdis_Common.cpp
+++ b/NetKVM/Common/ParaNdis_Common.cpp
@@ -767,7 +767,8 @@ NDIS_STATUS ParaNdis_InitializeContext(
         DumpVirtIOFeatures(pContext);
 
         // Enable VIRTIO_F_IOMMU_PLATFORM feature on Windows 10 and Windows Server 2016
-#if (WINVER == 0x0A00)
+        // Do not enable it on arm64 yet.
+#if (WINVER == 0x0A00) && !defined(_ARM64_)
         AckFeature(pContext, VIRTIO_F_IOMMU_PLATFORM);
 #endif
         AckFeature(pContext, VIRTIO_NET_F_STANDBY);
@@ -785,11 +786,6 @@ NDIS_STATUS ParaNdis_InitializeContext(
 
         InitializeLinkPropertiesConfig(pContext);
 
-#if !defined(_ARM64_)
-        pContext->bControlQueueSupported = AckFeature(pContext, VIRTIO_NET_F_CTRL_VQ);
-#else
-        DPrintf(0, "[%s] Control queue disabled for ARM64\n", __FUNCTION__);
-#endif
         pContext->bGuestAnnounceSupported = pContext->bLinkDetectSupported && pContext->bControlQueueSupported && AckFeature(pContext, VIRTIO_NET_F_GUEST_ANNOUNCE);
         InitializeMAC(pContext, CurrentMAC);
         InitializeMaxMTUConfig(pContext);
--- a/NetKVM/Common/ParaNdis_Util.cpp
+++ b/NetKVM/Common/ParaNdis_Util.cpp
@@ -23,7 +23,19 @@ bool CNdisSharedMemory::Allocate(ULONG Size, bool IsCached)
 {
     m_Size = Size;
     m_IsCached = IsCached;
+#if defined(_ARM64_)
+    /*
+     * On Windows on Arm, do not use NdisMAllocateSharedMemory.
+     * TODO: Figure out a neater way to allocate memory in those cases.
+     */
+    LARGE_INTEGER bound1, bound2;
+    bound1.QuadPart = 0;
+    bound2.QuadPart = MAXUINT64;
+    m_VA = MmAllocateContiguousMemorySpecifyCache(m_Size, bound1, bound2, bound1, MmCached);
+    m_PA = MmGetPhysicalAddress(m_VA);
+#else
     NdisMAllocateSharedMemory(m_DrvHandle, Size, m_IsCached, &m_VA, &m_PA);
+#endif
     return m_VA != nullptr;
 }
 
@@ -31,7 +43,11 @@ CNdisSharedMemory::~CNdisSharedMemory()
 {
     if(m_VA != nullptr)
     {
+#if defined(_ARM64_)
+        MmFreeContiguousMemorySpecifyCache(m_VA, m_Size, MmCached);
+#else
         NdisMFreeSharedMemory(m_DrvHandle, m_Size, m_IsCached, m_VA, m_PA);
+#endif
         m_VA = nullptr;
     }
 }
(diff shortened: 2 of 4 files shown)


YanVugenfirer:
@mediouni-m Thanks a lot!

Can you please add "Signed-off-by:" line at the end of the commit message.

Thanks,
Yan.

mediouni-m:
@YanVugenfirer Done now.

Thank you,

ybendito:
@mediouni-m
All our previous ARM-related fixes aimed to solve all the real misalignments in the driver.

> (tested with QEMU emulator version 5.0.50)

Initial ARM64 version (November 2018) with a lot of misalignments worked without problems on QEMU ARM emulation.
Before we accept this commit (**as temporary workaround only**) we need to ensure if does work on real ARM64.

In general, the driver currently should not have misaligned accesses, except of one that **should** be fixed by this patch (still not applied, as we do not have any ARM system on the table) https://github.com/ybendito/kvm-guest-drivers-windows/commit/a5df856e0ace1b3470467f2e47b0af4fa8f3775b.
We can't control what happens in tcpip stack, so for ARM we copy the data of incoming packet and indicate up its cached copy.

Several points for the patch:
- let's use _ARM64_ as in other places in the driver
- if we use Mm API, we need to avoid ACK for IOMMU feature, it is incompatible with this API

mediouni-m:
@ybendito 
Hello,

This commit works on real Arm machines. (tested on Graviton 2 here).

> Initial ARM64 version (November 2018) with a lot of misalignments worked without problems on QEMU ARM emulation.

This means that Qemu didn't honor the Arm memory access alignement requirements on device memory types.

> We can't control what happens in tcpip stack, so for ARM we copy the data of incoming packet and indicate up its cached copy.

For a NIC driver that works at line speed, we just cannot allow this to happen for speed reasons.
The Windows TCP stack cannot cope with passing through packets in uncached buffers too, by design. 

> if we use Mm API, we need to avoid ACK for IOMMU feature, it is incompatible with this API
The device works fine with iommu=smmuv3 on in practice.

mediouni-m:
Note that in the refresh to change _M_ARM64 to _ARM64_ I will add my previously local change to disable the copy of incoming packets too, as it is not needed in that configuration at all.

ybendito:
Last point when the functionality is reduced for ARM
https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/NetKVM/Common/ParaNdis_Common.cpp#L788

YanVugenfirer:
> @ybendito
> Hello,
> 
> This commit works on real Arm machines. (tested on Graviton 2 here).
> 
> > Initial ARM64 version (November 2018) with a lot of misalignments worked without problems on QEMU ARM emulation.
> 
> This means that Qemu didn't honor the Arm memory access alignement requirements on device memory types.
> 
> > We can't control what happens in tcpip stack, so for ARM we copy the data of incoming packet and indicate up its cached copy.
> 
> For a NIC driver that works at line speed, we just cannot allow this to happen for speed reasons.
> The Windows network stack cannot cope with passing through packets in uncached buffers too, by design. In practice that means copying every incoming packet, which isn't a sustainable overhead.
> 
> > if we use Mm API, we need to avoid ACK for IOMMU feature, it is incompatible with this API
> 
> The device works fine with iommu=smmuv3 on in practice. And those limitations is why a TODO comment is present. :-)
> (note that we only test Qemu emulating virtio-net, if any of you uses a true PCIe card emulating the virtio-net adapter that concern is valid)
> 

We have a code that ACKs virtio feature called VIRTIO_F_IOMMU_PLATFORM (it was recently renamed to VIRTIO_F_ACCESS_PLATFORM). MM* functions are incompatible with IOMMU. 

So the below line should be also put under #ifdef (it should not be called on ARM64 platform if we use MM* functions):
AckFeature(pContext, VIRTIO_F_IOMMU_PLATFORM);

https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html - VIRTIO_F_ACCESS_PLATFORM (the new name for VIRTIO_F_ACCESS_PLATFORM).

Best regards,
Yan.

> Hope this helps,

mediouni-m:
@ybendito good news, tested without control queue disabled in the driver source code and things still worked fine :)

mediouni-m:
@YanVugenfirer Acknowledged, We are still evaluating other functions that should work too. :)

Thank you,

YanVugenfirer:
@mediouni-m Thanks a lot! I am merging the PR.

Comment regarding CI failure: I manually revied the results. They are not related to the changes in this PR. The failure is under investigation, could be caused by PR#472.

pmsjt:
While this will work, because the NetKVM device is really a synthetic device, the solution is fundamentally incorrectly. We can only hope no one will copy and paste this to a real device driver.

NdisMAllocateSharedMemory determines if the memory should be cached or uncached based on the PCI attributes of the device. If the device is declaring these buffers as device memory, then this should really be mapped as uncached/device-memory. Anything else would cause a real device to misbehave. That's what this patch is doing.

There are two possible good alternate fixes:
- QEMU declares (via PCI) these buffers as something else other than device memory;
- The driver allocates cached shadow buffers and memcpys data from the device memory to the cached buffer which then passes to NDIS/TCP. This is the fix I proposed here: https://github.com/pmsjt/kvm-guest-drivers-windows/commit/ad3fe37e6c7fea02c2886f63daac69ac7d353bae

mediouni-m:
@pmsjt 

Hello,

> NdisMAllocateSharedMemory determines if the memory should be cached or uncached based on the PCI attributes of the device. 
In practice to us it didn't respect _CCA. That was tested on our platforms, on bare metal too.

We in fact use this exact approach as a workaround currently in our PCIe networking drivers. (on our processor families, DMA is cache coherent).

I've been talking about this with some engineers at MS at a while, I'd recommend continuing it via mediou@amazon.com.

Thank you,

pmsjt:
Thanks for reporting that. On ARM64, indeed _CCA is ignored when NDIS Scatter-Gather DMA is being used against WDM/HAL v2 APIs.

It is very much recommended to move the code away from NDIS SG, into WDF or WDM DMA APIs. Not only it solves this problem, but it also will improve Tx performance significantly.

The good news is that there is a flag that can be provided to NDIS SG to make is use HAL DMA v3 instead. You won't get the Tx performance improvements but it does make NDIS honor _CCA.

This simplifies things quite a lot. https://github.com/virtio-win/kvm-guest-drivers-windows/pull/479 

Thanks again for reporting this. The first time I looked into this, this device still was not being marked with _CCA. This was fixed since then but I had never looked again.

PR #478, opened 2020-07-07
Minor SRIOV-NetKVM changes

ybendito:


YanVugenfirer:
Comment regarding CI failure: I manually revied the results. They are not related to the changes in this PR. The failure is under investigation, could be caused by PR#472.

PR #479, opened 2020-07-10
Instruct NDIS Scatter-Gather DMA stack to use HAL DMA v3 APIs when available (NDIS_SUPPORT_NDIS650)

pmsjt:
NDIS Scatter-Gather DMA APIs are showing signs of old age and are becoming inefficient, especially on SoC systems. This driver should ultimately migrate from NDIS Scatter-Gather DMA APIs and move to either the WDF or WDM (HAL DMA v3) APIs as instructed here.

https://docs.microsoft.com/en-us/windows-hardware/drivers/network/ndis-scatter-gather-dma

This will result in a significantly improved performance, especially on SoC systems.

The migration process can be found here, by Jeffrey_Tippet_[MSFT], maitanier of these APIs in HAL by Microsoft:
https://community.osr.com/discussion/291413/getscattergatherlistex-and-ownership-lifetime-semantics

Meanwhile, the NDIS_SG_DMA_V3_HAL_API flag, added to NDIS_SG_DMA_DESCRIPTION::Flags allows unlocking a small portion of these benefits. One of such benefits is to allow NDIS on ARM64 to honor the PCI cacheable attributes (_CCA) for DMA memory as expressed by the device (or its parent bridge).

This means we can have NDIS naturally allocate cacheable memory, without having to special case / override memory allocation paths for ARM64.

### DIFF
--- a/NetKVM/Common/ParaNdis_Common.cpp
+++ b/NetKVM/Common/ParaNdis_Common.cpp
@@ -767,8 +767,7 @@ NDIS_STATUS ParaNdis_InitializeContext(
         DumpVirtIOFeatures(pContext);
 
         // Enable VIRTIO_F_IOMMU_PLATFORM feature on Windows 10 and Windows Server 2016
-        // Do not enable it on arm64 yet.
-#if (WINVER == 0x0A00) && !defined(_ARM64_)
+#if (WINVER == 0x0A00)
         AckFeature(pContext, VIRTIO_F_IOMMU_PLATFORM);
 #endif
         AckFeature(pContext, VIRTIO_NET_F_STANDBY);
@@ -785,7 +784,7 @@ NDIS_STATUS ParaNdis_InitializeContext(
         }
 
         InitializeLinkPropertiesConfig(pContext);
-
+        pContext->bControlQueueSupported = AckFeature(pContext, VIRTIO_NET_F_CTRL_VQ);
         pContext->bGuestAnnounceSupported = pContext->bLinkDetectSupported && pContext->bControlQueueSupported && AckFeature(pContext, VIRTIO_NET_F_GUEST_ANNOUNCE);
         InitializeMAC(pContext, CurrentMAC);
         InitializeMaxMTUConfig(pContext);
--- a/NetKVM/Common/ParaNdis_Util.cpp
+++ b/NetKVM/Common/ParaNdis_Util.cpp
@@ -23,19 +23,7 @@ bool CNdisSharedMemory::Allocate(ULONG Size, bool IsCached)
 {
     m_Size = Size;
     m_IsCached = IsCached;
-#if defined(_ARM64_)
-    /*
-     * On Windows on Arm, do not use NdisMAllocateSharedMemory.
-     * TODO: Figure out a neater way to allocate memory in those cases.
-     */
-    LARGE_INTEGER bound1, bound2;
-    bound1.QuadPart = 0;
-    bound2.QuadPart = MAXUINT64;
-    m_VA = MmAllocateContiguousMemorySpecifyCache(m_Size, bound1, bound2, bound1, MmCached);
-    m_PA = MmGetPhysicalAddress(m_VA);
-#else
     NdisMAllocateSharedMemory(m_DrvHandle, Size, m_IsCached, &m_VA, &m_PA);
-#endif
     return m_VA != nullptr;
 }
 
@@ -43,11 +31,7 @@ CNdisSharedMemory::~CNdisSharedMemory()
 {
     if(m_VA != nullptr)
     {
-#if defined(_ARM64_)
-        MmFreeContiguousMemorySpecifyCache(m_VA, m_Size, MmCached);
-#else
         NdisMFreeSharedMemory(m_DrvHandle, m_Size, m_IsCached, m_VA, m_PA);
-#endif
         m_VA = nullptr;
     }
 }
--- a/NetKVM/wlh/ParaNdis6_Impl.cpp
+++ b/NetKVM/wlh/ParaNdis6_Impl.cpp
@@ -110,25 +110,12 @@ BOOLEAN ParaNdis_InitialAllocatePhysicalMemory(
     ULONG ulSize,
     tCompletePhysicalAddress *pAddresses)
 {
-#if defined(_ARM64_)
-    UNREFERENCED_PARAMETER(pContext);
-    /*
-     * On Windows on Arm, do not use NdisMAllocateSharedMemory.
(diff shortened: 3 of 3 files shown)


pmsjt:
Thanks to @mediouni-m for pointing me in the right direction. I had been under the assumption that _CCA was still missing from this PCI device tree in QEMU and didn't bother check beyond. This was true as some point but has been fixed since then. My mistake for not double-checking.

ybendito:
@pmsjt Can you please do 3 things in this PR?
1.  Squash commits 2 and 3 (otherwise after commit 2 the build fails)
2. Change the last commit. In the commit "netkvm: Arm Windows support" was erroneously removed the ack for ctrl_vq feature, when the intention was to enable it for all the builds. I.e. should be just "pContext->bControlQueueSupported = AckFeature(pContext, VIRTIO_NET_F_CTRL_VQ);" without any ifdefs around.
3. Add signoff (git commit -s) for every commit.
Thanks in advance

YanVugenfirer:
Thanks a lot for the patch.

1. Please add “Signed-off-by” line to commit message (it can be done automatically by using git commit -s).

2. Did you test it on actual HW?

Thanks,
Tan.

pmsjt:
> Hello,
> 
> This change removed acknowledgment of control queue feature completely.
> Only the #if !defined should be removed with the debug message.
> 
> Thanks.

@YanVugenfirer  and @ybendito , there is a 4th commit which restores the control line. Is the 4th commit not visible to you in the PR?

pmsjt:
> @pmsjt Can you please do 3 things in this PR?
> 
> 1. Squash commits 2 and 3 (otherwise after commit 2 the build fails)
> 2. Change the last commit. In the commit "netkvm: Arm Windows support" was erroneously removed the ack for ctrl_vq feature, when the intention was to enable it for all the builds. I.e. should be just "pContext->bControlQueueSupported = AckFeature(pContext, VIRTIO_NET_F_CTRL_VQ);" without any ifdefs around.
> 3. Add signoff (git commit -s) for every commit.
>    Thanks in advance

Sounds good. I'll do that.

Is it OK if I just squash all 4 commits? It seems as the 4th commit is somehow going unnoticed.

pmsjt:
> 2. Did you test it on actual HW?

I did test the path on:
- ARM64: AMD Seattle (Operon A series)
- x64: HP Z420

Both running current Ubuntu with stock QEMU.

pmsjt:
> > @pmsjt Can you please do 3 things in this PR?
> > 
> > 1. Squash commits 2 and 3 (otherwise after commit 2 the build fails)
> > 2. Change the last commit. In the commit "netkvm: Arm Windows support" was erroneously removed the ack for ctrl_vq feature, when the intention was to enable it for all the builds. I.e. should be just "pContext->bControlQueueSupported = AckFeature(pContext, VIRTIO_NET_F_CTRL_VQ);" without any ifdefs around.
> > 3. Add signoff (git commit -s) for every commit.
> >    Thanks in advance
> 
> Sounds good. I'll do that.
> 
> Is it OK if I just squash all 4 commits? It seems as the 4th commit is somehow going unnoticed.

Completed:
- I quashed commits 2-4 which addresses requests 1 and 2
- Added signoff message.

Thank you.

NetKVM/Common/ParaNdis_Common.cpp
```diff
@@ -785,11 +785,6 @@ NDIS_STATUS ParaNdis_InitializeContext(
 
         InitializeLinkPropertiesConfig(pContext);
 
-#if !defined(_ARM64_)
```
YanVugenfirer: Sorry, but the second commit still removes the needed line.

You should keep this:
```
pContext->bControlQueueSupported = AckFeature(pContext, VIRTIO_NET_F_CTRL_VQ);
```

You can checkout all PR and see also the final result: https://docs.github.com/en/enterprise/2.15/user/articles/checking-out-pull-requests-locally

In any case, line 787 after "InitializeLinkPropertiesConfig(pContext);" line should be:
pContext->bControlQueueSupported = AckFeature(pContext, VIRTIO_NET_F_CTRL_VQ);

pmsjt: This does not make sense to me. When I look at the combined changes in "files" I see no such code removal:

https://github.com/virtio-win/kvm-guest-drivers-windows/pull/479/files

image

pmsjt: @YanVugenfirer , I see what is going on. The line was also missing even before mediouni-m's commit.

pmsjt: Re-added.

pmsjt: I also squashed all the commits, so it is now pretty obvious that my change was not removing the line. It was just not adding it.

pmsjt: I am rebuilding the driver and testing it with control queues now.

pmsjt: Tested - new driver with line added works.

mediouni-m:
Can confirm that this driver works too. Currently testing the change on... other NICs. Thank you again for finally documenting this. :D

pmsjt:
> Currently testing the change on... other NICs. Thank you again for finally

Sure! And thanks again for sounding the alarm. When I first looked into it, QEMU still didn't have the _CCA node on the device or any of its parents, so I didn't look any deeper. I wasn't aware that _CCA had been added in the meantime. This is great!

I tested the ARM64 version on my AMD Seattle (Opteron A Series) with Ubuntu 20.04. The x86_64 version I tested on my HP z420 running the same Distro/version.

mediouni-m:
Tested here in AWS Graviton 2 (Arm Neoverse N1).

@pmsjt for Tx we already use:
`The ENA driver supports two Queue Operation modes for Tx SQs:
- Regular mode
  * In this mode the Tx SQs reside in the host's memory. The ENA
    device fetches the ENA Tx descriptors and packet data from host
    memory.
- Low Latency Queue (LLQ) mode or "push-mode".
  * In this mode the driver pushes the transmit descriptors and the
    first 128 bytes of the packet directly to the ENA device memory
    space. The rest of the packet payload is fetched by the
    device. For this operation mode, the driver uses a dedicated PCI
    device memory BAR, which is mapped with write-combine capability.`

We will still look at using WDM if it'll allow us to drop latency further than what we currently have with that.

pmsjt:
Hi @YanVugenfirer. HCK-CI/NetKVM-Win10_1809x64 test seems to be still marked as "failed", but the details/log folder seems to be empty. Do you happen to know what this is about?

Thanks.

PR #495, opened 2020-08-24
nekvm: Use standard *JumboPacket instead of private MTU

ybendito:
*JamboPacket provides the same control over MTU.
The difference is that *JumboPacket includes ethernet header
size while the previous MTU does not.


YanVugenfirer:
rerun tests

Issue #498, opened 2020-08-31
vioscsi (generic/block): does not work with xhci hosts that accept only read/write of 512 blocks


tomty89:
For example, with https://github.com/torvalds/linux/commit/ec37198acca7b4c17b96247697406e47aafe0605, uas drives connected an ASM1142 will have max_sectors being [512 blocks](https://github.com/torvalds/linux/blob/v5.8/drivers/scsi/scsi_lib.c#L1796) instead  [1024 blocks](https://github.com/torvalds/linux/blob/v5.8/drivers/scsi/hosts.c#L450).

However, the Windows guest will insist on sending read/write commands of larger payload (e.g. 520 blocks). Sense data of the failure is apparently ignored (or maybe it's malformed) and the commands are kept being retried. I tried to set max_sectors of virtio-scsi-pci to 512 or lower and it does not help.

Currently qemu can manage to overwrite max_xfer_len and opt_xfer_len in the block limit VPD with the max_sectors set in the linux host (the non-qemu-specific queue limit) if scsi-generic is used. With scsi-block (only) max_xfer_len is overwritten to MAX_INT (opt_xfer_len in the verbatim VPD is 65535). I don't know if the fields matter at all in this issue. Neither of them works anyway.

scsi-hd seems to work fine.

I think it is supposed to work as long as I match the max_sectors of virtio-scsi-pci to the queue limit of the drive. Does the vioscsi driver care about it (like the scsi driver in a linux guest) at all?

vrozenfe:
@tomty89 Thank you for reporting this issue.

When it comes to scsi related stuff, including VPD_BLOCK_LIMITS page , vioscsi driver fully relied on data provided by qemu. 
vioscsi driver reads max_sectors information from scsi_config structure, but doesn't use it for any kind of sanity checks or internal adjustments. 

Definitely, it wouldn't be a problem to intercept VPD_BLOCK_LIMITS call and adjust MaximumTransferLength, but currently I'm not sure if it has to be handled by a guest driver, rather than by qemu. 

In any case, please provide information about qemu/kernel/virtio drivers versions along with qemu command line to help our QE reproducing the problem.

Best,
Vadim.

tomty89:
> @tomty89 Thank you for reporting this issue.
> 
> When it comes to scsi related stuff, including VPD_BLOCK_LIMITS page , vioscsi driver fully relied on data provided by qemu.
[...]

PR #502, opened 2020-09-04
[vioscsi] clamp NumberOfPhysicalBreaks against max_sectors

tomty89:
Since qemu commit 1bf8a989a566b2ba41c197004ec2a02562a766a4, seg_max
can be adjusted via the parameter virtqueue_size. There's no reason
for us not to at least allow NumberOfPhysicalBreaks to be lowered.

Signed-off-by: Tom Yan <tom.ty89@gmail.com>

### DIFF
--- a/vioscsi/vioscsi.c
+++ b/vioscsi/vioscsi.c
@@ -330,7 +330,7 @@ BOOLEAN VioScsiReadRegistry(
            (PVOID)pBuf,
            sizeof(ULONG));
     adaptExt->max_physical_breaks = min(
-                                        max(SCSI_MINIMUM_PHYSICAL_BREAKS, adaptExt->max_physical_breaks),
+                                        max(SCSI_MINIMUM_PHYSICAL_BREAKS, adaptExt->max_physical_breaks + 1),
                                         SCSI_MAXIMUM_PHYSICAL_BREAKS);
 
     StorPortFreeRegistryBuffer(DeviceExtension, pBuf );
@@ -500,13 +500,20 @@ ENTER_FN();
     if(adaptExt->dump_mode) {
         ConfigInfo->NumberOfPhysicalBreaks  = SCSI_MINIMUM_PHYSICAL_BREAKS;
     } else {
-        adaptExt->max_physical_breaks = MAX_PHYS_SEGMENTS;
+        adaptExt->max_physical_breaks = min(
+                                            max(SCSI_MINIMUM_PHYSICAL_BREAKS,
+                                                /* Say NumberOfPhysicalBreaks of 64 is required, it won't work
+                                                   if we optimize the user-input requirement to 65. So just
+                                                   let the user choose whether to use 64 or 33 and save him
+                                                   or her from unnecessary calculation */
+                                                adaptExt->scsi_config.max_sectors << SECTOR_SHIFT >> PAGE_SHIFT),
+                                            MAX_PHYS_SEGMENTS + 1);
 #if (NTDDI_VERSION > NTDDI_WIN7)
         if (adaptExt->indirect) {
             VioScsiReadRegistry(DeviceExtension);
         }
 #endif
-        ConfigInfo->NumberOfPhysicalBreaks = adaptExt->max_physical_breaks + 1;
+        ConfigInfo->NumberOfPhysicalBreaks = adaptExt->max_physical_breaks;
     }
     RhelDbgPrint(TRACE_LEVEL_INFORMATION, " NumberOfPhysicalBreaks %d\n", ConfigInfo->NumberOfPhysicalBreaks);
     ConfigInfo->MaximumTransferLength = SP_UNINITIALIZED_VALUE;
--- a/vioscsi/vioscsi.h
+++ b/vioscsi/vioscsi.h
@@ -51,6 +51,7 @@ typedef struct VirtIOBufferDescriptor VIO_SG, *PVIO_SG;
 #define VIRTIO_MAX_SG            (3+MAX_PHYS_SEGMENTS)
 
 #define SECTOR_SIZE             512
+#define SECTOR_SHIFT            9
 #define IO_PORT_LENGTH          0x40
 #define MAX_CPU                 256
 


tomty89:
I wonder if we want to remedy the `-2` qemu did to `seg_max`...I'm still confused what's the point of `+1` (like, is it supposed to be some_power_of_2 + 1?)

vioscsi/vioscsi.c
```diff
@@ -485,7 +485,16 @@ ENTER_FN();
 
     /* Set num_queues and seg_max to some sane values, to keep "Static Driver Verification" happy */
     adaptExt->scsi_config.num_queues = 1;
-    adaptExt->scsi_config.seg_max = MAX_PHYS_SEGMENTS + 1;
+    /* Happy? (Begin) */
+    /* Getting virtqueue_size is prioritized over your happiness. I don't care. */
+    adaptExt->max_physical_breaks = min(
+                                        max(SCSI_MINIMUM_PHYSICAL_BREAKS, adaptExt->scsi_config.seg_max),
```
tomty89, line 491: `Should I do `SCSI_MINIMUM_PHYSICAL_BREAKS + 1` instead? (And also the clamp in for the registry)

tomty89: For the record, apparently the default value of `ConfigInfo->NumberOfPhysicalBreaks` is 0x11 (i.e. `17` / `SCSI_MINIMUM_PHYSICAL_BREAKS + 1`)...

vioscsi/vioscsi.c
```diff
@@ -500,13 +509,15 @@ ENTER_FN();
     if(adaptExt->dump_mode) {
         ConfigInfo->NumberOfPhysicalBreaks  = SCSI_MINIMUM_PHYSICAL_BREAKS;
```
tomty89, line 510: And this?

vioscsi/vioscsi.c
```diff
@@ -485,7 +485,16 @@ ENTER_FN();
 
     /* Set num_queues and seg_max to some sane values, to keep "Static Driver Verification" happy */
     adaptExt->scsi_config.num_queues = 1;
-    adaptExt->scsi_config.seg_max = MAX_PHYS_SEGMENTS + 1;
+    /* Happy? (Begin) */
+    /* Getting virtqueue_size is prioritized over your happiness. I don't care. */
+    adaptExt->max_physical_breaks = min(
+                                        max(SCSI_MINIMUM_PHYSICAL_BREAKS, adaptExt->scsi_config.seg_max),
+    /* It is unknown that whether clamping to MAX_PHYS_SEGMENTS (+1) is necessary. For now stick to the habit. */
+                                        // SCSI_MAXIMUM_PHYSICAL_BREAKS);
```
tomty89, line 493: We can consider `128` as well:
https://github.com/torvalds/linux/blob/v5.8/include/linux/scatterlist.h#L329
https://github.com/torvalds/linux/blob/v5.8/include/scsi/scsi_host.h#L383 (`128 * 4096 / 512 = 1024`)

YanVugenfirer:
@tomty89 when sending the final version of the PR, please add "Signed-off-by:" to the commit message.

Thanks.

tomty89:
@vrozenfe I think it's basically finished. Let me know if you have any thoughts. Thanks.

vrozenfe:
Sorry guys.
I cannot approve the above changes for the following reasons:

1. it changes the default values of MaximumPhysicalPages which means that
now in case of 256KB Read/Write IO (File Cache block size) five request will be 
initiated  instead of one.

Was:
C:\Tools>spti.exe d:
MaximumTransferLength: ffffffff (bytes)
MaximumPhysicalPages: 00000041
TrueMaximumTransfer: 00040000 (bytes)

Now:
MaximumTransferLength: ffffffff (bytes)
MaximumPhysicalPages: 00000010
TrueMaximumTransfer: 0000f000 (bytes)

2. The second reason is that MaximumPhysicalPages has to be increased by one.
I was trying to explain in the previous post, but probably was unclear in my explanations. 

Tom, can you trying out the following change on unmodified upstream code to see if it helps 
solving your problem?

index 93e2d2f4..41ffad7b 100755
--- a/vioscsi/vioscsi.c
+++ b/vioscsi/vioscsi.c
@@ -509,7 +509,7 @@ ENTER_FN();
         ConfigInfo->NumberOfPhysicalBreaks = adaptExt->max_physical_breaks + 1;
     }
     RhelDbgPrint(TRACE_LEVEL_INFORMATION, " NumberOfPhysicalBreaks %d\n", ConfigInfo->NumberOfPhysicalBreaks);
-    ConfigInfo->MaximumTransferLength = SP_UNINITIALIZED_VALUE;
+    ConfigInfo->MaximumTransferLength = (ConfigInfo->NumberOfPhysicalBreaks - 1) * PAGE_SIZE;//SP_UNINITIALIZED_VALUE;
 
 #if (NTDDI_VERSION >= NTDDI_WIN7)
     num_cpus = KeQueryActiveProcessorCountEx(ALL_PROCESSOR_GROUPS);

Thanks,
Vadim.

tomty89:
@vrozenfe Problem 1 should now be fixed. I forgot to check when/where the GetScsiConfig help was used.

As for problem 2, If you read the code carefully, you'll realize I didn't remove your +1. The +1 will only be "gone" if the user try to set/lower seg_max (the default will either be 126 or 254, both will be clamped down to 64+1). This PR is about making it possible for the user to adjust `NumberOfPhysicalBreaks` (via the `virtqueue_size` command line parameter, in addition to registry approach, which might be unavailable in some cases).

When they do so, they should guarantee the entered value is adjusted properly (+1 for registry, +3 for virtqueue_size), because there's no point in assuming whether they are already doing so or not. It's an *arbitrary* value.

Besides, you haven't even been clear about what the +1 is based on. (i.e. *what* + 1? some power of 2?) What does arbitrary + 1 even mean? If you can provide a base / a rule, perhaps we can do some rounding up / down (e.g. to some nearest power of 2). Only then will +1 makes sense.

Regardlessly, as I've mentioned above, `64+1` is being blindly kept in this PR.

tomty89:
Okay, after the last commit things become even clearer. I can summarize what this PR does now:
1. Move the +1 hack/optimization to the definition of MAX_PHYS_SEGMENTS; there are so many reasons to do so: to make the code less confusing and more consistent, to be ready for adjustable seg_max, to make it "clamp ceiling-ready".
2. make seg_max adjustable, while keeping the default and maximum as 64+1 (`MAX_PHYS_SEGEMTNS`)
3. fix the clamp ceiling in the registry reader

I can redo/resplit the commits into those "steps" if the PR is approved and if it is necessary.

If you can provide the base/rule for the +1, and would like to do some rounding to the user-input values (which may make +1 to them sensible), we can do it in such approach. But if you can't, this should be good enough. (I don't think the rounding approach is always a decent thing to do though, as it can still be an obstacle in some cases and/or causes confusions/surprises to users; but it is at least then do-able if you can tell: *what* + 1.)

tomty89:
Btw, I don't think we should make changes to MaximumTransferLength (again). This PR is not really/entirely about the problem I experienced anymore (as it was fixed on the kernel side), but about getting the code/driver here to make more sense / become more usable.

vrozenfe:
@tomty89 
Did you try walking through the Microsoft code I mentioned in https://github.com/virtio-win/kvm-guest-drivers-windows/issues/498#issuecomment-687009274
Can you describe  the problem that you are going to fix with your commits?

Thanks,
Vadim.

tomty89:
@vrozenfe I did, but it's not giving me any sense either. It only says it does -1, without giving a reason, and later says "because we did that so now we can add it back". So regardless of reason for the -1, it's doesn't seem like something *we* need to care about anyway.

And then the comment also says: "adapters that return MaximumPhysicalPages=0x10 depend on this to transfer aligned 64K requests in one piece", so it seems to me that what we want is the requests to be aligned to some power of 2 number of blocks / bytes (basically, making e.g. 0x10 - 1 then + 1 = 0x10), but with the +1 here (both in the original code or in my code), it will **prevent** that from happening (as I said, the trace results shows me the requests will be as big as 520 blocks / 260K, instead of  512 blocks / 256K; which means the +1 we are doing is wrong).

No offense, *can* you even tell its/your point in some "human language"? I mean, do *you* really know what the -1 / +1 is about? As I'm still not seeing that from anywhere. (It doesn't appear to be something we need to take care of anyway..)

I've listed the points of my current commits above. The main point is to make it possible for the users to lower the (maximum) number of pages when needed.

vrozenfe:
https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/ns-storport-_port_configuration_information
Maximum number of physical pages the storage adapter can manage in a single transfer (in other words, the extent of its scatter/gather support). By default, the value of this member is 0x11.  ....

Now think about that buffer can be non-page aligned. So if you allocate 4K non-page aligned buffer it can be two physical pages instead of one. For 256 KB it will be 256KB/4KB + 1. No one can grantee the first page in DMA request will be page-aligned. virtio
knows how to handle this situation, some old DMA chips required double-buffering to overcome such limitations.  

Now, speaking about, lowering the number on thy physical pages, can you please help me understand how the existing code prevents you from lowering this number down to SCSI_MINIMUM_PHYSICAL_BREAKS  ? What is technical reason for doing  such thing on a Virtual Machine ? What kind of storage work load scenario do you have in mind? 

Don't get me wrong, but if there is any technical issue with the current code, our QE needs to know how to reproduce the problem and confirm that your patch helps to fix it without any regressions.  If you don't mind please file a bug at RH bugzilla 
https://bugzilla.redhat.com/enter_bug.cgi?product=Virtualization%20Tools&component=virtio-win
to track this problem. 

Thanks,
Vadim.

tomty89:
> https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/ns-storport-_port_configuration_information
> Maximum number of physical pages the storage adapter can manage in a single transfer (in other words, the extent of its scatter/gather support). By default, the value of this member is 0x11. ....
> 
> Now think about that buffer can be non-page aligned. So if you allocate 4K non-page aligned buffer it can be two physical pages instead of one. For 256 KB it will be 256KB/4KB + 1. No one can grantee the first page in DMA request will be page-aligned. virtio
> knows how to handle this situation, some old DMA chips required double-buffering to overcome such limitations.

You are still not getting (to) the point.

I've been asking the real question *repeatedly*: *what* + 1. +1 does not have any meaning if the *what* is *arbitrary*, because arbitrary + 1 is still *arbitrary*.

I've given up on trying to figure out whether the +1 is necessary (as I stated *repeatedly*, the +1 has **NOT** been removed in my PR; please just see the code), because what matters is that, whether there's any base/rule for the `256KB` part of `256KB/4KB + 1` (i.e. it should not be arbitrary).

Say that part should always be a power of 2, what we really need to do is to round it to the nearest one. (`+1` or not is an entirely different subject, which I'm simply not going to touch; **but the rounding does make it possible/sensible for us to do +1 to user-input values** because then the "what" won't be arbitrary anymore.)

> 
> Now, speaking about, lowering the number on thy physical pages, can you please help me understand how the existing code prevents you from lowering this number down to SCSI_MINIMUM_PHYSICAL_BREAKS ? What is technical reason for doing such thing on a Virtual Machine ? What kind of storage work load scenario do you have in mind?

I'm not even sure why you asked me that. You only provided the method of registry for user to adjust the value, but registry isn't always available (e.g. Installation stage, or even Windows version, as checked by the code itself). What I'm doing is merely providing an extra method (command line parameter), which has been made possible in qemu itself (`virtqueue_size`).

Because of that (the method is just an extra, more accessible, one), I don't know what technical reason I need to provide. (It's just "possibility" / "flexibility" / "in case".) But as I mentioned in my original issue, apparently Windows ignores BL VPD in at least some case and therefore we need (or would want, at least) a (more accessible) method to clamp down the request size, when `hw_max_sectors` (times logical sector size) is smaller than what we allow here.

> 
> Don't get me wrong, but if there is any technical issue with the current code, our QE needs to know how to reproduce the problem and confirm that your patch helps to fix it without any regressions. If you don't mind please file a bug at RH bugzilla
> https://bugzilla.redhat.com/enter_bug.cgi?product=Virtualization%20Tools&component=virtio-win
> to track this problem.
> 
> Thanks,
> Vadim.

P.S. No offense but I really feel like I have been typing the same thing over and over again...have you even read?

tomty89:
I put it in the wrong way, about the rounding (in my last message, because I've been repeating things over and over again but seemingly ignored; and I end up forgetting).

What I really mean is: if we insist on doing +1 to user-input values (Again, I'm not talking about the clamp ceiling 64), it will only make sense if we *also* round/limit the base, so that the +1 will have *meaning*. If we do not want to make it inflexible like that, then the current PR is good to go: accept the user-input values as-in, i.e. let the users decide and do the job when they want *customize* the value).

(The reason we don't add back the 2 qemu deducted for us, which is unnecessary to our code, is that we don't know if one day it will stop doing it; so it's best to leave it to the users as well, add it if their qemu still does it. I should add that to the comment.)

tomty89:
@vrozenfe @YanVugenfirer Just redid the PR. Now it should be way less "offensive" and way more clearer. You can omit the fourth/last commit if necessary (e.g. breaks "static driver verification", whatever it is), as it doesn't matter to the users anyway. I hope you will actually read the commits and their messages before replying "I/We already said no".

vrozenfe:
@tomty89 
Thanks a lot. Just give me some time to go over your PR.
Best,
Vadim.

vrozenfe:
@tomty89 

There are some problems in your patch.

First of all, please let's keep all that "+1" stuff right were it is now. Belive me, anyone hwo knows how
scatter/gather bus master DMA works, togatehre with the fact that first physical  segment can be not 
4K aligned will be able to understand it.

Second, the logic behind of "/* We don't want to blow off VIO_SG and VRING_DESC_ALIAS */" is not completely 
correct, because you must be missing the following section https://github.com/tomty89/kvm-guest-drivers-windows/blob/master/vioscsi/vioscsi.c#L1362
Thired, using seg_max seems to be incorrect in this context. Please take a look at qemu code 
    virtio_stl_p(vdev, &scsiconf->seg_max,
                 s->conf.seg_max_adjust ? s->conf.virtqueue_size - 2 : 128 - 2);
which is not the maximum transfer size in case of  indirect descriptors. 
virtio spec  at https://docs.oasis-open.org/virtio/virtio/v1.1/virtio-v1.1.pdf
has a good explanation for that

seg_maxis the maximum number of segments that can be in a command. A bidirectional command caninclude
seg_max input segments and seg_max output segments.

max_sectors is a hint to the driver about the maximum transfer size to use

Please take a look at the following code (untested yet). Maybe it is enought to solve the problem. 

diff --git a/vioscsi/vioscsi.c b/vioscsi/vioscsi.c
index 93e2d2f4..c9a4ce27 100755
--- a/vioscsi/vioscsi.c
+++ b/vioscsi/vioscsi.c
@@ -498,7 +498,7 @@ ENTER_FN();
     ConfigInfo->MaximumNumberOfLogicalUnits = min((UCHAR)adaptExt->scsi_config.max_lun, SCSI_MAXIMUM_LUNS_PER_TARGET);
 
     if(adaptExt->dump_mode) {
-        ConfigInfo->NumberOfPhysicalBreaks  = SCSI_MINIMUM_PHYSICAL_BREAKS;
+        adaptExt->max_physical_breaks  = SCSI_MINIMUM_PHYSICAL_BREAKS;
     } else {
         adaptExt->max_physical_breaks = MAX_PHYS_SEGMENTS;
 #if (NTDDI_VERSION > NTDDI_WIN7)
@@ -506,10 +506,12 @@ ENTER_FN();
             VioScsiReadRegistry(DeviceExtension);
         }
 #endif
-        ConfigInfo->NumberOfPhysicalBreaks = adaptExt->max_physical_breaks + 1;
     }
-    RhelDbgPrint(TRACE_LEVEL_INFORMATION, " NumberOfPhysicalBreaks %d\n", ConfigInfo->NumberOfPhysicalBreaks);
-    ConfigInfo->MaximumTransferLength = SP_UNINITIALIZED_VALUE;
+    adaptExt->max_physical_breaks = min(adaptExt->max_physical_breaks, adaptExt->scsi_config.max_sectors);
+    ConfigInfo->NumberOfPhysicalBreaks = adaptExt->max_physical_breaks + 1;
+    ConfigInfo->MaximumTransferLength = adaptExt->max_physical_breaks * PAGE_SIZE;
+    RhelDbgPrint(TRACE_LEVEL_INFORMATION, " NumberOfPhysicalBreaks = %d, MaximumTransferLength = 0x%x\n",
+        ConfigInfo->NumberOfPhysicalBreaks, ConfigInfo->MaximumTransferLength);

Feel free to ping me ofline at vrozenfe_at_redhat_dot_com if you have any furthure questions.
Vadim.

tomty89:
@vrozenfe I couldn't read your patch properly because you didn't code-block it, but apparently the idea is to clamp max_physical_breaks against max_sectors.

I suddenly realize that back then I didn't know max_sectors is always in "linux sector" (i.e. 512 bytes) instead of logical block (of which the size is a variable). That's one of the reasons I went to seg_max, with the other being we are setting max_physical_size to a "MAX_PHYS_SEGMENT", which sound like the "same thing" as seg_max anyway.

So I'm okay with taking max_sectors as the clamp instead.

But your code isn't completely right. The clamp should be `max_sectors << 9 >> PAGE_SHIFT` (a.k.a. max_sectors * 512 / PAGE_SIZE) instead.

I'd suggest leaving `ConfigInfo->MaximumTransferLength` untouched, judging by the change history of it and the practical effect of it in the past experiments.

tomty89:
Just one more thing. Can you accept that we do the plus one *before* the clamping?

Here's the scenario/reason: when one needs the maximum physical breaks lowered (to fit host max_sectors of its drives or whatever limit), it doesn't matter how important this plus one is.

For example, if I need it to be 256 sectors / 32 page, it won't work if it's "optimized" to 33 page. So unless you sort of want to force me to use 16 (+1) page, doing it post-clamp adds nothing but extra calculation to me (max sector requirement - PAGE_SIZE / 512 instead of just max sector requirement). (It's the same reason you don't do `SCSI_MINIMUM_PHYSICAL_BREAKS + 1`  for dump mode, I think.)

That's why I still think we should do +1 in the definition of `MAX_PHYS_SEGMENTS` instead.

vioscsi/vioscsi.c
```diff
@@ -330,7 +330,7 @@ BOOLEAN VioScsiReadRegistry(
            (PVOID)pBuf,
            sizeof(ULONG));
     adaptExt->max_physical_breaks = min(
-                                        max(SCSI_MINIMUM_PHYSICAL_BREAKS, adaptExt->max_physical_breaks),
+                                        max(SCSI_MINIMUM_PHYSICAL_BREAKS, adaptExt->max_physical_breaks + 1),
```
tomty89, line 333: While I have this in the commit, there's no reason we should do so other than the fact that we set number of pages / physical breaks in registry (so it takes less work to calculate here than in the max_sectors case).

vioscsi/vioscsi.c
```diff
@@ -500,13 +500,20 @@ ENTER_FN();
...
     } else {
-        adaptExt->max_physical_breaks = MAX_PHYS_SEGMENTS;
+        adaptExt->max_physical_breaks = min(
+                                            max(SCSI_MINIMUM_PHYSICAL_BREAKS,
+                                                /* Say NumberOfPhysicalBreaks of 64 is required, it won't work
+                                                   if we optimize the user-input requirement to 65. So just
+                                                   let the user choose whether to use 64 or 33 and save him
+                                                   or her from unnecessary calculation */
+                                                adaptExt->scsi_config.max_sectors << SECTOR_SHIFT >> PAGE_SHIFT),
+                                            MAX_PHYS_SEGMENTS + 1);
```
tomty89, line 510: Obviously we can/should simply define `MAX_PHYS_SEGMENTS` to `64 + 1` instead.

vioscsi/vioscsi.c
```diff
@@ -500,13 +500,20 @@ ENTER_FN();
...
         ConfigInfo->NumberOfPhysicalBreaks  = SCSI_MINIMUM_PHYSICAL_BREAKS;
     } else {
-        adaptExt->max_physical_breaks = MAX_PHYS_SEGMENTS;
+        adaptExt->max_physical_breaks = min(
+                                            max(SCSI_MINIMUM_PHYSICAL_BREAKS,
+                                                /* Say NumberOfPhysicalBreaks of 64 is required, it won't work
+                                                   if we optimize the user-input requirement to 65. So just
+                                                   let the user choose whether to use 64 or 33 and save him
+                                                   or her from unnecessary calculation */
+                                                adaptExt->scsi_config.max_sectors << SECTOR_SHIFT >> PAGE_SHIFT),
```
tomty89, line 509: To further elaborate, when a user *need* the clamp, the limit would most likely be a max_sectors queue limit, so if we don't do anything stupid, he or she can use the limit directly as the clamp (so no calculation needed).

vioscsi/vioscsi.c
```diff
@@ -500,13 +500,20 @@ ENTER_FN();
...
+                                            MAX_PHYS_SEGMENTS + 1);
 #if (NTDDI_VERSION > NTDDI_WIN7)
         if (adaptExt->indirect) {
             VioScsiReadRegistry(DeviceExtension);
         }
 #endif
-        ConfigInfo->NumberOfPhysicalBreaks = adaptExt->max_physical_breaks + 1;
+        ConfigInfo->NumberOfPhysicalBreaks = adaptExt->max_physical_breaks;
```
tomty89, line 516: Read above before you say no.

JonKohler:
Can we close this PR now given there was a separate commit #685 ?

YanVugenfirer:
@vrozenfe Can we close this PR ?

tomty89:
While it's convenient, the meaning of max_sectors is twisted in #685, since it expect the value to be in PB instead of Linux/512b "sectors" (NOT logical block btw).

JonKohler:
> While it's convenient, the meaning of max_sectors is twisted in #685, since it expect the value to be in PB instead of Linux/512b "sector" (NOT logical block btw).

yea, I'm quite confused by it all. I'm going to file a separate bug, as we're not able to boot windows VMs on these latest drivers in our configuration, as we're using a vhost-scsi-user backend which clamps inflight IO to match up with seg_max, which I believe is the way that linux does it

vrozenfe:
I'm not  a Linux storage guy, but my understanding of the storage devices in Linux is:

- a segment is not  the same as sector or block size which is usually equal to 512b, butt the number of 
scatter-gather memory buffers the block device can translate. For example:
cat /sys/block/sdb/queue/max_segments
64

- if I check the max_segment_size on the same device , I get the following number
cat /sys/block/sdb/queue/max_segment_size 
4294967295
but max_segment_size  is just a maximum block size for one SG element. But the default size is equal to one page size (4K) on 
Intel CPUs (Windows recognizes the page size as 4K for all currently supported architectures).

- the max_sectors_kb parameter is just another proof of what I said above 
cat /sys/block/sdb/queue/max_sectors_kb 
256
This parameter determines the maximum buffer size that can be transferred down.
And max_sectors_kb/max_segments = 256KB/64 = 4KB

So, max_segments is the number of 4KB blocks that can be passed to the block device at once. 
And the vioscsi driver is completely following this logic. (Unless I am missing something, of cause).

All the best,
Vadim.

tomty89:
@vrozenfe what's the point are you trying to make? It doesn't matter what a segment is here.
The problem is that there's no qemu option that expect number of number PB/segment/page as its value except virtqueue_size.

What I was trying to say is, the option max_sectors expect number of 512b sectors as its value, as the name suggests, but while semantically speaking is wrong, practically speaking it might not cause any problem -- only that it makes no sense to users.

That's the reason why I had `<< SECTOR_SHIFT >> PAGE_SHIFT` in my patch. But of course it means that users will have to find out the number of sectors equivalent to the desired NOPB clamp, e.g. 520 sectors to get the clamp 65.

As for why with vioscsi (+ scsi_block) NOPB cannot exceed the device (hw_)max_sectors on the host, I have no idea, but it's a fact. I wish it's some sort of kernel bug that could be fixed so that there will be no more need or reason to continue this endless conversation / discussion.

tomty89:
For the record, max_segments is not directly related to (hw_)max_sectors:

```
$ cat /sys/block/sda/queue/max_{hw_,}sectors_kb
1024
1024
$ cat /sys/block/sda/queue/max_segments
2048
$ cat /sys/block/sdb/queue/max_{hw_,}sectors_kb
512
512
$ cat /sys/block/sdb/queue/max_segments
2048
```

Of course you can argue that the specific drivers are not setting max_segments to the actual limit, but max_segments is not used as a check or a clamp when (hw_)max_sectors is being determined / set either.

PR #503, opened 2020-09-12
Viosock: implement timeouts

irudakov77:


YanVugenfirer:
@irudakov77 There is a consistent failure in "DF - Reinstall with IO Before and After (Reliability)". Can you please take a look?

Thanks,
Yan.

Issue #513, opened 2020-10-22
viostor: 4Kn disk corruption after DISCARD/UNMAP


gusev-vitaliy:
4Kn virtio disks (virtio block device). After TRIM in Windows 10, disk becomes corrupted or Unavailable with message `E: is not accessible. Data error (cyclic redundancy check)`.

The reason of issue  - LBA sectors in guest virtio driver considered as 4K (in case of 4K emulated disk) and passed in discard operation w/o any translation, but VIRTIO b1.1 standard considers sectors as 512-byte units. That leads to discarding wrong data space in emulated disks.

vrozenfe:
Hi Vitaliy,

I already asked our QE to reproduce it 
( https://bugzilla.redhat.com/show_bug.cgi?id=1890810#c1 )
But it will be nice if you can confirm that the problem is reproducible on qemu and
provide us with qemu command line.

Thanks,
Vadim.

gusev-vitaliy:
Hi, Vadim. 

We saw it with bhyve hypervisor. I am not sure about qemu. Can qemu emulate 4Kn drives ?

----
Vitaliy Gusev

vrozenfe:
Sure. 
See https://bugzilla.redhat.com/show_bug.cgi?id=1410964 for example.

gusev-vitaliy:
Ok. Will try to setup it and reproduce. I guess behaviour should be the same.

----
Vitaliy Gusev

gusev-vitaliy:
Vadim, 

I reproduced corruption on Qemu, with both “raw” and “qcow2” formats.

To speedup corruption just run command in PowerShell (Replace F with your drive name) :

     Optimize-Volume -DriveLetter F -ReTrim -Verbose

Wait for 20-30 seconds and reboot VM.

Please ensure system supporting TRIM for that drive:

    C:\>fsutil fsinfo sectorInfo F:

It should show "Trim Supported” and "LogicalBytesPerSector :  4096"

I added “discard=unmap”  to the qemu’s drive option, but it seems as default value:

     -drive file=./disk2.img,format=raw,if=none,id=drive-virtio-disk2,discard=unmap

Result for “raw” format:  VM OS saw Drive F totally corrupted, i.e. “Not Initialized”.

For “qcow2” format result was less dramatic (maybe because data blocks haven’t been really unmapped and were alive), but VM OS saw corruption too. 

----
Vitaliy Gusev

[...]

PR #514, opened 2020-10-22
viostor: 4Kn disk corruption after DISCARD/UNMAP

gusev-vitaliy:
For DISCARD operation correct disk's sectors into virtio units (512-byte)
and vice versa. For 4Kn virtual disks factor will be 8, for disks with
512-byte sector size factor will be 1.

https://github.com/virtio-win/kvm-guest-drivers-windows/issues/513

Signed-off-by: Vitaliy Gusev <gusev.vitaliy@vstack.com>


vrozenfe:
verified by QE https://bugzilla.redhat.com/show_bug.cgi?id=1890810#c12

PR #516, opened 2020-11-03
This PR set supports of PVPANIC_CRASHLOADED in the pvpanic driver

annie-li:
This patch set supports PVPANIC_CRASHLOADED in the pvpanic driver, and it contains 3 patches.
The 1/3 patch unifies the code style.
The 2/3 patch requires the patch posted for QEMU - https://lore.kernel.org/qemu-devel/1604346639-27090-1-git-send-email-alejandro.j.jimenez@oracle.com/
In Patch 3/3, when the crash dump is enabled in Windows guest, PVPANIC_CRASHLOADED will be triggered before the crash dump. It indicates a crash dump file will be generated in the guest. Otherwise, PVPANIC_PANICKED will be emitted. This means the guest crashes without a crash dump file created.

### DIFF
--- a/pvpanic/pvpanic/bugcheck.c
+++ b/pvpanic/pvpanic/bugcheck.c
@@ -36,22 +36,22 @@ KBUGCHECK_CALLBACK_ROUTINE PVPanicOnBugCheck;
 
 VOID PVPanicOnBugCheck(IN PVOID Buffer, IN ULONG Length)
 {
-	if ((Buffer != NULL) && (Length == sizeof(PVOID)))
-	{
-		PUCHAR PortAddress = (PUCHAR)Buffer;
-		WRITE_PORT_UCHAR(PortAddress, (UCHAR)(PVPANIC_PANICKED));
-	}
+    if ((Buffer != NULL) && (Length == sizeof(PVOID)))
+    {
+        PUCHAR PortAddress = (PUCHAR)Buffer;
+        WRITE_PORT_UCHAR(PortAddress, (UCHAR)(PVPANIC_PANICKED));
+    }
 }
 
 BOOLEAN PVPanicRegisterBugCheckCallback(IN PVOID PortAddress)
 {
-	KeInitializeCallbackRecord(&CallbackRecord);
+    KeInitializeCallbackRecord(&CallbackRecord);
 
-	return KeRegisterBugCheckCallback(&CallbackRecord, PVPanicOnBugCheck,
-		(PVOID)PortAddress, sizeof(PVOID), (PUCHAR)("PVPanic"));
+    return KeRegisterBugCheckCallback(&CallbackRecord, PVPanicOnBugCheck,
+        (PVOID)PortAddress, sizeof(PVOID), (PUCHAR)("PVPanic"));
 }
 
 BOOLEAN PVPanicDeregisterBugCheckCallback()
 {
-	return KeDeregisterBugCheckCallback(&CallbackRecord);
+    return KeDeregisterBugCheckCallback(&CallbackRecord);
 }
--- a/pvpanic/pvpanic/bugcheck.c
+++ b/pvpanic/pvpanic/bugcheck.c
@@ -31,27 +31,71 @@
 #include "bugcheck.tmh"
 
 static KBUGCHECK_CALLBACK_RECORD CallbackRecord;
+static KBUGCHECK_REASON_CALLBACK_RECORD DumpCallbackRecord;
 
 KBUGCHECK_CALLBACK_ROUTINE PVPanicOnBugCheck;
+KBUGCHECK_REASON_CALLBACK_ROUTINE PVPanicOnDumpBugCheck;
 
 VOID PVPanicOnBugCheck(IN PVOID Buffer, IN ULONG Length)
 {
-    if ((Buffer != NULL) && (Length == sizeof(PVOID)))
+    //Trigger the PVPANIC_PANICKED event if the crash dump isn't enabled,
+    if ((Buffer != NULL) && (Length == sizeof(PVOID)) && !bEmitCrashLoadedEvent)
     {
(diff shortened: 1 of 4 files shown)


YanVugenfirer:
HI @annie-li,

Is there some more documentation other than https://lore.kernel.org/qemu-devel/1604346639-27090-1-git-send-email-alejandro.j.jimenez@oracle.com/?  I want to understand what should be the right behavior after the driver notifies that OS is in the process of saving the crash. Should it still notify the guest crash at the end of the process (marking the place where it will be safe to kill the guest VM process)?

Best regards,
Yan.

annie-li:
Hi Yan,

The document of pvpanic in QEMU side is following, 
https://github.com/qemu/qemu/blob/master/docs/specs/pvpanic.txt
The related patch of supporting PVPANIC_CRASHLOADED in QEMU is here,
https://github.com/qemu/qemu/commit/7dc58deea79a343ac3adc5cadb97215086054c86

PVPANIC_CRASHLOADED event is sent out before the crash dump. It notifies user that a crash dump file is generated, and the guest keeps running to dump the memory. After crash dump is done, the Windows guest will reboot automatically if the "Automatically restart" is configured in Windows guest. No event is sent after the crash dump is done. See function qemu_system_guest_crashloaded that gets called in QEMU after PVPANIC_CRASHLOADED is triggered.
https://github.com/qemu/qemu/blob/master/softmmu/vl.c#L1465

PVPANIC_PANICKED event only notifies user that the guest crashes. With no-shutdown option, the guest stays in "STOP" state. Without no-shutdown option, the guest will be shutdown. See function qemu_system_guest_panicked that gets calls in QEMU after PVPANIC_PANICKED  is triggered.
https://github.com/qemu/qemu/blob/master/softmmu/vl.c#L1428

Thanks
Annie

YanVugenfirer:
Hi Annie,

Thank you for the information.

Best regards,
Yan.

YanVugenfirer (approved):
Waiting for CI to finish the run.

YanVugenfirer:
@annie-li Hi Annie, I have a question about the PR and the driver behavior.

```
if ((features & (PVPANIC_PANICKED | PVPANIC_CRASHLOADED))
                       == (PVPANIC_PANICKED | PVPANIC_CRASHLOADED))
    {
        bSupportCrashLoaded = TRUE;
        TraceEvents(TRACE_LEVEL_INFORMATION, DBG_POWER,
            "PVPANIC_PANICKED and PVPANIC_CRASHLOADED notification features are supported.");
    }
    else if ((features & PVPANIC_PANICKED) == PVPANIC_PANICKED)
    {
        TraceEvents(TRACE_LEVEL_INFORMATION, DBG_POWER,
            "PVPANIC_PANICKED notification feature is supported.");
    }
```
Why bSupportCrashLoaded is enabled only if both flags are set?
QEMU  command line enables the selection of only one of the events. And if PVPANIC_CRASHLOADED selected alone, the driver will not notify it.

Best regards,
Yan.

annie-li:
@YanVugenfirer Hi Yan, By default, PVPANIC_PANICKED is supported in QEMU. PVPANIC_CRASHLOADED is the feature supported in later version of QEMU. 'bSupportCrashLoaded' represents PVPANIC_CRASHLOADED is supported besides the original PVPANIC_PANICKED feature.
Theses features can be read from the IO port. See the code in hw/misc/pvpanic.c of older QEMU version, pvpanic_ioport_read only return PVPANIC_PANICKED. In newer version of QEMU, PVPANIC_PANICKED | PVPANIC_CRASHLOADED is returned. The above code in Windows pvpanic driver is meant to check if these features are supported in QEMU itself. Related patch is https://lore.kernel.org/qemu-devel/1604346639-27090-1-git-send-email-alejandro.j.jimenez@oracle.com/, the final patch in upstream is https://www.mail-archive.com/qemu-devel@nongnu.org/msg758879.html.
With QEMU only supporting PVPANIC_PANICKED, Windows pvpanic driver only triggers the PVPANIC_PANICKED event. With QEMU supporting both PVPANIC_PANICKED and PVPANIC_CRASHLOADED, Windows pvpanic driver triggers PVPANIC_CRASHLOADED when crashdump file is being generated, otherwise, PVPANIC_PANICKED event is triggered.
In QEMU command line, the option of enabling pvpanic is '-device pvpanic'. What is the option that you mentioned "QEMU command line enables the selection of only one of the events"?

Thanks
Annie

YanVugenfirer:
@annie-li  Hi Annie,

> What is the option that you mentioned "QEMU command line enables the selection of only one of the events"?

You can use the "events" command-line option to set a specific feature. The default is "3" (PVPANIC_PANICKED  and PVPANIC_CRASHLOADED are set). 
```-device pvpanic, events=2``` - is a valid option from the QEMU point of view today. 
So either the driver is such case should trigger PVPANIC_CRASHLOADED or QEMU should not allow such option. 

Best regards,
Yan.

aljimenezb:
Hi Annie, Yan,
When I described the behavior of the QEMU patches to Annie, I missed the fact that the events property of the device could be used to specify a "non-default" value for the capabilities.
I believe the reason that Paolo added the events property to the original patch I submitted was to be able to restrict it to new machines types (he never replied to my question in: https://lore.kernel.org/qemu-devel/112eba9b-0305-f40d-8b81-6b3e46d62d75@oracle.com/)
But as Yan points out, the ability to specify a event mask with only PVPANIC_CRASHLOADED is available, even though it might not have many use cases.

annie-li:
I see. 'event = 2' scenario is missed. I'll fix it.
Current driver covers scenario of 'event = 1' and 'event = 3'. From debug point of view, user may not intend to only choose 'event = 2' since the event is only triggered when crashdump file is generated. But it is better to deal with this scenario just in case.

YanVugenfirer:
@aljimenezb Hi Alejandro,

Thank you for the clarification.

PR #518, opened 2020-11-13
balloon: skip reporting to the host when it is not available

ybendito:
https://bugzilla.redhat.com/show_bug.cgi?id=1828654

When the device was surprisingly removed the driver begins
removal flow and the host is not available. The driver
tries to report freed pages, each report takes 1 second
(timeout value), so the driver might not work for several
minutes. Skip reporting to host after surprise removal detected.


ybendito:
rerun test

ybendito:
rerun tests

YanVugenfirer (approved):

PR #589, opened 2021-06-17
viofs-svc: fix commit 298ff119 (resolve links on path walkthough)

hammerg:
File's path was not null-terminated in the Lookup request.

Signed-off-by: Gal Hammer <ghammer@redhat.com>

### DIFF
--- a/viofs/svc/virtiofs.c
+++ b/viofs/svc/virtiofs.c
@@ -637,6 +637,8 @@ static NTSTATUS PathWalkthough(HANDLE Device, CHAR *FullPath,
                 break;
             }
 
+            SubstituteName[SubstituteNameLength] = L'\0';
+
             Status = VirtFsLookupFileName(Device, SubstituteName, &LookupOut);
             if (!NT_SUCCESS(Status))
             {


viofs/svc/virtiofs.c
```diff
@@ -637,6 +637,8 @@ static NTSTATUS PathWalkthough(HANDLE Device, CHAR *FullPath,
                 break;
             }
 
+            SubstituteName[SubstituteNameLength] = L'\0';
```
YanVugenfirer, line 640: What if SubmitReadLinkRequest returned MAX_PATH as SubstituteNameLength?

hammerg: The name length received from the back-end doesn't include the null-terminated char. So the return result from the MultiByteToWideChar  doesn't include it as well. So in case of MAX_PATH we're still remain within limit.

YanVugenfirer: So if the back-end doesn't include null-terminated char, then couldn't FUSE_READLINK_OUT include "name" that is completely filled out? 

```
typedef struct{    
struct fuse_out_header  hdr; 
char                    name[MAX_PATH];
} FUSE_READLINK_OUT, *PFUSE_READLINK_OUT;
```
https://github.com/virtio-win/kvm-guest-drivers-windows/blob/775ec67d8391a5d5884687cf3ab7c9da9c2cfd16/viofs/svc/fusereq.h#L255

 And then MultiByteToWideChar will fill out the buffer without space left:
``` 
 *SubstituteNameLength = (USHORT)MultiByteToWideChar(CP_UTF8, 0,            readlink_out.name, namelen, SubstituteName, MAX_PATH);
```

hammerg: It is a typo. The back-end send the file name *with* a null-terminated char, limited by MAX_PATH.

PR #594, opened 2021-07-05
Sdv2022

irudakov77:
Preparation steps for generating DVLs for System class devices required by  WHQL for Windows 2022. Run SDV, cleanup, additional build scripts, no integration with CodeQL yet.

### DIFF
--- a/Balloon/sys/cleanAll.bat
+++ b/Balloon/sys/cleanAll.bat
@@ -2,5 +2,13 @@
 
 for /d %%x in (objfre_*) do rmdir /S /Q %%x
 for /d %%x in (objchk_*) do rmdir /S /Q %%x
+rmdir /S /Q .\sdv
+rmdir /S /Q .\sdv.temp
 
 del /F *.log *.wrn *.err
+
+del *.dvl.xml
+del sdv-map.h
+del sdv-user.sdv
+del SDV-default.xml
+
--- a/Balloon/sys/cleanAll.bat
+++ b/Balloon/sys/cleanAll.bat
@@ -8,6 +8,7 @@ rmdir /S /Q .\sdv.temp
 del /F *.log *.wrn *.err
 
 del *.dvl.xml
+del *.dvl-compat.xml
 del sdv-map.h
 del sdv-user.sdv
 del SDV-default.xml
--- a/buildAll.bat
+++ b/buildAll.bat
@@ -11,8 +11,25 @@ call tools\build.bat vioscsi\vioscsi.vcxproj "Win8_SDV Win10_SDV" %*
 if errorlevel 1 goto :fail
 call tools\build.bat viostor\viostor.vcxproj "Win8_SDV Win10_SDV" %*
 if errorlevel 1 goto :fail
+if "%VIRTIO_WIN_SDV_2022%"=="" goto :nosdv2022
+call tools\build.bat Balloon\sys\balloon.vcxproj "Win10_SDV" %*
+if errorlevel 1 goto :fail
+call tools\build.bat fwcfg64\fwcfg.vcxproj "Win10_SDV" %*
+if errorlevel 1 goto :fail
 call tools\build.bat ivshmem\ivshmem.vcxproj "Win10_SDV" %*
 if errorlevel 1 goto :fail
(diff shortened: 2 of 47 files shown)


buildAll.bat
```diff
@@ -11,8 +11,25 @@ call tools\build.bat vioscsi\vioscsi.vcxproj "Win8_SDV Win10_SDV" %*
 if errorlevel 1 goto :fail
 call tools\build.bat viostor\viostor.vcxproj "Win8_SDV Win10_SDV" %*
 if errorlevel 1 goto :fail
+if "%VIRTIO_WIN_SDV_2022%"=="" goto :nosdv2022
```
kostyanf14, line 14: I think buildAll.bat should always run SDV tests. 
@YanVugenfirer what do you think?

Balloon/sys/cleanAll.bat
```diff
@@ -2,5 +2,13 @@
 
 for /d %%x in (objfre_*) do rmdir /S /Q %%x
 for /d %%x in (objchk_*) do rmdir /S /Q %%x
+rmdir /S /Q .\sdv
+rmdir /S /Q .\sdv.temp
 
 del /F *.log *.wrn *.err
+
+del *.dvl.xml
```
kostyanf14, line 10: Please also remove *.dvl-compat.xml. We create *.dvl-compat.xml for compatibility with Windows 8.

kostyanf14 (commented):
In general LGTM

kostyanf14:
All other CI tests canceled.

Balloon/sys/cleanAll.bat
irudakov77: Done

buildAll.bat
YanVugenfirer: > I think buildAll.bat should always run SDV tests.
> @YanVugenfirer what do you think?

I am not sure yet.
Depending on if we want to break the build into two steps or not.
For now, I would prefer to keep it as it is in the patch.

kostyanf14: VIRTIO_WIN_SDV_2022 variable looks very specifically. We also will run all SDV for Windows 11. Maybe better to use VIRTIO_WIN_ALL_SDV or something like this?

buildAll.bat
```diff
@@ -11,8 +11,25 @@ call tools\build.bat vioscsi\vioscsi.vcxproj "Win8_SDV Win10_SDV" %*
...
+call tools\build.bat pvpanic\pvpanic\pvpanic.vcxproj "Win10_SDV" %*
+if errorlevel 1 goto :fail
+call tools\build.bat viocrypt\sys\viocrypt.vcxproj "Win10_SDV" %*
+if errorlevel 1 goto :fail
+call tools\build.bat viorng\viorng\viorng.vcxproj "Win10_SDV" %*
+if errorlevel 1 goto :fail
+call tools\build.bat vioserial\sys\vioser.vcxproj "Win10_SDV" %*
+if errorlevel 1 goto :fail
+call tools\build.bat viosock\sys\viosock.vcxproj "Win10_SDV" %*
+if errorlevel 1 goto :fail
```
kostyanf14, line 30: There are missing 3 drivers: viogpu, vioinput and viofs

```
call tools\build.bat viogpu\viogpudo\viogpudo.vcxproj "Win10_SDV" %*
if errorlevel 1 goto :fail
call tools\build.bat vioinput\sys\vioinput.vcxproj "Win10_SDV" %*
if errorlevel 1 goto :fail
call tools\build.bat viofs\pci\viofs.vcxproj "Win10_SDV" %*
if errorlevel 1 goto :fail
```

buildAll.bat
irudakov77: The first idea was to enable SDV only for drivers that really need to have valid DVL to pass Win2022 Static Logo Test. Because the static analysis takes quite a long time, and here, at Virtuozzo, we found that DVL for the Display class drivers, for example, is still not required. But if all decide that now we will always run SDV for each driver, then it is reasonable to use VIRTIO_WIN_ALL_SDV.

buildAll.bat
irudakov77: Only viofs is really missing, the rest are skipped for the reason, mentioned above.:)

YanVugenfirer: @kostyanf14 we can add it in the next PR.

PR #596, opened 2021-07-07
vioscsi: Increasing max phys breaks to 512

irudakov77:
The maximum block size has been increased to 512 pages and is now the same as Viostor's maximum block size. This can give io performance gains of up to 30% on fast storage. Dynamic memory allocation for sglist has been removed because the SRB extension now has enough pre-allocated space to hold the complete sglist. In addition, there was a bug in this code that led to unnecessary allocation and copying for requests of MAX_PHYS_SEGMENTS size, which also reduced performance.
We found that in some cases Vioscsi is up to 4 times slower than Viostore, setting the same number of max physical breaks (512) reduces this difference up to 2,5 times. :)
PS: Several months ago, @vrozenfe kindly suggested increasing the number of physical breaks to 512, but then it seemed unnecessary.

### DIFF
--- a/vioscsi/helper.c
+++ b/vioscsi/helper.c
@@ -209,7 +209,6 @@ ENTER_FN();
 
     srbExt->psgl = srbExt->vio_sg;
     srbExt->pdesc = srbExt->desc_alias;
-    srbExt->allocated = 0;
     sgElement = 0;
     srbExt->psgl[sgElement].physAddr = StorPortGetPhysicalAddress(DeviceExtension, NULL, &cmd->req.tmf, &fragLen);
     srbExt->psgl[sgElement].length   = sizeof(cmd->req.tmf);
--- a/vioscsi/vioscsi.c
+++ b/vioscsi/vioscsi.c
@@ -331,7 +331,7 @@ BOOLEAN VioScsiReadRegistry(
            sizeof(ULONG));
     adaptExt->max_physical_breaks = min(
                                         max(SCSI_MINIMUM_PHYSICAL_BREAKS, adaptExt->max_physical_breaks),
-                                        SCSI_MAXIMUM_PHYSICAL_BREAKS);
+                                        MAX_PHYS_SEGMENTS);
 
     StorPortFreeRegistryBuffer(DeviceExtension, pBuf );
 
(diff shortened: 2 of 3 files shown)


vrozenfe:
Thanks, @irudakov77 

This is the right direction. We needed to to do it long time ago.

However several things seems to be missed.

We do know that 512 + 1 physical breaks works fine on Win 10 ( and this is what MSFT seems to be using in their nvme driver)
_NVMeHwFindAdapter@24:
......
  00011B12: E8 B5 F5 FF FF     call        
_GetControllerMaxTransferSize@4
  00011B17: 89 46 18           mov         dword ptr [esi+18h],eax
  00011B1A: 6A 20              push        20h
  00011B1C: 58                 pop         eax
  00011B1D: C7 46 1C 01 02 00  mov         dword ptr [esi+1Ch],201h  <--PhysicalBreaks

But we have no idea if the same limit can be used on pre-Win10 systems. 
Did  you managed to check 512 physical breaks on Win8/8.1 and WS2012/(R2) systems?

Second, if you want to get rid of the dynamic allocation in vioscsi driver code entirely and make it equal with viostor on this matter,
then all the relevant code needs to be cleaned out , including

https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/vioscsi/vioscsi.h#L245

and the following block
https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/vioscsi/vioscsi.c#L1628
.....
https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/vioscsi/vioscsi.c#L1637

Cheers,
Vadim.

irudakov77:
Hello, Vadim.

> But we have no idea if the same limit can be used on pre-Win10 systems.
> Did you managed to check 512 physical breaks on Win8/8.1 and WS2012/(R2) systems?
I'm on the way, give me a day pls. But viostor easily uses this value for all OS greater than win7, why shouldn't vioscsi work fine too?:)

Regards,
Ilya

vrozenfe:
Hi Ilya,
  But viostor easily uses this value for all OS greater than win7, why shouldn't vioscsi work fine too?:)

I hope that it works for Win8 as well, but I cannot say it for sure.  

The following patch 
https://github.com/virtio-win/kvm-guest-drivers-windows/pull/487 
was tested quite well on Win10 platforms under the different load scenarios. 
Unfortunately,  I cannot say the same about Win8. 

So, it will be nice if you can run some sort of testing to see that Win8.1
and WS2012 are working fine with your changes.

Other than that the patch itself looks good to me.

Cheers,
Vadim.

vrozenfe:
@irudakov77 

Merging upstream with request to check this PR on Win8/Win8.1/WS2012(R2)  platforms  properly.

Best,
Vadim.

irudakov77:
Hello Vadim,

Thanks for waiting. I found that MaxPhysicalBreakes is limited to the NumberOfMapRegisters returned from IoGetDmaAdapter being called inside storport. On win2012 with 4 GB RAM NumberOfMapRegisters is 0x80001 :)

Best regards,
Ilya

PR #600, opened 2021-07-14
NetKVM: RHBZ#1979024: restrict max VLAN ID to 4094

viktor-prutyanov:
NetKVM treats VLAN ID = 4095 as well as IDs from 1-4094, that is frames
with ID = 4095 pass and others don't. According to IEEE 802.1Q, 4095
can't be valid VLAN ID. Also, Win2022 HLK assumes that NDIS accepts
ID <= 4094.

Signed-off-by: Viktor Prutyanov <viktor.prutyanov@phystech.edu>

### DIFF
--- a/NetKVM/Common/ParaNdis_Common.cpp
+++ b/NetKVM/Common/ParaNdis_Common.cpp
@@ -40,7 +40,7 @@ static VOID ParaNdis_UpdateMAC(PARANDIS_ADAPTER *pContext);
 
 static __inline pRxNetDescriptor ReceiveQueueGetBuffer(PPARANDIS_RECEIVE_QUEUE pQueue);
 
-#define MAX_VLAN_ID     4095
+#define MAX_VLAN_ID     4094
 
 /**********************************************************
 Validates MAC address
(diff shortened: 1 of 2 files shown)


NetKVM/Common/ParaNdis_Common.cpp
```diff
@@ -1899,8 +1899,8 @@ static VOID SetAllVlanFilters(PARANDIS_ADAPTER *pContext, BOOLEAN bOn)
 /*
     possible values of filter set (pContext->ulCurrentVlansFilterSet):
     0 - all disabled
-    1..4095 - one selected enabled
-    4096 - all enabled
+    1..4094 - one selected enabled
+    4095 - all enabled
```
YanVugenfirer, line 1903: But now 4095 will not be used, right? Or in case of 0, we will set to 4095 the filter value?

viktor-prutyanov: > But now 4095 will not be used, right? 

It will not be used as VlanId, but used as filter wildcard (instead of 4096)

> Or in case of 0, we will set to 4095 the filter value?

Yes

Before the patch:
4095 is valid VlanId, 0 sets filter to 4096 (wildcard value)
After:
4095 is not valid VlanId, 0 sets filter to 4095 (wildcard value)

YanVugenfirer (approved):

PR #601, opened 2021-07-15
NetKVM: set NumRssQueues default and max values

viktor-prutyanov:
Win2022 HLK (NICStrictPropertyValidation, Test-NetHLK) requires NumRssQueues parameter default value to be 16 for 10Gbps adapters with MSI-X support.
Also, set NumRssQueues max value to 32.


YanVugenfirer (commented):
I think this is good timing to evaluate the needed MAX number of queues as well.

PR #608, opened 2021-07-26
Myewdk2022

irudakov77:
Universal build environment, we can now build from EWDK and the current VS2017 environment. Just run a build from EWDK env or set EWDK_DIR to switch build env. I checked EWDK 2022 and EWDK 1903, everything seems to be working fine. I didn't do anything with build warnings.

### DIFF
--- a/Tools/build.bat
+++ b/Tools/build.bat
@@ -29,6 +29,7 @@ set SUPPORTED_BUILD_SPECS=WinXP_wxp Win2k3_wnet Vista_wlh Win7_win7 Win8_win8 Wi
 set BUILD_TARGETS=%~2
 set BUILD_DIR=%~dp1
 set BUILD_FILE=%~nx1
+set BUILD_NAME=%~n1
 
 rem We do an incremental Release build for all specs and all archs by default
 set BUILD_FLAVOR=Release
@@ -189,6 +190,34 @@ set __VERBOSITY__=n
 msbuild.exe %BUILD_FILE% /t:%__TARGET__% /p:Configuration="%~1" /P:Platform=%2 -fileLoggerParameters:Verbosity=%__VERBOSITY__%;LogFile=%BUILD_LOG_FILE%
 goto :eof
 
+:run_semmle
+if "%SEMMLE_HOME%"=="" (
+  echo "Semmle path is not configured, skipping"
+  goto :eof
+)
+
+echo "Removing previously created rules database"
+rmdir /s/q semmle_db
+
+echo "Creating rules database for %BUILD_FILE%"
+echo msbuild.exe %BUILD_FILE% /t:rebuild /p:Configuration="%~1" /P:Platform=%2 > semmle_bld.bat
+call %SEMMLE_HOME%\codeql\codeql.exe database create -l=cpp -c "semmle_bld.bat" semmle_db -j 0
+call del semmle_bld.bat
+
+IF ERRORLEVEL 1 (
+  set BUILD_FAILED=1
+  goto :eof
+)
+
+echo "Analyzing rules database for %BUILD_FILE%"
+call %SEMMLE_HOME%\codeql\codeql.exe database analyze semmle_db "%SEMMLE_HOME%\Windows-Driver-Developer-Supplemental-Tools\codeql\windows-drivers\suites\windows_driver_mustfix.qls" --format=sarifv2.1.0 --output="%BUILD_NAME%.sarif" -j 0 --rerun
+
+IF ERRORLEVEL 1 (
+  set BUILD_FAILED=1
+)
+
+goto :eof
+
 :configure_sdv
 echo Set SDV mode to LEGACY
 set SDVLEGACY=1
@@ -217,6 +246,11 @@ IF ERRORLEVEL 1 (
 
 msbuild.exe %BUILD_FILE% /t:sdv /p:inputs="/check /devenv" /p:Configuration="%~1" /P:platform=%2
 
+IF ERRORLEVEL 1 (
+  set BUILD_FAILED=1
+)
+
+call :run_semmle %1 %2
 IF ERRORLEVEL 1 (
   set BUILD_FAILED=1
 )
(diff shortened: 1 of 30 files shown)


kostyanf14:
Hi @irudakov77,

Please look at https://github.com/virtio-win/kvm-guest-drivers-windows/pull/602.

irudakov77:
> Hi @irudakov77,
> 
> Please look at #602.

Hi @kostyanf14,

I've seen this patch before. A lot of work has been done, the build env has been switched to EVDK, the CPDK has been updated to 10 (by the way, where did you get it? I can't find it yet. :)), build scripts have been fixed, features have been added to netkvm. My patch is not a substitute, but a supplement. Here at Virtuozzo, we need a solution that allows us to try EWDK 2022 (for SDV, for example) without breaking the VS2017 build. Perhaps this idea will seem interesting to someone, I do not insist on approving it for master, so the patch is targeted to the ewdk2022 branch, and now the branch can be built in this environment without issues, by the way.

Regards,
Ilya

kostyanf14:
Hi @irudakov77,

Thanks for your answer.

>  the CPDK has been updated to 10 (by the way, where did you get it? I can't find it yet. :)

Please check [wiki](https://github.com/virtio-win/kvm-guest-drivers-windows/wiki/Building-the-drivers-using-EWDK-1903-2004). EWDK build works fine after all preparation. We use Windows Server 2016 on this machine. 

> Here at Virtuozzo, we need a solution that allows us to try EWDK 2022 (for SDV, for example) without breaking the VS2017 build

Thanks for the explanation. 

Konstantin.

YanVugenfirer:
@irudakov77 Thanks for sharing!

BWT: did you experience some stability issues with eDWK2022? We are having issues with ARM64 build.

irudakov77:
Hi @YanVugenfirer,

We are not yet building ARM64 drivers in Virtuozzo, however, I just ran ARM64 build locally and I didn't see any issues.

kostyanf14:
Hi @irudakov77,

Do you run ARM64 build in eWDK 2022 environment or in VS2017 environment?

irudakov77:
@kostyanf14, indeed, I was not careful, my environment was switched to EWDK1903. With EWDK2022 I see NetKVM ProtocolService build errors due to missing MFC libraries with Specter Mitigations.

kostyanf14:
@irudakov77 Yes, we see the same. I go deeper into EWDK2022 and found that all MFC libraries for ARM64 are present. But MSBuild can't find it. So looks like this is EWDK bug.

YanVugenfirer:
I think I will keep PR pending, so the moment we have eWDK branch merged and move to deal with eWDK 2022, we will not forget to integrate the changes.
@irudakov77 any objections?

irudakov77:
@YanVugenfirer,
No objections, thanks.

Tools/build.bat
```diff
@@ -189,6 +190,34 @@ set __VERBOSITY__=n
...
+  echo "Semmle path is not configured, skipping"
+  goto :eof
+)
+
+echo "Removing previously created rules database"
+rmdir /s/q semmle_db
+
+echo "Creating rules database for %BUILD_FILE%"
+echo msbuild.exe %BUILD_FILE% /t:rebuild /p:Configuration="%~1" /P:Platform=%2 > semmle_bld.bat
+call %SEMMLE_HOME%\codeql\codeql.exe database create -l=cpp -c "semmle_bld.bat" semmle_db -j 0
```
kostyanf14, line 204: I think it better uses the current project as a `semmle_db` name. This allows saving all codeql logs in case of buildAll was called.

This is my draft for codeql https://github.com/kostyanf14/kvm-guest-drivers-windows/commit/a20aec94a98eeee52ef2d47fc5ff468d3fc31119

Tools/build.bat
```diff
@@ -189,6 +190,34 @@ set __VERBOSITY__=n
...
+call del semmle_bld.bat
+
+IF ERRORLEVEL 1 (
+  set BUILD_FAILED=1
+  goto :eof
+)
+
+echo "Analyzing rules database for %BUILD_FILE%"
+call %SEMMLE_HOME%\codeql\codeql.exe database analyze semmle_db "%SEMMLE_HOME%\Windows-Driver-Developer-Supplemental-Tools\codeql\windows-drivers\suites\windows_driver_mustfix.qls" --format=sarifv2.1.0 --output="%BUILD_NAME%.sarif" -j 0 --rerun
```
kostyanf14, line 213: I preferred to use all suites, not only windows_driver_mustfix. DVL ignores errors that are not **mustfix**, but we can look at it.

irudakov77: Agree

Tools/build.bat
irudakov77: But it seems to me that for building individual drivers and for cleanup it is more convenient to create a codeql base inside the driver directory as SDV does.

kostyanf14: It makes sense. I agree.

kostyanf14:
@irudakov77

I preparing the branch for Windows 11 EWDK. ARM build and viogpu build were fixed already. CodeQL tools integrated.

Please rebase your `autoselect build env` commits to it.

build: autoselect build env
build: Auto target for UM apps
build: Retarget UM app projects to Auto toolset ver 

https://github.com/virtio-win/kvm-guest-drivers-windows/tree/ewdk11

irudakov77:
@kostyanf14 

Hi, Kostiantyn

Thanks for waiting, I see I need to rebase the patches already on the "master" branch. :) This is not a problem, but I want to ask, do we really need the "autoselect build env" patch? It changes the behavior of the build script: if EWDK_DIR is not defined, I do not set it to C:\ewdk11, but use VS2017 to build instead. I like this approach, but maybe you don't expect it.

Regards,
Ilya

kostyanf14:
@irudakov77 

Hi Ilya,

I think the "autoselect build env" patch is good for upstream. Someone may want to build drivers from the console in VS env.

>if EWDK_DIR is not defined, I do not set it to C:\ewdk11, but use VS2017 to build instead

This is ok. CI will define EWDK11 dir, just need to update the wiki when the patch will be merged.

YanVugenfirer:
Changes are already merged or in other pending pull requests

PR #611, opened 2021-07-27
Rss fix

ybendito:


ybendito:
2019: Mni6RssOid - requires 16 CPUs to test 16 queues
2022:
NICStrictPropertyValidation - no logs (would be good to rerun to see the failure reason)
Hardware-enforced Stack Protection Compatibility Test - no logs, but this test is manual, please check why it runs

YanVugenfirer:
> 2019: Mni6RssOid - requires 16 CPUs to test 16 queues
> 2022:
> NICStrictPropertyValidation - no logs (would be good to rerun to see the failure reason)
The test should be separately installed
> Hardware-enforced Stack Protection Compatibility Test - no logs, but this test is manual, please check why it runs
@kostyanf14 please check what happen.

kostyanf14:
>NICStrictPropertyValidation - no logs (would be good to rerun to see the failure reason)

We have an issue in PlanIO about this test. Some tool is missing.

PR #615, opened 2021-08-02
[vioscsi] Firmware update handler. Bugfix for bz#1983901 HLK S"torage HBA Firmware Update" job fails

vrozenfe:


vioscsi/vioscsi.h
```diff
@@ -45,6 +45,10 @@ typedef struct VirtIOBufferDescriptor VIO_SG, *PVIO_SG;
 #define VIRTIO_SCSI_CDB_SIZE   32
 #define VIRTIO_SCSI_SENSE_SIZE 96
 
+#ifndef NTDDI_WINTHRESHOLD
+#define NTDDI_WINTHRESHOLD                  0x0A000000  /* ABRACADABRA_THRESHOLD */
```
YanVugenfirer, line 49: :)

vrozenfe: This part is needed to make sure that the code can be compiled successfully  for Vista targets, which is building with old 
WDK tools. WinDDK 7600, which we use in this case, has Windows7 as the most advanced version.

PR #620, opened 2021-08-07
PVPanic: RHBZ#1826336: revert "pvpanic: Add IOCTL_GET_CRASH_DUMP_HEADER"

viktor-prutyanov:
First of all, according to Microsoft Driver Security Guidance driver shouldn't provide named device object (`\\.\PVPanicDevice`). Besides of that, no one uses PVPanic device in userspace. So, remove device object symbolic link.


YanVugenfirer (commented):
As far as I know, no one in userspace should use PVPanic device.

viktor-prutyanov:
> As far as I know, no one in userspace should use PVPanic device.
> 

@YanVugenfirer may be remove device object link at all?

YanVugenfirer:
> > As far as I know, no one in userspace should use PVPanic device.
> 
> @YanVugenfirer may be remove device object link at all?

@viktor-prutyanov Good idea.

PR #622, opened 2021-08-09
fwcfg: fix issues found by CodeQL

viktor-prutyanov:
- Replace ExAllocatePoolWithTag by ExAllocatePoolUninitialized
- Fix comparison of different types

### DIFF
--- a/fwcfg64/driver.c
+++ b/fwcfg64/driver.c
@@ -72,7 +72,7 @@ NTSTATUS GetKdbg(PDEVICE_CONTEXT ctx)
     CONTEXT context = { 0 };
     NTSTATUS status = STATUS_SUCCESS;
 
-    minidump = ExAllocatePoolWithTag(NonPagedPoolNx, 0x40000, 'pmdm');
+    minidump = ExAllocatePoolUninitialized(NonPagedPoolNx, 0x40000, 'pmdm');
     if (!minidump)
     {
          return STATUS_MEMORY_NOT_ALLOCATED;
@@ -87,7 +87,7 @@ NTSTATUS GetKdbg(PDEVICE_CONTEXT ctx)
         "KdDebuggerDataBlock size = %lx, offset = 0x%lx",
         kdbg_size, kdbg_offset);
 
-    ctx->kdbg = ExAllocatePoolWithTag(NonPagedPoolNx, kdbg_size, 'gbdk');
+    ctx->kdbg = ExAllocatePoolUninitialized(NonPagedPoolNx, kdbg_size, 'gbdk');
     if (!ctx->kdbg)
     {
         status = STATUS_MEMORY_NOT_ALLOCATED;
--- a/fwcfg64/fwcfg.c
+++ b/fwcfg64/fwcfg.c
@@ -92,7 +92,7 @@ UINT32 FWCfgGetEntriesNum(PVOID ioBase)
 NTSTATUS FWCfgFindEntry(PVOID ioBase, const char *name,
                         PUSHORT index, ULONG size)
 {
-    UINT16 i;
+    UINT32 i;
     UINT32 total = FWCfgGetEntriesNum(ioBase);
     FWCfgFile f;
 


kostyanf14:
Build failed expectedly.
```
ExAllocatePoolUninitialized:

Minimum supported client: Requires WDK for Windows 10, version 2004. Targets Windows 7 and later versions of the Windows operating system.
```

Let's wait for Windows 11 EWDK build preparation and then rebuild this PR.

viktor-prutyanov:
> ```
> ExAllocatePoolUninitialized:
> 
> Minimum supported client: Requires WDK for Windows 10, version 2004. Targets Windows 7 and later versions of the Windows operating system.
> ```
> 
> Let's wait for Windows 11 EWDK build preparation and then rebuild this PR.

It turns out that the driver can only be build with WDK 2004 and build with previous WDKs will be broken

kostyanf14:
> It turns out that the driver can only be build with WDK 2004 and build with previous WDKs will be broken

Have we any other solution except building two versions of the driver?

viktor-prutyanov:
MSDN says that 

> ExAllocatePoolWithTag has been deprecated in Windows 10, version 2004

They suggest to replace the function with ExAllocatePoolZero/ExAllocatePoolUninitialized if driver is for Windows 10 less than 2004 and  ExAllocatePool2 if >= 2004. All of these functions are first appeared in WDK 2004.

It looks like either new functions from WDK 2004 are used and CodeQL passes or WDK is less than 2004 and CodeQL fails.

viktor-prutyanov:
This is the CodeQL query I'm talking about: https://github.com/microsoft/Windows-Driver-Developer-Supplemental-Tools/blob/main/codeql/windows-drivers/queries/Windows/wdk/wdk-deprecated-api.ql

kostyanf14:
> MSDN says that
> 
> > ExAllocatePoolWithTag has been deprecated in Windows 10, version 2004
> 
> They suggest to replace the function with ExAllocatePoolZero/ExAllocatePoolUninitialized if driver is for Windows 10 less than 2004 and ExAllocatePool2 if >= 2004. All of these functions are first appeared in WDK 2004.
> 
> It looks like either new functions from WDK 2004 are used and CodeQL passes or WDK is less than 2004 and CodeQL fails.

Looks like this. Another solution (https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/codeql-windows-driver-wdkdeprecatedapi#driver-updates-for-versions-of-windows-earlier-than-windows-10-version-2004) has the same problem.

viktor-prutyanov:
I think we can use ExAllocatePoolZero everywhere and build with WDK 2004. We don't know exact revision of Windows 10 on target system.

viktor-prutyanov:
This PR builds successfully with EWDK 2004

YanVugenfirer:
> > MSDN says that
> > > ExAllocatePoolWithTag has been deprecated in Windows 10, version 2004
> > 
> > 
> > They suggest to replace the function with ExAllocatePoolZero/ExAllocatePoolUninitialized if driver is for Windows 10 less than 2004 and ExAllocatePool2 if >= 2004. All of these functions are first appeared in WDK 2004.
> > It looks like either new functions from WDK 2004 are used and CodeQL passes or WDK is less than 2004 and CodeQL fails.
> 
> Looks like this. Another solution (https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/codeql-windows-driver-wdkdeprecatedapi#driver-updates-for-versions-of-windows-earlier-than-windows-10-version-2004) has the same problem.

1. The main goal is to try and avoid situations when we need to have different packages for Widows Server 2022\Windows 11 and previous versions of Windows 10. So don't forget to test binaries on previous Windows 10 releases.

2. https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/codeql-windows-driver-wdkdeprecatedapi#driver-updates-for-versions-of-windows-earlier-than-windows-10-version-2004 - let's try to follow recommendations here including ExInitializeDriverRuntime during driver initialization

kostyanf14:
@YanVugenfirer 

https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/codeql-windows-driver-wdkdeprecatedapi#driver-updates-for-versions-of-windows-earlier-than-windows-10-version-2004 - has the same problem.

ExInitializeDriverRuntime does not have a minimal SDK version in documentation, but for ExAllocatePoolZero - Requires WDK for Windows 10, version 2004.

YanVugenfirer:
Got it

viktor-prutyanov:
> 1. The main goal is to try and avoid situations when we need to have different packages for Widows Server 2022\Windows 11 and previous versions of Windows 10. So don't forget to test binaries on previous Windows 10 releases.

As I see, we don't need different packages. ExAllocatePoolZero and friends are just force inline wrappers over ExAllocatePoolWithTag. At least, this PR builds successfuly for Win7 and later

> 2. https://docs.microsoft.com/en-us/windows-hardware/drivers/devtest/codeql-windows-driver-wdkdeprecatedapi#driver-updates-for-versions-of-windows-earlier-than-windows-10-version-2004 - let's try to follow recommendations here including ExInitializeDriverRuntime during driver initialization

ExAllocatePoolZero requires ExInitializeDriverRuntime, but ExAllocateUninitialized doesn't

PR #636, opened 2021-08-22
Auto target UM apps

irudakov77:


YanVugenfirer:
@irudakov77 Sorry, the eWDK11 changes are merged into master from Konstantin's PR (we had to split some commits in the eWDK branch). Can you please resent PR to master?

YanVugenfirer:
Closing. 

Addressed by #642

ybendito:
https://github.com/virtio-win/kvm-guest-drivers-windows/commit/66d8b30a838ae77c34f36eaad1a9048b0eaed1a6#diff-aa4ba8ce0b69a55e6627c4da2d202b0fad717cdc7019c5f5aaa72ecbbbb87aed
I think this commit is problematic and probably should be reverted (just because it assumes that Studio 2017 is enough to build the drivers). 
IMO, the "build all" procedure should not be used without EWDK11.
Each specific project and also entire virtio-win solution should be buildable by Visual Studio 2019 with EWDK 22000, but we can't  make it buildable with Studio 2017 (which can use only much earlier WDK).
For example, NetKvm now requires NDIS 6.85 (i.e. at least WDK 2004).

irudakov77:
Hi, @ybendito!
Why don't we just switch from vs2017 to vs2019? Then only one character needs to be changed here.:)

ybendito:
@irudakov77 We can do that but **what is the use case**? The build using 'build all' using Visual Studio is not guaranteed even to succeed depending on exact flavor and version of Visual Studio. (We see a lot of examples and I myself use Visual Studio for partial builds of specific solution/projects). But the final result of "whether the branch/PR/commit is good" depends on "what it does when it is built using 'build all' with EWDK".
IMO, each developer can maintain his own Visual Studio environment and do with it everything he wants - this way of compilation does not provide reproducible results. Sad but true. And the current idea of build.bat is to have something that on every system behaves the same way.

YanVugenfirer:
@irudakov77 

Hi Ilya,

Some thoughts:

1. If you still need the ability to build using VS after the move to eWDK, can you please describe your use case? Maybe we are missing something.
2. I think the default should be built with eWDK. That's what we are testing and this is what we guarantee to work.
3. If you need it and can maintain the ability to work with VS, then please create a wiki page describing the installation of the build environment. We can create an appropriate CI machine to test next PRs. In such a case, I want to ask you to be responsible for the build failures fixes and maintaining the description of the build configuration.

Thanks,
Yan.

irudakov77:
@YanVugenfirer, @ybendito 

In general, in Virtuozzo we have already switched to the EWDK, but I personally use VS2019 also. I think it is good to have an alternative way to build drivers, but If nobody needs it, then I won't worry if you revert this patch :)

PR #656, opened 2021-09-18
NetKVM: Fix dropping IPv6 IPsec packets.

yach-yf:
IPv6 ESP header is treated as usual header(TCP, UDP etc.) by this change.

ESP header should not be treated like other extension headers.
Because it does not start with 'Next Header' field and next header is encrypted.
So base version of NetKVM drops IPv6 ESP(IPsec) packet as malformed packet.

### DIFF
--- a/NetKVM/Common/sw_offload.cpp
+++ b/NetKVM/Common/sw_offload.cpp
@@ -955,6 +955,8 @@ BOOLEAN AnalyzeIP6Hdr(
         case PROTOCOL_UDP:
             __fallthrough;
         case IP6_HDR_FRAGMENT:
+            __fallthrough;
+        case IP6_HDR_ESP:
             return TRUE;
         case IP6_HDR_DESTINATION:
             {
@@ -986,8 +988,6 @@ BOOLEAN AnalyzeIP6Hdr(
             }
             break;
         case IP6_HDR_HOP_BY_HOP:
-            __fallthrough;
-        case IP6_HDR_ESP:
             __fallthrough;
         case IP6_HDR_AUTHENTICATION:
             __fallthrough;


ybendito:
If I understand correctly the ESP header should have "next header" = 0 and on the next round we should return true.
Can you please refer the RFC that states that ESP IPv6 extension header does not start with 'Next Header" field? (I might be mistaken, but as far as this type of header is defined among IPv6 header it should be of the same format). Even if we should recognize it as the last one automatically, I think it would be better to make it clearly in the code and not just remove any mention of it.

yach-yf:
ESP header start with 'Security Parameters Index (SPI)' field, see the section 2 of RFC4303.
https://datatracker.ietf.org/doc/html/rfc4303#section-2

Sorry, my explanation was insufficient.
'Hdr Ext Len' field next to 'Next Header' field is the cause of ESP packet being determined to be malformed.
It is not on ESP header too.

For reference, I attached a Wireshark screen capture of an ESP packet.
IPv6_ESP_Wireshark

yach-yf:
> I think it would be better to make it clearly in the code and not just remove any mention of it.

Yes, you're right.
'Fragment Header' is so.
I re-wrote the code.
It has no effect on behavior.

kostyanf14:
rerun tests

YanVugenfirer:
@yach-yf 

Please rebase your branch and remove merge commit.
Thanks.

ybendito:
I'd suggest to make just one commit instead of 2

yach-yf:
Rebased.
I confirmed no file differences.

yach-yf:
I confirmed that IPv6 IPsec works fine at HEAD revision(328edb5c6c9ba6227e9ed00b05855add52022726).

YanVugenfirer:
@yach-yf 

Sorry, last formality before merging - please add “Signed-off-by: “ to commit message (check other commits for example).

YanVugenfirer:
rerun tests

YanVugenfirer (approved):

YanVugenfirer:
Thanks a lot!

yach-yf:
Thank you very much.
I'm looking forward to get signed driver, to turn off the test mode of my VPN server.

PR #693, opened 2021-12-20
FWCfg64: add 32-bit Windows support

viktor-prutyanov:
Make FWCfg64 driver produce VMCoreInfo record with Complete Memory Dump header for 32-bit Windows guest.
Producing 32-bit Windows dump requires support on QEMU side.

How to test:
1. Install driver into 32-bit Windows
2. Run `(qemu) dump-guest-memory -w memory.dmp`
3. Open dump in WinDbg and check `lm n t`, `k`, etc. works


YanVugenfirer (changes requested):
Please also update solution in the root directory.

YanVugenfirer (approved):

Issue #735, opened 2022-02-22
vioscsi IO sizes too large, causing BSOD on boot


JonKohler:
Using the virtio-latest drivers results in BSODs on our platform.

There are three key issues described below: 
1. vioscsi calculation for MaximumTransferLength results in IO's that are oddly sized, which tickles a fault in our storage code
2. vioscsi handling of max_sectors does not take into account the storage platform's preferred max IO size
3. vioscsi handling of MaximumPhysicalBreaks results in a virtio spec violation, which tickles a fault in our vhost-user-scsi handling code.

**Issue 1: MaximumTransferLength off-by-one** 
The driver code currently has an off-by-one logic error in the way that MaximumTransferLength is calculated. Note: this bug report isn't litigating the use of +1 page, but rather strictly talking about max IO size.
```
    if(adaptExt->dump_mode) {
        ConfigInfo->NumberOfPhysicalBreaks  = SCSI_MINIMUM_PHYSICAL_BREAKS;
    } else {
        adaptExt->max_physical_breaks = PHYS_SEGMENTS;
        if (adaptExt->indirect) {                              <<< we advertise indirection in our vhost-user-scsi backend, so we hit this path
            adaptExt->max_physical_breaks = MAX_PHYS_SEGMENTS; <<< this is equal to 512
            VioScsiReadRegistry(DeviceExtension);
        }
        ConfigInfo->NumberOfPhysicalBreaks = adaptExt->max_physical_breaks + 1; <<< Assuming we don't override the value in VioScsiReadRegistry(), this is equal to 513
...
    ConfigInfo->MaximumTransferLength = ConfigInfo->NumberOfPhysicalBreaks * PAGE_SIZE; <<< 513 * 4096 = 2MB + 4K max IO size
```

This is a problem for us because our storage code has a maximum enforced single IO size of 2MB. This historically has never been a problem for us with any other guest, as we indicate preferred transfer size through max_sectors. More on that in "Issue 2" below.

In debugging this, I could see that this was tickling the fault on the storage side, but to get some more context on *exactly* what the guest was issuing, I added an assertion in our vhost-user-scsi handling application. The simplified version of the logs look like this: 
```
DBG Indirect descriptor table at 0x77abda5a4558 size 4 len 64
DBG DESC[0] (addr=0x8025A2474, len=51, flags=0x1, next=1)
DBG DESC[1] (addr=0x8025A24A7, len=108, flags=0x3, next=2)
[...]

Issue #617, opened 2021-08-03
[vioscsi] ConfigInfo->NumberOfPhysicalBreaks should be dependent on adaptExt->scsi_config.seg_max advertised by qemu


albertlav:
As per Microsoft doc NumberOfPhysicalBreaks must be set according to HBA's scatter/gather capacity:

https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/srb/ns-srb-_port_configuration_information

> Specifies the maximum number of breaks between address ranges that a data buffer can have if the HBA supports scatter/gather. In other words, the number of scatter/gather lists minus one. By default, the value of this member is SP_UNINITIALIZED_VALUE, which indicates the HBA can support an unlimited number of physical discontiguities. If the port driver sets a value for this member, the miniport driver can adjust the value lower but no higher. If this member is SP_UNINITIALIZED_VALUE, the miniport driver must reset this member according to the HBA's scatter/gather capacity, with zero representing no scatter/gather support.
> 

If my understanding is correct HBA scatter/gather capacity advertised by QEMU in https://github.com/qemu/qemu/blob/master/hw/scsi/virtio-scsi.c#L741
as `scsiconf->seg_max` and this value visible to guest as `adaptExt->scsi_config.seg_max`  via `GetScsiConfig()`

However vioscsi currently set NumberOfPhysicalBreaks value based on registry or hardcoded value of 512

Probably NumberOfPhysicalBreaks  should depend on seg_max.

tomty89:
In reality, from what I tested / experienced, NumberOfPhysicalBreaks cannot exceed `max_hw_sectors` of the backend drive. (The adjustable "soft limit" `max_sectors` did not seem to matter though.)

Say `/sys/block/$node/queue/max_hw_sectors*_kb*` is `512` (which is quite a common case). Then the `NumberOfPhysicalBreaks` (which is essentially a number of *pages*) cannot exceed `512 * 1024 = 524288` bytes, which it cannot be larger than `524288 / 4096 = 128` (in the typical 4KiB page size case). (Due to the "+1" requirement, whatever it really is, apparently the current hard-coded value `65` is maximum we can / would want to use for such drive. If for example / some reason the drive has `max_hw_sectors_kb` with a lower value like `256`, NOPB will need to be less than `64`, in which case the "optimal value" appears to be `33`.)

[...]

PR #736, opened 2022-02-22
vioscsi: fix MaximumTransferLength off-by-one and max_sectors handling

JonKohler:
vioscsi: fix MaximumTransferLength off-by-one and max_sectors handling

This change fixes MaximumTransferLength and max_sectors handling as
described in #498 and #735.

- Changed initialization of seg_max to SCSI_MINIMUM_PHYSICAL_BREAKS. This
  gets immediately overwritten anyways; however, this seems like the most
  sane value to initialize this to, rather than MAX_PHYS_SEGMENTS.
- Squashed adaptExt->dump_mode branching into one set of if/then, which allows
  for adaptExt->indirect to be set immediately before it is evaluated.
- Moved adaptExt->indirect configuration to GetScsiConfig.
- Readded initialization of ConfigInfo->MaximumTransferLength to
  SP_UNINITIALIZED_VALUE. This immediately gets overwritten; however, it
  makes it clear what the starting point is (unlimited) so that we can
  intentionally *decrease* the value *down* from there.
- Initialized ConfigInfo->NumberOfPhysicalBreaks to
  SCSI_MINIMUM_PHYSICAL_BREAKS (i.e. 16), so we have a starting point to
  intentionally *increase* the value *up* from there.
- Calculate ConfigInfo->MaximumTransferLength using
  adaptExt->max_physical_breaks instead of ConfigInfo->NumberOfPhysicalBreaks,
  which fixes an off-by-one error that causes IO to slightly be larger than
  a platform might before (e.g. 2MB+4K max size vs 2MB IO max size when using
  512 segments).
- Respect max_sectors, such that max_sectors will have the final say in the
  maximum IO size. Note: If the user intentionally restricted the max IO size
  using the PhysicalBreaks registry key, it may be smaller than max_sectors.
- Changed name of VioScsiReadRegistry to reflect what its true purpose is,
  which is to read PhysicalBreaks key.

Fixes: #498 vioscsi (generic/block): does not work with xhci hosts that accept only read/write of 512 blocks
Fixes: #617 ConfigInfo->NumberOfPhysicalBreaks should be dependent on adaptExt->scsi_config.seg_max advertised by qemu
Fixes: #735 vioscsi IO sizes too large, causing BSOD on boot

Signed-off-by: Jon Kohler <jon@nutanix.com>

### DIFF
--- a/vioscsi/vioscsi.c
+++ b/vioscsi/vioscsi.c
@@ -283,7 +283,7 @@ USHORT CopyBufferToAnsiString(void* _pDest, const void* _pSrc, const char delimi
     return _length;
 }
 
-BOOLEAN VioScsiReadRegistry(
+BOOLEAN VioScsiReadRegistryPhysicalBreaks(
     IN PVOID DeviceExtension
 )
 {
@@ -462,36 +462,44 @@ ENTER_FN();
         return SP_RETURN_NOT_FOUND;
     }
 
-    /* Set num_queues and seg_max to some sane values, to keep "Static Driver Verification" happy */
+    /* Initialize the following variables to temporary values to keep
+     * Static Driver Verification happy. These values will be immediately
+     * reconfigured within GetScsiConfig below, which will retrieve the runtime
+     * values advertised by the underlying device.
+     */
     adaptExt->scsi_config.num_queues = 1;
-    adaptExt->scsi_config.seg_max = MAX_PHYS_SEGMENTS + 1;
+    adaptExt->scsi_config.seg_max    = SCSI_MINIMUM_PHYSICAL_BREAKS;
+    adaptExt->indirect               = FALSE;
     GetScsiConfig(DeviceExtension);
     SetGuestFeatures(DeviceExtension);
 
-    if(!adaptExt->dump_mode) {
-        adaptExt->indirect = CHECKBIT(adaptExt->features, VIRTIO_RING_F_INDIRECT_DESC);
-    }
-
-    ConfigInfo->NumberOfBuses               = 1;//(UCHAR)adaptExt->num_queues;
+    ConfigInfo->NumberOfBuses               = 1;
     ConfigInfo->MaximumNumberOfTargets      = min((UCHAR)adaptExt->scsi_config.max_target, 255/*SCSI_MAXIMUM_TARGETS_PER_BUS*/);
     ConfigInfo->MaximumNumberOfLogicalUnits = min((UCHAR)adaptExt->scsi_config.max_lun, SCSI_MAXIMUM_LUNS_PER_TARGET);
+    ConfigInfo->MaximumTransferLength       = SP_UNINITIALIZED_VALUE;       // Unlimited
+    ConfigInfo->NumberOfPhysicalBreaks      = SCSI_MINIMUM_PHYSICAL_BREAKS; // 16
 
-    if(adaptExt->dump_mode) {
-        ConfigInfo->NumberOfPhysicalBreaks  = SCSI_MINIMUM_PHYSICAL_BREAKS;
-    } else {
-        adaptExt->max_physical_breaks = PHYS_SEGMENTS;
-        if (adaptExt->indirect) {
-            adaptExt->max_physical_breaks = MAX_PHYS_SEGMENTS;
-            VioScsiReadRegistry(DeviceExtension);
+    if (!adaptExt->dump_mode) {
+        adaptExt->max_physical_breaks = adaptExt->indirect ? MAX_PHYS_SEGMENTS : PHYS_SEGMENTS;
+
+        /* Allow user to override max_physical_breaks via reg key
+         * [HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\vioscsi\Parameters\Device]
+         * "PhysicalBreaks"={dword value here}
+         */
+        VioScsiReadRegistryPhysicalBreaks(DeviceExtension);
+
+        if (adaptExt->scsi_config.max_sectors > 0 &&
+            adaptExt->scsi_config.max_sectors != 0xFFFF &&
+            adaptExt->max_physical_breaks * PAGE_SIZE > adaptExt->scsi_config.max_sectors * SECTOR_SIZE) {
+
+            adaptExt->max_physical_breaks = adaptExt->scsi_config.max_sectors * SECTOR_SIZE / PAGE_SIZE;
         }
         ConfigInfo->NumberOfPhysicalBreaks = adaptExt->max_physical_breaks + 1;
-        if (adaptExt->scsi_config.max_sectors > 0 && adaptExt->scsi_config.max_sectors != 0xFFFF) {
-            ConfigInfo->NumberOfPhysicalBreaks = min (ConfigInfo->NumberOfPhysicalBreaks, adaptExt->scsi_config.max_sectors);
-            adaptExt->max_physical_breaks = ConfigInfo->NumberOfPhysicalBreaks - 1;
-        }
+        ConfigInfo->MaximumTransferLength  = adaptExt->max_physical_breaks * PAGE_SIZE;
     }
+
     RhelDbgPrint(TRACE_LEVEL_INFORMATION, " NumberOfPhysicalBreaks %d\n", ConfigInfo->NumberOfPhysicalBreaks);
-    ConfigInfo->MaximumTransferLength = ConfigInfo->NumberOfPhysicalBreaks * PAGE_SIZE;
+    RhelDbgPrint(TRACE_LEVEL_INFORMATION, " MaximumTransferLength %d\n", ConfigInfo->MaximumTransferLength);
 
     num_cpus = KeQueryActiveProcessorCountEx(ALL_PROCESSOR_GROUPS);
     max_cpus = KeQueryMaximumProcessorCountEx(ALL_PROCESSOR_GROUPS);
--- a/vioscsi/helper.c
+++ b/vioscsi/helper.c
@@ -233,6 +233,7 @@ GetScsiConfig(
 ENTER_FN();
 
     adaptExt->features = virtio_get_features(&adaptExt->vdev);
+    adaptExt->indirect = CHECKBIT(adaptExt->features, VIRTIO_RING_F_INDIRECT_DESC);
 
     virtio_get_config(&adaptExt->vdev, FIELD_OFFSET(VirtIOSCSIConfig, seg_max),
                       &adaptExt->scsi_config.seg_max, sizeof(adaptExt->scsi_config.seg_max));


JonKohler:
cc @vrozenfe @sb-ntnx @rob-scheepens

also cc @tomty89 - this should address your xhci issue on 498

also cc @albertlav - this addresses the seg_max questions on 617

vioscsi/vioscsi.c
```diff
@@ -438,60 +438,189 @@ VioScsiFindAdapter(
...
+         * sure we didn't accidentally set the transfer size higher than the
+         * max_sector restricts, as some platforms may act strange (or even
+         * error/fail) if their max_sectors limits are violated.
+         */
+        if (adaptExt->scsi_config.max_sectors > 0 &&
+            adaptExt->scsi_config.max_sectors != 0xFFFF &&
+            ConfigInfo->MaximumTransferLength > (adaptExt->scsi_config.max_sectors * SECTOR_SIZE)) {
+
+            ConfigInfo->MaximumTransferLength  = adaptExt->scsi_config.max_sectors * SECTOR_SIZE;
+            adaptExt->max_physical_breaks      = ConfigInfo->MaximumTransferLength / PAGE_SIZE;
```
tomty89, line 617: I think it would be less clumsy to replace this whole block with:
```
int mbp_ms_clamp = adaptExt->scsi_config.max_sectors * SECTOR_SIZE / PAGE_SIZE;
if (mbp_ms_clamp > 0)
    adaptExt->max_physical_breaks = min(mbp_ms_clamp, adaptExt->max_physical_breaks);
```
and put it after the registry reading but before NOPB and MTL setting.

The `!= 0xffff` check is unnecessary and bad anyway.

vioscsi/vioscsi.c
```diff
@@ -438,60 +438,189 @@ VioScsiFindAdapter(
...
+         * "PhysicalBreaks"={dword value here}
+         *
+         * TODO: Update this logic to incorporate indirect descriptor queue
+         * size limits if/when it is fully ratified into the virtio spec.
+         * This will allow the guest and host to negotiate on this specific
+         * limit. Reference:
+         *   https://github.com/oasis-tcs/virtio-spec/issues/122
+         */
+        adaptExt->indirect = CHECKBIT(adaptExt->features, VIRTIO_RING_F_INDIRECT_DESC);
         if (adaptExt->indirect) {
```
tomty89, line 585: Btw, does any real reason even exist for this check? I mean, the fact that we have two default values disgusts me. For one it's not like `32` ("`PHY_SEGMENTS`") is always safe enough.

For example, for devices with max_(hw_)sector being 240 (USB 2.0 drive), supposedly it can only deal with NOPB being 16+1.

Besides, it would look like that we are suggesting indirect_desc should be set to off when we bump into problems, but the real problem might be *just* the fact that NOPB is too high (i.e. we need to set a clamp).

Let alone putting the registry reading inside a condition.

vioscsi/vioscsi.c
```diff
@@ -438,60 +438,189 @@ VioScsiFindAdapter(
...
 
-    /* Set num_queues and seg_max to some sane values, to keep "Static Driver Verification" happy */
+    /* Initialize num_queues and seg_max to temporary values to keep
+     * Static Driver Verification happy. These values will be immediately
+     * reconfigured within GetScsiConfig below, which will retrieve the runtime
+     * values advertised by the underlying device.
+     */
     adaptExt->scsi_config.num_queues = 1;
-    adaptExt->scsi_config.seg_max = MAX_PHYS_SEGMENTS + 1;
+    adaptExt->scsi_config.seg_max    = SCSI_MINIMUM_PHYSICAL_BREAKS;
```
tomty89, line 483: Perhaps `SCSI_MAXIMUM_PHYSICAL_BREAKS` would make this silly line looks more rational?

tomty89:
For the record, I have no idea whether MTL matters at all when it comes to the max_(hw_)sectors-related problem. I only know that NOPB does. So whether MTL should or should not remain as `SP_UNINITIALIZED_VALUE` when `!adaptExt->dump_mode` is unknown to me.

All I know is that, this kind of code makes no sense to people when it doesn't come with a comment.

tomty89:
> It also seems to me that perhaps the default value(s) and clamps should instead be in bytes / about MTL, as we are practically deriving NOPB from MTL.

This also leads us back to the most important question that no one seems to care: for NOPB should be X + 1, what exactly is the X we are talking about? An arbitrary number of pages derived from a specific MTL? or, an arbitrary power of 2?

Without a concrete definition for X, there's no way to tell what exactly we should do.

Take the aforementioned max_(hw_)sectors 240 as an example, what NOPB should be would be 30+1 and 16+1 respectively for the two possibilities I mentioned.

It's strange that when there are people who knows to +1 but they seemingly have no idea what it is being + to, and it's pathetic whenever this question is raised, it's treated as a challenge to the +1 itself.

JonKohler:
Notes from the virtio-win community meeting this morning with respect to this PR:
1. AI - Jon - Simplify PR and comments
2. AI - Jon - post performance before/after numbers to make sure there are no regressions

JonKohler:
> > It also seems to me that perhaps the default value(s) and clamps should instead be in bytes / about MTL, as we are practically deriving NOPB from MTL.
> 
> This also leads us back to the most important question that no one seems to care: for NOPB should be X + 1, what exactly is the X we are talking about? An arbitrary number of pages derived from a specific MTL? or, an arbitrary power of 2?
> 
> Without a concrete definition for X, there's no way to tell what exactly we should do.
> 
> Take the aforementioned max_(hw_)sectors 240 as an example, what NOPB should be would be 30+1 and 16+1* respectively for the two possibilities I mentioned.
> 
> It's strange that people who know to +1 don't seem to know what that is being + to, and it's pathetic whenever this question is raised, it's treated as a challenge to the +1 itself.
> 
> *: _Assuming_ setting an explicit MTL does _not_ eliminate the need that NOPB (with +1 counted) cannot exceed max_(hw_)sectors equivalence.

In short, yes, I agree. The only thing I really care about here is MTL, and NOPB is just a means to the end for that. and as long as the underlying device handler can handle whatever the value is, its fine

Thanks for the comments. I'll go through them as I clean up this PR a bit

vioscsi/vioscsi.c
JonKohler: Keep in mind my comments are from an the point of view of an outside contributor, but IMHO yes there is a point for this check, as the mere presence of adapter supported indirect descriptors means that we dont have to be as careful about filling up the actual virtio ring itself. That seems to be the original intention of the code

As for USB drive limits -> take this comment with a grain of salt, but do we care? I'd imagine most of the stuff that users would use vioscsi on is not a USB drive, but some sort of more capable storage device :) 

Anyhow, you bring up a good point, let me see if I can clean this up a bit more

JonKohler:
@tomty89 - WRT the +1 business here

I did some additional digging on this +1 and found the answer in the windows API documentation:

https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/srb/ns-srb-_port_configuration_information

> NumberOfPhysicalBreaks
> Specifies the maximum number of breaks between address ranges that a data buffer can have if the HBA supports scatter/gather. In other words, **the number of scatter/gather lists minus one**. By default, the value of this member is SP_UNINITIALIZED_VALUE, which indicates the HBA can support an unlimited number of physical discontiguities. If the port driver sets a value for this member, the miniport driver can adjust the value lower but no higher. If this member is SP_UNINITIALIZED_VALUE, the miniport driver must reset this member according to the HBA's scatter/gather capacity, with zero representing no scatter/gather support.

So, the driver is doing exactly this "minus 1" noted in the API documentation and another way to phrase adaptExt->max_physical_breaks would be "number of scatter gather lists"

You can see it being used in VioScsiBuildIo() in such a fashion, albeit also having the +1 there.

Note that in VioScsiBuildIo() we dont have access to ConfigInfo, so we're mirroring that value to match up with what would come back from the storport API call.

https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nf-storport-storportgetscattergatherlist
https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/ns-storport-_stor_scatter_gather_list

vioscsi/vioscsi.c
JonKohler: yea, we need that because vhost-scsi.c in qemu sets max_sectors to 0xFFFF as default, which is 65k, so yes it looks clunky, but it serves a purpose.

JonKohler:
@vrozenfe - I've simplified the PR as discussed. I'll set up some tests in the lab now and see about performance

tomty89:
For one I don't think the math or interpretation is right there. Take the following as an example:
```
-|-
```
I think what the quoted text merely means that, NOPB is not referring to the number of `-` ("the number of scatter/gather lists") but the number of `|` ("the number of scatter/gather lists minus one").

But you know what, I don't even want to bother digging deep into it, because it is not leading anyone anywhere anyway. The true question remains unanswered, what should the X be?

As you mentioned, `adaptExt->max_physical_breaks + 1`, which is equal to NOPB but not NOPB - 1, is used in `VioScsiBuildIo()`. Unless X is e.g. required to be a power of 2 (*which no ones seems to want to confirm or deny*), +1 doesn't mean anything, because e.g 30 is 29 + 1, 29 is 28 + 1. Before MTL is made (again) to be set recently, `adaptExt->max_physical_breaks` is not used solely for anything. (And no one seem to actually know whether MTL is related to NOPB; it has been changed back and forth as specified or unspecified as well.)

And whether you put +1 (a cosmetic one, or practical one) in the code is not the concern here. The concern is: what's the maximum optimal NOPB I should use given the limit a max_(hw_)sectors value poses.

> Take the aforementioned max_(hw_)sectors 240 as an example, what NOPB should be would be 30+1 and 16+1* respectively for the two possibilities I mentioned.

Should I set the max_sector clamp to `128`, `240`, or `232`? NOPB will be `16+1`, `30+1` and `29+1` respectively.

JonKohler:
The maximum optimal NOPB would be simply whatever you want the MTL divided by PAGE_SIZE + 1

So, if you had max_sectors which would be (240*512)/1024 = 120KB max IO size, that would be ((240*512)/4096)+1 = 31

that should work in the code in this PR.

For kicks, I'll see if I can restrict my setup to max_sectors = 240 and see what happens

vioscsi/vioscsi.c
```diff
@@ -462,36 +462,44 @@ ENTER_FN();
...
+         * [HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\vioscsi\Parameters\Device]
+         * "PhysicalBreaks"={dword value here}
+         */
+        VioScsiReadRegistryPhysicalBreaks(DeviceExtension);
+
+        if (adaptExt->scsi_config.max_sectors > 0 &&
+            adaptExt->scsi_config.max_sectors != 0xFFFF &&
+            adaptExt->max_physical_breaks * PAGE_SIZE > adaptExt->scsi_config.max_sectors * SECTOR_SIZE) {
+
+            adaptExt->max_physical_breaks = adaptExt->scsi_config.max_sectors * SECTOR_SIZE / PAGE_SIZE;
```
tomty89, line 495: Again, instead of all the clumsy conditions, it should really just be analogous to what's in `VioScsiReadRegistryPhysicalBreaks()`. `!= 0xFFFF` is just bad, and `> 0` is more or less wrong.

Is there any "verification" that would be "unhappy" if a local variable is introduced?

JonKohler: Why is 0xFFFF bad? That's 65535 IIRC, which is hardcoded in qemu unless the user overrides it with their own value. 

I dont think >0 is wrong, though we could probably drop the > 0 as you'd probably have to try hard to advertise max_sectors as 0, I don't even know if you could do that :) though I have a feeling that _**someone**_ out there will end up filing a bug report saying that they indeed set max_sectors to zero for some reason

RE Variable - I dont think the static analyzer cares about such a variable

Anyhow, you are right, we could use the clamp snippet you provided before, it would get us to the same place.

Here's a logical walkthrough for the sake of doing it long hand
```
int mbp_ms_clamp = adaptExt->scsi_config.max_sectors * SECTOR_SIZE / PAGE_SIZE; 
    // mbp_ms_clamp = 65535 * 512 / 4096 = 8191.875
adaptExt->max_physical_breaks = min(max(SCSI_MINIMUM_PHYSICAL_BREAKS, mbp_ms_clamp), adaptExt->max_physical_breaks);
    // max_physical_breaks min(max(16, 8191.875), 512) = 512 -> good
```

Then there is our case where max_sectors = 2048 (1MB)
```
int mbp_ms_clamp = adaptExt->scsi_config.max_sectors * SECTOR_SIZE / PAGE_SIZE; 
    // mbp_ms_clamp = 2048 * 512 / 4096 = 256
adaptExt->max_physical_breaks = min(max(SCSI_MINIMUM_PHYSICAL_BREAKS, mbp_ms_clamp), adaptExt->max_physical_breaks);
    // max_physical_breaks min(max(16, 256), 512) = 256 -> good
```

vrozenfe:
@JonKohler 
Looks good to me.
Thank you a lot.
Vadim.

vioscsi/vioscsi.c
tomty89: What I mean is, instead of 0, we should have the same lower limit for this as the registry clamp (i.e. SCSI_MINIMUM_PHYSICAL_BREAKS)

As for `!= 0xffff`, for one the maximum qemu can take could change; regardless of whether that would actually happen, so `!=` (instead of `>`) is bad. What I mean by "wrong" is, similar to the case of the lower limit, it's disgusting to have a clamp somewhere (registry), and to have a override somewhere else (here). I think the least we can do is to have some consistency. (For the record, there's nothing wrong with an override itself; that's purely a matter of preference.)

vrozenfe:
@tomty89 

Thank you for your feedback.
Please open a separate PR on top of this one, if you know how to make it better.

Best,
Vadim.

PR #779, opened 2022-06-15
pvpanic: Support pvpanic PCI driver

annie-li:
As we know, QEMU supports PCI pvpanic. Also, Linux has implemented
its pvpanic PCI driver. However, the pvpanic PCI driver has been
missing on Windows side.

This patch implements supporting pvpanic PCI driver for Windows.
Both ISA and PCI pvpanic drivers share the same set of driver
source code, as well as same driver binary(.sys). But they have
separate INF file since the ID of ISA and PCI device are different.

QEMU options for pvpanic PCI device
   -device pvpanic-pci

Since the pvpanic driver is being deployed for just sending the
panic event to notify the crash of the guest. It is not necessary
to support multiple pvpanic devices in one guest. Since ISA and
PCI device share the same driver, one of the driver will fail
to be loaded if both ISA and PCI device exist in the same guest.
However, this doesn't affect the crash event notification since
the loaded driver always works as expected.

In case of supporting the co-existence of ISA and PCI pvpanic
devices, two implementations are available.
1. Split the code and build out separate driver binaries
2. Share the same code but split the variables and callback
   functions for ISA and PCI driver
However, these implementations are more complicated. These may be
implemented in future if required.

Signed-off-by: Annie Li <annie.li@oracle.com>
Reviewed-by: Darren Kenny <darren.kenny@oracle.com>

### DIFF
--- a/pvpanic/pvpanic/power.c
+++ b/pvpanic/pvpanic/power.c
@@ -51,6 +51,9 @@ NTSTATUS PVPanicEvtDevicePrepareHardware(IN WDFDEVICE Device,
         Device);
 
     PAGED_CODE();
+    context->MappedPort = FALSE;
+    context->IoBaseAddress = NULL;
+    context->MemBaseAddress = NULL;
 
     for (i = 0; i < WdfCmResourceListGetCount(ResourcesTranslated); ++i)
     {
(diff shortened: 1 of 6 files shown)


kostyanf14:
Hi @annie-li,

Thanks for your contribution. Please check HKC-CI logs. For some reason, this driver failed 4 tests and generate BSOD during testing.

kostyanf14:
Hardware-enforced Stack Protection Compatibility Test failure is expected because this is a manual test. Just ignore this.

pvpanic/pvpanic/power.c
```diff
@@ -81,23 +84,65 @@ NTSTATUS PVPanicEvtDevicePrepareHardware(IN WDFDEVICE Device,
...
+                else
+                    return STATUS_DEVICE_CONFIGURATION_ERROR;
 
                 break;
             }
 
+            case CmResourceTypeMemory:
+            {
+                TraceEvents(TRACE_LEVEL_VERBOSE, DBG_POWER,
+                    "Memory mapped CSR: (%x) Length: (%d)",
```
viktor-prutyanov, line 98:
```suggestion
                    "Memory mapped CSR: (%lx) Length: (%lu)",
```

pvpanic/pvpanic/power.c
```diff
@@ -81,23 +84,65 @@ NTSTATUS PVPanicEvtDevicePrepareHardware(IN WDFDEVICE Device,
...
+#else
+                context->MemBaseAddress = MmMapIoSpace(
+                    desc->u.Memory.Start,
+                    desc->u.Memory.Length,
+                    MmNonCached);
+#endif
+                if (!PvPanicPortOrMemAddress)
+                    PvPanicPortOrMemAddress = (PUCHAR)context->MemBaseAddress;
+                else
+                    return STATUS_DEVICE_CONFIGURATION_ERROR;
```
viktor-prutyanov, line 116: I think some trace message can be printed to detect situation when 2nd PVPanic device is being added over existing one. It would be helpful when user accidentally added PCI and ISA devices together.

viktor-prutyanov (changes requested):

pvpanic/pvpanic/power.c
```diff
@@ -81,23 +84,65 @@ NTSTATUS PVPanicEvtDevicePrepareHardware(IN WDFDEVICE Device,
                     context->IoBaseAddress =
                         (PVOID)(ULONG_PTR)desc->u.Port.Start.QuadPart;
                 }
-                PvPanicPortAddress = (PUCHAR)context->IoBaseAddress;
+                if(!PvPanicPortOrMemAddress)
```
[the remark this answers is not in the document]
annie-li: Good catch.

pvpanic/pvpanic/power.c
annie-li: Will add trace there, thanks.

annie-li:
Thanks for the feedback, Viktor.
I will post a new patch with the fix of WHQL test failures, as well as the code change as suggested.

Issue #793, opened 2022-07-12
Hidden files


Daniel-Trevitz:
Files and folders that start with "." should be flagged as hidden to windows.

Probably should have flag to enable or disable? Default of enabled?

YanVugenfirer:
@Daniel-Trevitz 

I personally prefer not to have too much flags, as 99% of people will not use them :)
But, in case we have. My preference would be enabled (flag "." files as hidden on Windows).

Best regards,
Yan.

Daniel-Trevitz:
@YanVugenfirer, Good with me. I also added a registry flag; do you want me to keep that?

YanVugenfirer:
@Daniel-Trevitz Let's keep it.

scottdmilner:
Is there an update on this? This would be a useful feature to have. This is the default behavior when mounting a Linux samba share: all files starting with `.` are automatically flagged as hidden.

viktor-prutyanov:
> Is there an update on this? This would be a useful feature to have. This is the default behavior when mounting a Linux samba share: all files starting with `.` are automatically flagged as hidden.

What exactly is the benefit here?

scottdmilner:
> What exactly is the benefit here?
The benefit is that the same files are shown/hidden in the Windows filesystem as are shown/hidden in the Linux filesystem. If I look at a filesystem in Nemo or Dolphin (or in a shell with ls), then I will see the same folders/files when looking at that filesystem in the Windows File Manager. It cleans up visual clutter.

YanVugenfirer:
@scottdmilner Currently, we are not planning to implement file hiding. If someone is motivated enough to do it (controlled by the flag, with default off), we will happily review the patches.

scottdmilner:
@YanVugenfirer if I'm reading the discussion correctly, I believe that this PR implemented this functionality? https://github.com/virtio-win/kvm-guest-drivers-windows/pull/797

YanVugenfirer:
@scottdmilner  Thank you very much for finding this discussion!
I tried to recall yesterday where we documented the discussion regarding hidden files and could find out.
So trying to recap what was said and decided: because there is no exact specification on how the hidden files should be treated, we might end with strange corner case bugs.

PR #797, opened 2022-07-14
Hide files/folders that start with "."

Daniel-Trevitz:
This change reads the file & folder names and flags the ".*" files as hidden.

Ext4 (and may other Linux filesystems) there is no metadata to specify a file is hidden.
Instead, it is understood by most applications that files and folders starting with a "." should be hidden.

Adds DWORD the registry setting "HideFilesStartingWithDot" which when set to 0 disables the hiding.
When set to any other value, the functionality is enabled.

Signed-off-by: Daniel Trevitz <daniel.trevitz@wika.com>

### DIFF
--- a/viofs/svc/virtiofs.cpp
+++ b/viofs/svc/virtiofs.cpp
@@ -130,12 +130,14 @@ struct VIRTFS
     UINT32  OwnerUid{ 0 };
     UINT32  OwnerGid{ 0 };
 
+    BOOL HideFilesStartingWithDot{ true };
     BOOL IsRunningAsLocalSystem{ false };
 
     // Maps NodeId to its Nlookup counter.
     std::map<UINT64, UINT64> LookupMap{};
 
-    VIRTFS(ULONG DebugFlags, const std::wstring& MountPoint);
+    VIRTFS(ULONG DebugFlags, const std::wstring& MountPoint,
+           BOOL HideFilesStartingWithDot);
     NTSTATUS Start();
     VOID Stop();
     NTSTATUS SubmitInitRequest();
@@ -144,8 +146,11 @@ struct VIRTFS
     NTSTATUS SubmitDestroyRequest();
 };
 
-VIRTFS::VIRTFS(ULONG DebugFlags, const std::wstring& MountPoint)
-    : DebugFlags{ DebugFlags }, MountPoint{ MountPoint }
+VIRTFS::VIRTFS(ULONG debugFlags, const std::wstring& mountPoint,
+               BOOL hideFilesStartingWithDot)
+    : DebugFlags{ debugFlags }
+    , MountPoint{ mountPoint }
+    , HideFilesStartingWithDot{ hideFilesStartingWithDot }
 {
 }
 
@@ -654,6 +659,23 @@ static VOID UnixTimeToFileTime(uint64_t time, uint32_t nsec,
     FspPosixUnixTimeToFileTime(UnixTime, PFileTime);
 }
 
+static VOID CheckSetFileHidden(VIRTFS *VirtFs, CHAR *FileName, FSP_FSCTL_FILE_INFO *FileInfo)
+{
+    // if len(FileName) == 0 then FileName[0] == '\0'
+    if (VirtFs->HideFilesStartingWithDot && (FileName[0] == '.'))
+    {
+        FileInfo->FileAttributes |= FILE_ATTRIBUTE_HIDDEN;
+    }
+}
+
+static VOID CheckSetFileHidden(VIRTFS *VirtFs, WCHAR *FileName, FSP_FSCTL_FILE_INFO *FileInfo)
+{
+    if (VirtFs->HideFilesStartingWithDot && (FileName[0] == L'.'))
+    {
+        FileInfo->FileAttributes |= FILE_ATTRIBUTE_HIDDEN;
+    }
+}
+
 static VOID SetFileInfo(VIRTFS *VirtFs, struct fuse_entry_out *entry,
     FSP_FSCTL_FILE_INFO *FileInfo)
 {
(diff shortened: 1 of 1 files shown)


Daniel-Trevitz:
Address issue #793.

My development approach was to find every call to SetFileInfo that had a filename.
If that call site had obvious access to a filename, I called one of the two CheckSetFileHidden().

Line 1216 of GetFileInfoInternal does not set the hidden flag.
I wouldn't be surprised if this means some stat operation somewhere fails to specify that a file is hidden, but FWIW, explorer hides the files now.

In a VM without active directory:
```
PS Z:\test> &"C:\Program Files (x86)\winfsp\bin\winfsp-tests-x64.exe" --resilient --external -reparse* -stream* -getfile
info_test -create_fileattr_test -create_readonlydir_test -create_allocation_test -create_notraverse_test -create_backup_
test -create_restore_test -create_share_test -getfileattr_test -getfileinfo_name_test -setfileinfo_test -setsecurity_tes
t -querydir_test
create_test............................ OK 0.16s
create_related_test.................... OK 0.01s
create_sd_test......................... OK 0.03s
create_curdir_test..................... OK 0.01s
create_namelen_test.................... OK 0.02s
delete_test............................ OK 0.03s
delete_access_test..................... OK 0.03s
delete_pending_test.................... OK 0.03s
delete_mmap_test....................... OK 0.01s
delete_standby_test.................... OK 0.16s
delete_ex_test......................... OK 0.01s
rename_test............................ OK 0.16s
rename_backslash_test.................. OK 0.05s
rename_open_test....................... OK 0.02s
rename_caseins_test.................... OK 0.06s
rename_flipflop_test................... OK 0.56s
rename_mmap_test....................... OK 0.00s
rename_standby_test.................... OK 0.30s
rename_ex_test......................... OK 0.00s
getvolinfo_test........................ OK 0.00s
setvolinfo_test........................ OK 0.00s
getsecurity_test....................... OK 0.02s
rdwr_noncached_test.................... OK 0.05s
rdwr_noncached_overlapped_test......... OK 0.06s
rdwr_cached_test....................... OK 0.05s
rdwr_cached_append_test................ OK 0.03s
rdwr_cached_overlapped_test............ OK 0.03s
rdwr_writethru_test.................... OK 0.03s
rdwr_writethru_append_test............. OK 0.02s
rdwr_writethru_overlapped_test......... OK 0.03s
rdwr_mmap_test......................... OK 0.22s
rdwr_mixed_test........................ OK 0.03s
flush_test............................. OK 0.08s
flush_volume_test...................... OK 0.00s
lock_noncached_test.................... OK 0.03s
lock_noncached_overlapped_test......... OK 0.02s
lock_cached_test....................... OK 0.01s
lock_cached_overlapped_test............ OK 0.05s
querydir_expire_cache_test............. OK 0.00s
querydir_buffer_overflow_test.......... OK 0.02s
querydir_namelen_test.................. OK 0.01s
dirnotify_test......................... OK 1.05s
exec_test.............................. OK 0.11s
exec_delete_test....................... OK 1.06s
exec_rename_test....................... OK 1.12s
exec_rename_dir_test................... OK 2.06s
--- COMPLETE ---
```

viktor-prutyanov:
Hi @Daniel-Trevitz, thank you for the PR.

I've tried to find any connections between files that start with dot and hidden files in Windows and found nothing.

For example, in NTFS (to which VirtIO-FS tries to be similar), file `.file.txt` doesn't become hidden automatically (there is no `H` attribute):

image

In general, it seems reasonable to me that programs, but not the filesystem, should decide which files are hidden and which are not.

Daniel-Trevitz:
> I've tried to find any connections between files that start with dot and hidden files in Windows and found nothing.

@viktor-prutyanov, I would not expect you to find such a correlation since this exists in the Linux file manager space.

Linux file managers have a name based hidden file system.
Windows has a filesystem attribute based hidden file system.

Since we are sharing Linux FS to Windows FS, this patch lets Windows know that the host's "hidden" files are such.

> In general, it seems reasonable to me that programs, but not the filesystem, should decide which files are hidden and which are not.
Unfortunately, I don't think MS will honor a pull request to change explorer ;)

viktor-prutyanov:
> Linux file managers have a name based hidden file system. 

As far as I know, Linux doesn't have such a thing as "hidden file". Only r/w/x for u/g/o +  setuid/setgid/sticky/directory.

> Unfortunately, I don't think MS will honor a pull request to change explorer ;)

But there are other filemanagers which may have different file display rules.

Filemanager is on a different abstraction layer other then filesystem and it is filemanager's responsibility to show/hide files based on any logic it wants. Filesystem is only responsible to show files and metadata properly, it shouldn't add anything of its own. In this sense, the VirtIO-FS lacks support of "hidden" attribute. It should be so that the program sets/unsets the attribute and VirtIO-FS shows actual state of the attribute.

Daniel-Trevitz:
@viktor-prutyanov, As a user VirtIO-FS incorrectly shows hidden Linux files on Windows.

I do suggest, that if you would prefer this feature be turned off by default, I am amenable to that change.

We are not in a position to change how Windows applications are, or were, developed.

You stated that the "." part of the name is not an attribute. You are correct. 
I contend, that to provide our users with the most intuitive experience, we should treat it as one.
This will make all existing and future Windows applications know that a Linux file was meant to be a hidden file, at the cost of a byte comparison.

I do recognize that this behavior is not perfect. If I mark a file as hidden in windows, I do not trigger a file rename.
This is identical to how VirtualBox behaves, hiding ".*" files by default, and not renaming files when I mark them as hidden (vbox-sf just ignores that request from Explorer), but will hide the file if I rename it to start with a ".".

YanVugenfirer:
> Since we are sharing Linux FS to Windows FS, this patch lets Windows know that the host's "hidden" files are such.

I think this is not entirely correct. The sharing works both ways, and for some "guest correctness" might be more important.

In any case, I put some additional thought into this issue and looked at what Samba does (it is not hiding the files). 
I think we are risking introducing very hard to debug bugs and behaviours by implementing "nice to have" feature. 
So for now I am leaning not to merging the code at all. If there is a high demand for the feature, we can try to merge it disabled by default.

Best regards,
Yan.

PR #799, opened 2022-07-14
Fix login permissions

Daniel-Trevitz:
The optimization to detect if we are the LocalSystem user fails
  when the service is started before a user logs in (AD connnected or not).

Simply remove the check, and try both methods to get a token.
As long as UpdateLocalUidGid is not a hot path, this should be fine.

Signed-off-by: Daniel Trevitz <daniel.trevitz@wika.com>

### DIFF
--- a/viofs/svc/virtiofs.cpp
+++ b/viofs/svc/virtiofs.cpp
@@ -130,8 +130,6 @@ struct VIRTFS
     UINT32  OwnerUid{ 0 };
     UINT32  OwnerGid{ 0 };
 
-    BOOL IsRunningAsLocalSystem{ false };
-
     // Maps NodeId to its Nlookup counter.
     std::map<UINT64, UINT64> LookupMap{};
 
@@ -539,22 +537,15 @@ static VOID UpdateLocalUidGid(VIRTFS *VirtFs, DWORD SessionId)
     TOKEN_USER *Uinfo{ NULL };
     TOKEN_PRIMARY_GROUP *Ginfo{ NULL };
 
-    if (VirtFs->IsRunningAsLocalSystem)
+    if (!WTSQueryUserToken(SessionId, &Token))
     {
-        if (!WTSQueryUserToken(SessionId, &Token))
+        if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &Token))
         {
             VirtFs->LocalUid = 0;
             VirtFs->LocalGid = 0;
-            DBG("Failed to open a process token as LocalSystem; Error=%d", GetLastError());
+            DBG("Failed to open a process token; Error=%d", GetLastError());
             return;
         }
-    }
-    else if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &Token))
-    {
-        VirtFs->LocalUid = 0;
-        VirtFs->LocalGid = 0;
-        DBG("Failed to open a process token; Error=%d", GetLastError());
-        return;
     }
 
     Result = FspToolGetTokenInfo(Token, TokenUser, (PVOID *)&Uinfo);
@@ -2590,7 +2581,6 @@ NTSTATUS VIRTFS::SubmitDestroyRequest()
 
 NTSTATUS VIRTFS::Start()
 {
-    HANDLE Token{ NULL };
     NTSTATUS Status;
     DWORD SessionId;
     FILETIME FileTime;
@@ -2605,18 +2595,6 @@ NTSTATUS VIRTFS::Start()
     SessionId = WTSGetActiveConsoleSessionId();
     if (SessionId != 0xFFFFFFFF)
     {
-        // Will fail if we are not the LocalSystem user.
-        IsRunningAsLocalSystem = WTSQueryUserToken(SessionId, &Token);
-
-        if (IsRunningAsLocalSystem)
-        {
-            CloseHandle(Token);
-        }
-        else
-        {
-            DBG("The service %s was not run as the LocalSytem user", FS_SERVICE_NAME);
-        }
-
         UpdateLocalUidGid(this, SessionId);
     }
     else


YanVugenfirer:
What will happen if another service will try to use virtio-fs mount before any user logged in?

Daniel-Trevitz:
All tests pass on windows (with no AD)

```
 &"C:\Program Files (x86)\winfsp\bin\winfsp-tests-x64.exe" --resilient --external -reparse* -stream* -getfile
info_test -create_fileattr_test -create_readonlydir_test -create_allocation_test -create_notraverse_test -create_backup_
test -create_restore_test -create_share_test -getfileattr_test -getfileinfo_name_test -setfileinfo_test -setsecurity_tes
t -querydir_test
create_test............................ OK 0.01s
create_related_test.................... OK 0.03s
create_sd_test......................... OK 0.02s
create_curdir_test..................... OK 0.00s
create_namelen_test.................... OK 0.00s
delete_test............................ OK 0.00s
delete_access_test..................... OK 0.00s
delete_pending_test.................... OK 0.00s
delete_mmap_test....................... OK 0.00s
delete_standby_test.................... OK 0.06s
delete_ex_test......................... OK 0.00s
rename_test............................ OK 0.00s
rename_backslash_test.................. OK 0.01s
rename_open_test....................... OK 0.00s
rename_caseins_test.................... OK 0.00s
rename_flipflop_test................... OK 0.03s
rename_mmap_test....................... OK 0.00s
rename_standby_test.................... OK 0.12s
rename_ex_test......................... OK 0.00s
getvolinfo_test........................ OK 0.00s
setvolinfo_test........................ OK 0.00s
getsecurity_test....................... OK 0.00s
rdwr_noncached_test.................... OK 0.02s
rdwr_noncached_overlapped_test......... OK 0.02s
rdwr_cached_test....................... OK 0.01s
rdwr_cached_append_test................ OK 0.00s
rdwr_cached_overlapped_test............ OK 0.02s
rdwr_writethru_test.................... OK 0.00s
rdwr_writethru_append_test............. OK 0.02s
rdwr_writethru_overlapped_test......... OK 0.00s
rdwr_mmap_test......................... OK 0.09s
rdwr_mixed_test........................ OK 0.02s
flush_test............................. OK 0.02s
flush_volume_test...................... OK 0.00s
lock_noncached_test.................... OK 0.01s
lock_noncached_overlapped_test......... OK 0.00s
lock_cached_test....................... OK 0.02s
lock_cached_overlapped_test............ OK 0.01s
querydir_expire_cache_test............. OK 0.00s
querydir_buffer_overflow_test.......... OK 0.00s
querydir_namelen_test.................. OK 0.00s
dirnotify_test......................... OK 1.02s
exec_test.............................. OK 0.05s
exec_delete_test....................... OK 1.03s
exec_rename_test....................... OK 1.03s
exec_rename_dir_test................... OK 2.03s
--- COMPLETE ---
```

Daniel-Trevitz:
> What will happen if another service will try to use virtio-fs mount before any user logged in?

I have no clue how to test that...
What I think happens is that before a user logs in the service will fail to run with WTSQueryUserToken (because there is no user) and fall back to OpenProcessToken. 
But, since we are the LocalSystem user, the OpenProcessToken will set the uid/gid to 18 since the SID is S-1-5-18

I am not sure how to test it, but I believe it means only LocalSystem has write access in this case.

Daniel-Trevitz:
The important part FWIW is that on login of an AD user or a normal user SERVICE_CONTROL_SESSIONCHANGE triggers UpdateLocalUidGid and gives the correct permissions for the user.

YanVugenfirer:
> I also don't know how to test multiple seats to an RDP server / virtual desktop situation.

I think using "Switch user" regular Windows desktop feature is similar. In this case the session of the initial user will still be active, and you will have interactive session with a second user.

Daniel-Trevitz:
> I think using "Switch user" regular Windows desktop feature is similar. In this case the session of the initial user will still be active, and you will have interactive session with a second user.

Tested. It only works when you actually log in to the other user. When you switch users that not seem to trigger the login event (UpdateLocalUidGid is not in the logs for each switch, only each login)

Daniel-Trevitz:
Oh boy... lots of options: https://docs.microsoft.com/en-us/windows/win32/termserv/wm-wtssession-change

@YanVugenfirer, Is there any reason to filter the UpdateUID by the session change type? Obviously, if something is known to spam this, it could be an issue.

Daniel-Trevitz:
> DBG("The service %s was not run as the LocalSytem user", FS_SERVICE_NAME);

Short follow-up; I added this debug statement specifically so we would know when someone puts in the next bug report for AD that they didn't run it as LocalSystem which means that AD deployments can't work (and then instruct them to run it as a service).

Daniel-Trevitz:
> > DBG("The service %s was not run as the LocalSytem user", FS_SERVICE_NAME);
> 
> Short follow-up; I added this debug statement specifically so we would know when someone puts in the next bug report for AD that they didn't run it as LocalSystem which means that AD deployments can't work (and then instruct them to run it as a service).
> 
> That is gone now, which is probably suboptimal.

I remembered more about why I removed this debug statement. I removed it because this change makes it "just work" regardless of how it is run. The need for the debug doesn't exist with this change.

YanVugenfirer:
After some offline discussion: we have an architectural "miss" regarding how to handle mounts and permissions that needs additional thinking. 
Putting PR on hold until we think about initial ideas what's the correct way to mount for different users and how to handle multiple mounts.

Best regards,
Yan.

Daniel-Trevitz:
> what's the correct way to mount for different users and how to handle multiple mounts.
@YanVugenfirer, What kind of topologies are you trying to make this work with? Are you trying to handle multiple Linux users 1:1 mapping to multiple windows users? How does multiple mounts tie into this?

As far as this being a miss, my only contention is that I don't know what it breaks that already wor, or how it is an architectural change to switch the token source (which we need to get the correct UID/GID regardless the solution)

YanVugenfirer:
After a discussion on virtio-fs community meeting, we decided to take different approach:
https://github.com/virtio-win/kvm-guest-drivers-windows/pull/803

PR #804, opened 2022-07-31
RHBZ#2042247: VirtIO-FS: add '-t' option to mount FS with specific tag

viktor-prutyanov:
When running FS in console mode the user can set '-t' option to specify a tag.
Only FS with the specified tag will be considered to mount.
This option allows user to have multiple VirtIO-FS instances in the system.
Example:
```
start "" virtiofs.exe -d -1 -D C:\viofs_debug0.txt -m Y: -t mount0
start "" virtiofs.exe -d -1 -D C:\viofs_debug1.txt -m Z: -t mount1
```


viofs/svc/virtiofs.cpp
```diff
@@ -143,10 +144,23 @@ struct VIRTFS
 };
 
 VIRTFS::VIRTFS(ULONG DebugFlags, const std::wstring& MountPoint)
-    : DebugFlags{ DebugFlags }, MountPoint{ MountPoint }
+    : DebugFlags{ DebugFlags }, MountPoint{ MountPoint }, Tag{ Tag }
 {
 }
 
+template <typename EF>
+class scope_exit
```
kostyanf14, line 152: Why you manually add this class? C++ has it https://en.cppreference.com/w/cpp/experimental/scope_exit, any reason to don't use library?

viofs/svc/virtiofs.cpp
```diff
@@ -487,6 +483,35 @@ static DWORD FindDeviceInterface(PHANDLE Device)
...
+
+    for (DWORD MemberIndex = 0; ; MemberIndex++)
+    {
+        DWORD Error = FindDeviceInterface(Device, MemberIndex);
+        if (Error != ERROR_SUCCESS)
+        {
+            return Error;
+        }
+
+        WCHAR VolumeName[37];
```
kostyanf14, line 501: This array size looks strange. Is this size of something? Can you use some MS define with this value? Or just add comment.

viofs/svc/virtiofs.cpp
viktor-prutyanov: Because it is an experimental feature, and it seems to be unavailable in MSVC. When it become available custom variant can be easily replaced to variant from library.

viofs/svc/virtiofs.cpp
viktor-prutyanov: This is 36 + 1 where 36 is tag size from VirtIO-FS spec. We have such a definition for the driver. I should be moved to common header file and used here.

YanVugenfirer (approved):

kostyanf14 (approved):

Issue #809, opened 2022-08-12
[vioscsi] saving BSOD crash dumps fails


main--:
It seems that a similar issue was fixed back in #321, but I'm now observing the same problem again. Generating a minidump usually works, but other types of crash dump just fail/abort at 0% progress, producing a (very) corrupted dump file.

I can reproduce the problem with the latest build (221), but also with earlier builds, so not sure when it was introduced.

This is on Windows 10 21H2 x64 with qemu 7.0.0.

vrozenfe:
@main-- 
Can you please post the qemu command line, and let me know how did you trigger the crash ?

Thanks,
Vadim.

main--:
Sure!

Here is my command line:

```
cp /usr/share/edk2-ovmf/x64/OVMF_VARS.fd /tmp && qemu-system-x86_64 -enable-kvm -machine q35 -cpu host
-rtc base=localtime -nodefaults -net none -display gtk -vga qxl 
-drive if=pflash,format=raw,unit=0,readonly=on,file=/usr/share/edk2-ovmf/x64/OVMF_CODE.fd 
-drive if=pflash,format=raw,unit=1,file=/tmp/OVMF_VARS.fd -device virtio-scsi-pci,id=scsi 
-m 8G -smp cores=22 -device qemu-xhci -device usb-kbd -device usb-tablet 
-drive file=windows.raw,id=disk,format=raw,if=none,cache=none,aio=io_uring,discard=unmap 
-device scsi-hd,drive=disk
```

I triggered the crash using the Sysinternals NotMyFault tool.

vrozenfe:
@main-- 
Confirmed.
Reproduced on seabios setup with nmi interrupt as well.
I will ask QE to file a new bug in RH bugzilla and try to fix it
in the following days.

Thanks a lot for reporting the problem. 
Vadim.

vrozenfe:
https://bugzilla.redhat.com/show_bug.cgi?id=2118630

PR #811, opened 2022-08-17
[vioscsi] fix for BZ#2118630 saving BSOD crash dumps fails

vrozenfe:
Reported-by : main@ehvag.de
https://github.com/virtio-win/kvm-guest-drivers-windows/issues/809

Signed-off-by: Vadim Rozenfeld <vrozenfe@redhat.com>

### DIFF
--- a/vioscsi/vioscsi.c
+++ b/vioscsi/vioscsi.c
@@ -470,14 +470,15 @@ ENTER_FN();
     adaptExt->scsi_config.num_queues = 1;
     adaptExt->scsi_config.seg_max    = SCSI_MINIMUM_PHYSICAL_BREAKS;
     adaptExt->indirect               = FALSE;
+    adaptExt->max_physical_breaks    = SCSI_MINIMUM_PHYSICAL_BREAKS;
     GetScsiConfig(DeviceExtension);
     SetGuestFeatures(DeviceExtension);
 
     ConfigInfo->NumberOfBuses               = 1;
     ConfigInfo->MaximumNumberOfTargets      = min((UCHAR)adaptExt->scsi_config.max_target, 255/*SCSI_MAXIMUM_TARGETS_PER_BUS*/);
     ConfigInfo->MaximumNumberOfLogicalUnits = min((UCHAR)adaptExt->scsi_config.max_lun, SCSI_MAXIMUM_LUNS_PER_TARGET);
-    ConfigInfo->MaximumTransferLength       = SP_UNINITIALIZED_VALUE;       // Unlimited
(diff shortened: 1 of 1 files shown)


JonKohler:
Broken by my commit on #736 - sorry for the breakage, I didn't test BSOD saves in our setup during development, my mistake entirely!

@YanVugenfirer - is this something that the HCK has coverage for? Curious how it slipped through, or if this is a test case that needs to get added?

FYI @sb-ntnx - this would be a problem for us when we pick up the commit I did around max IO sizes, regardless of IO size/breaks configuration

sb-ntnx:
@JonKohler, I am aware of the issue as I am monitoring this repo. I checked it in my morning and it could be reproduced with the build that I had ready in January, so I doubt it is your commit. 
HCK/HLK is a Microsoft test suite and it does not check whether a dump can or cannot be generated. Also, if I am not mistaken, some of HCK/HLK tests require a device removal, so in my setup system drive is also using SATA bus, so even if we hit a BSOD during any HCK/HLK test (it is not uncommon), the dump will be collected properly.

vrozenfe:
@JonKohler @sb-ntnx 
This bug is fully mine, introduced with the following commit a54882c4fc4bfec8d5abb583ca84a3778c6e8bd4
Even though that the problem is reproducible only in crash dump stack and only on 64-bit platforms, it is
very sad  that the problem was not discovered and fixed earlier. We will bring the QE attention to check not
only presence MEMORY.DMP file but also the crash dump consistency bu running dumpchk utility or by
opening a generated crash dump file with WinDbg.

Vadim.

sb-ntnx:
I have also verified the fix on W2022 and it works fine, no issues found in HLK2022 either. Thanks Vadim.

vrozenfe:
@sb-ntnx 
Thank you for your feedback, Sergey.
Cheers,
Vadim.

YanVugenfirer (approved):

YanVugenfirer:
> @JonKohler @sb-ntnx This bug is fully mine, introduced with the following commit [a54882c](https://github.com/virtio-win/kvm-guest-drivers-windows/commit/a54882c4fc4bfec8d5abb583ca84a3778c6e8bd4) Even though that the problem is reproducible only in crash dump stack and only on 64-bit platforms, it is very sad that the problem was not discovered and fixed earlier. We will bring the QE attention to check not only presence MEMORY.DMP file but also the crash dump consistency bu running dumpchk utility or by opening a generated crash dump file with WinDbg.
> 
> Vadim.

We are also exploring to add additional tests to CI other than HCK. The first candidate will be virtio-fs driver, where HCK is not texting file system functionality (because the driver itself is just simple WDF driver). So it could be that in the future we will add more tests. The problem of course is the amount of time it takes to execute CI pipeline.

Best regards,
Yan.

PR #832, opened 2022-09-28
VirtIO-FS: initial case insensitive support

viktor-prutyanov:
When `LOOKUP` or `UNLINK`/`RMDIR` or `RENAME`/`RENAME2` request is required, the service tries to send it with exact filename (as usual). If the request result is not success, the service sends `OPEN` and `READDIR` requests to parent directory and tries to find target name among parent directory entries by case-insensitive comparisons and perform `LOOKUP`/`UNLINK`/`RMDIR`/`RENAME`/`RENAME2` on it and then `RELEASE` parent directory.

Add `-i` option and registry key `CaseInsensitive`.

Example:
```
$ ls Z:\
Hello.txt  Привет.txt
$ cat HELLO.txt
Hello
$ cat ПРИВЕТ.txt
Привет
```

### DIFF
--- a/viofs/svc/virtiofs.cpp
+++ b/viofs/svc/virtiofs.cpp
@@ -137,9 +137,15 @@ struct VIRTFS
     // Maps NodeId to its Nlookup counter.
     std::map<UINT64, UINT64> LookupMap{};
 
-    VIRTFS(ULONG DebugFlags, bool CaseInsensitive, const std::wstring& MountPoint);
+    VIRTFS(ULONG DebugFlags, bool CaseInsensitive, const std::wstring& MountPoint,
+        const std::wstring& Tag) : DebugFlags{ DebugFlags },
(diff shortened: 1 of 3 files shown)


YanVugenfirer (commented):
@viktor-prutyanov Just in case, did you run this change with Driver Verifier enabled?

viofs/svc/virtiofs.cpp
```diff
@@ -117,6 +117,7 @@ struct VIRTFS
     // Used to handle device remove notification.
     VIRTFS_NOTIFICATION DevHandleNotification{};
 
+    bool CaseInsensitive{ false };
```
YanVugenfirer, line 120: @viktor-prutyanov I am wondering if it should be the opposite. Case insensitivity as default.

viktor-prutyanov:
> @viktor-prutyanov Just in case, did you run this change with Driver Verifier enabled?

No, because all of the changes are in user-space service

YanVugenfirer (approved):

YanVugenfirer (approved):

PR #855, opened 2022-11-10
NETKVM: fix packet loss in special case

80886:
If a poor designed TCP server application doesn't receive data from windows socket function driver afd.sys as quickly as possible, then afd.sys will hold recv buffers temporaryly, and don't call NdisReturnNetBufferLists until afd return-buffer timer expired(at most 1 second).At this time, if there are many packets send in by tcp clients, there may not be enough receive buffer descriptors in the vring to receive packets, and finally packet loss.

Notes:
1. If someone found this patch could help you to avoid packet loss, the most probably root cause is the application is not well-designed;
2. An another way which can avoid packet loss under the situation described above is set DoNotHoldNICBuffers to 1(REG_DWORD) under registry path HKEY_LOCAL_MACHINE\SYSTEM\CCS\Services\AFD\Parameters. But microsoft had never mentioned this registry (even hadn't document the behavior of HoldNICBuffers), so the registry param maybe not working in the future version of windows.
3. According to my analysis, the pusrpose of HoldNICBuffers behavior in afd.sys is mainly to decrease memory copy times when deliver incomming tcp packets to the buffer of recv/wsarecv systemcall. But sometimes it may lead packet loss.

Signed-off-by: junjiehua <junjiehua@tencent.com>
Reviewed-by: yongduan <yongduan@tencent.com>

### DIFF
--- a/NetKVM/Common/ParaNdis_Common.cpp
+++ b/NetKVM/Common/ParaNdis_Common.cpp
@@ -1588,13 +1588,14 @@ UpdateReceiveFailStatistics(PPARANDIS_ADAPTER pContext, UINT nCoalescedSegmentsC
     pContext->Statistics.ifInDiscards += nCoalescedSegmentsCount;
 }
 
-static void ProcessReceiveQueue(PARANDIS_ADAPTER *pContext,
+static BOOLEAN ProcessReceiveQueue(PARANDIS_ADAPTER *pContext,
                                 PULONG pnPacketsToIndicateLeft,
                                 PPARANDIS_RECEIVE_QUEUE pTargetReceiveQueue,
                                 PNET_BUFFER_LIST *indicate,
                                 PNET_BUFFER_LIST *indicateTail,
                                 ULONG *nIndicate)
 {
+    BOOLEAN ret = FALSE;
     pRxNetDescriptor pBufferDescriptor;
 
     while( (*pnPacketsToIndicateLeft > 0) &&
@@ -1622,6 +1623,10 @@ static void ProcessReceiveQueue(PARANDIS_ADAPTER *pContext,
                 NET_BUFFER_LIST_NEXT_NBL(*indicateTail) = NULL;
                 (*pnPacketsToIndicateLeft)--;
                 (*nIndicate)++;
+
+                //According to the feedback from customer who originally faced packet loss issue, 
+                //64 is enough and no side effect observed.
+                ret = (!ret) ? (pBufferDescriptor->Queue->GetAvailRecvBufferCount() < 64) : ret;
             }
             else
             {
@@ -1635,6 +1640,8 @@ static void ProcessReceiveQueue(PARANDIS_ADAPTER *pContext,
             pBufferDescriptor->Queue->ReuseReceiveBuffer(pBufferDescriptor);
         }
     }
+
+    return  ret;
 }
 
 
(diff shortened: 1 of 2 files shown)


0x80386:
@YanVugenfirer  Seems the HCK-CI test hash on NDISTest 6.5 - [2 Machine] - GlitchFreeDevice [00:05:00] ?

kostyanf14:
Hi @0x80386,

I will rerun all failed tests. Looks like CI issue.

YanVugenfirer:
@0x80386 Thank you very much for the patch. I am on a long weekend, and I will review it on Monday.
I will also leave one comment inline.

Also, as this is a change on a data path - did you have a chance to run performance tests? If yes, please share the results.

Thanks,
Yan.

NetKVM/Common/ParaNdis_Common.cpp
```diff
@@ -1622,6 +1623,10 @@ static void ProcessReceiveQueue(PARANDIS_ADAPTER *pContext,
                 NET_BUFFER_LIST_NEXT_NBL(*indicateTail) = NULL;
                 (*pnPacketsToIndicateLeft)--;
                 (*nIndicate)++;
+
+                //According to the feedback from customer who originally faced packet loss issue, 
+                //64 is enough and no side effect observed.
+                ret = (!ret) ? (pBufferDescriptor->Queue->GetAvailRecvBufferCount() < 64) : ret;
```
YanVugenfirer, line 1629: I suggest to have a registry key to control this number, in this case is some setups will need different configuration it can be controlled without changing the code. If you think that this is not needed, in anyway please use a constant with meaningful name instead of magic number.

Thanks,
Yan.

0x80386: Thanks , I will add a registry value so that users can control this value. And the perf results will be uploaded here soon.

YanVugenfirer: Thanks a lot!

kostyanf14:
Hi @0x80386,

I rerun several tests and it fails again. Also, there are several minidumps where the NetKVM driver crashed. 
The stack trace from one of them:
```
fffffb82`939e29c8 fffff803`100b795e     : 00000000`0000007c 00000000`00000025 00000000`00000021 ffffd30b`a0e36fa8 : nt!KeBugCheckEx
fffffb82`939e29d0 fffff803`10189237     : 00000000`00000000 fffff803`0f058450 ffffd30b`a0e36ee0 00000000`00000120 : NDIS!ndisBugCheckEx+0x1e
fffffb82`939e2a10 fffff803`10189115     : 00000000`001209bb ffffd30b`a0e36ee0 fffff803`10148b60 00000000`00000000 : NDIS!ndisReportTimeoutWaitingForExternalDriver+0xc3
fffffb82`939e2a50 fffff803`10188f98     : ffffd30b`a0e36ee0 ffffd30b`a0e36fa8 ffffd30b`a0e36f60 00000000`00000018 : NDIS!ndisFindSomeoneToBlame+0x125
fffffb82`939e2ac0 fffff803`101893c9     : 00000000`00000010 00000000`00010246 fffffb82`939e2b18 00000000`00000018 : NDIS!NdisWatchdogState::ReportTimeout+0x68
fffffb82`939e2af0 fffff803`0e86d8e1     : ffffd30b`9ace9ca0 fffff803`0f343600 ffffd30b`00000000 fffff803`00000000 : NDIS!ndisWatchdogTimeoutWorkerRoutine+0x9
fffffb82`939e2b20 fffff803`0e8ee345     : ffffd30b`9f28f080 00000000`00000001 ffffd30b`9f28f080 00000000`00000000 : nt!ExpWorkerThread+0x161
fffffb82`939e2d30 fffff803`0ea1fef8     : ffffa980`163e8180 ffffd30b`9f28f080 fffff803`0e8ee2f0 00000000`00000000 : nt!PspSystemThreadStartup+0x55
fffffb82`939e2d80 00000000`00000000     : fffffb82`939e3000 fffffb82`939dd000 00000000`00000000 00000000`00000000 : nt!KiStartSystemThread+0x28
```

Please look for these bugs.

80886:
I'll close this PR first and will submit a new one.

PR #873, opened 2023-01-13
RHBZ#2142718: VirtIO-FS: fix 'nobody' shared root owner handling

viktor-prutyanov:
The host daemon may be running inside the Linux user namespace, so unmapped UIDs and GIDs are shown as `nobody` (65534). Files/directories must not be created with this ID, so the service should try to preserve a valid ID number if shared directory owner turns out to be `nobody`.


viktor-prutyanov:
If `nobody` UID/GID is changed on the host through `/proc/sys/fs/overflowuid,overflowgid`, the guest user can set these values in registry.

YanVugenfirer (approved):
Please add documentation in WiKi

PR #882, opened 2023-02-19
NetKVM: Do not report vendor-specific OIDs (WHQL)

MartinDrab:
This step is necessary to pass the "NDISTest 6.0 - [1 Machine] - 1c_WMICoverage" which gets into trouble when a driver reports vendor-specific OIDs the test cannot find names for. If you wish NetKVM to report those OIDs again, define the NETKVM_REPORT_SPECIFIC_OIDS symbol.

NetKVM still handles these specific OIDs although it does not report them.

This issue happened in the past and was fixed the same way at that time [1]. We owe current appearance to a small time window between expiration of filtered test cases and adoption of their new versions. This commit should fix the problem for good.

[1]: https://bugzilla.redhat.com/show_bug.cgi?format=multiple&id=1439085

### DIFF
--- a/NetKVM/wlh/ParaNdis6_Oid.cpp
+++ b/NetKVM/wlh/ParaNdis6_Oid.cpp
@@ -273,8 +273,8 @@ static NDIS_OID SupportedOids[] =
         OID_GEN_XMIT_OK,
         OID_GEN_RCV_OK,
         OID_GEN_VLAN_ID,
-#if NDIS_SUPPORT_NDIS61
-// disable WMI custom command on 2008 due to non-filtered NDIS test failure
+#ifdef NETKVM_REPORT_SPECIFIC_OIDS
+// disable WMI custom command on all OSes due to non-filtered NDIS test failure
         OID_GEN_SUPPORTED_GUIDS,
         OID_VENDOR_1,
         OID_VENDOR_2,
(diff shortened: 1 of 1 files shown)


YanVugenfirer:
Hi @MartinDrab ,

I suggest inversing the logic of disabling the report. The filter was already updated, so we would like to keep the reports enabled in the "default" build for debugging purposes in the field.

Best regards,
Yan.

kostyanf14:
I run `NDISTest 6.5 - [2 Machine] - GlitchFreeDevice` it failed two times

ybendito:
1. NDISTest 6.0 - [1 Machine] - 1c_WMICoverage has a filter that make this test to pass.
2. Such custom OIDs are completely legitimate and widely used by many network drivers, so this feature by irself should not present any problem.
3. If some of vendors want to disable usage of custom OIDs - we can make it conditionally, but currently I'd prefer the default to be "enabled".
We can discuss it in dedicated issue if needed, so I suggest to close the PR.

YanVugenfirer:
Agree with @ybendito. The main point if we change is proposed, the current behaviour should be default.
Closing for now.

ybendito:
@MartinDrab What I can suggest. Please try to comment in the table only OID_VENDOR_X (1-4), not OID_GEN_SUPPORTED_GUIDS.
If this helps to pass the HCK without filter - then submit the PR (or reopen existing) to remove only them.
I've checked that after that the WMI works, but I can't tell about HCK as our HCK passes.

PR #887, opened 2023-03-03
fix bug for query vq alloc of leagacy device

gaojunhao:
for leagacy virtio device, vio_leagacy_query_vq_alloc will check queue num or queue pfn, so need to add ! before ioread32(vdev, vdev->addr + VIRTIO_PCI_QUEUE_PFN).

### DIFF
--- a/VirtIO/VirtIOPCILegacy.c
+++ b/VirtIO/VirtIOPCILegacy.c
@@ -160,7 +160,7 @@ static NTSTATUS vio_legacy_query_vq_alloc(VirtIODevice *vdev,
 
     /* Check if queue is either not available or already active. */
     num = ioread16(vdev, vdev->addr + VIRTIO_PCI_QUEUE_NUM);
-    if (!num || ioread32(vdev, vdev->addr + VIRTIO_PCI_QUEUE_PFN)) {
+    if (!num || !ioread32(vdev, vdev->addr + VIRTIO_PCI_QUEUE_PFN)) {
         return STATUS_NOT_FOUND;
     }
 


VirtIO/VirtIOPCILegacy.c
```diff
@@ -160,7 +160,7 @@ static NTSTATUS vio_legacy_query_vq_alloc(VirtIODevice *vdev,
 
     /* Check if queue is either not available or already active. */
     num = ioread16(vdev, vdev->addr + VIRTIO_PCI_QUEUE_NUM);
-    if (!num || ioread32(vdev, vdev->addr + VIRTIO_PCI_QUEUE_PFN)) {
+    if (!num || !ioread32(vdev, vdev->addr + VIRTIO_PCI_QUEUE_PFN)) {
```
YanVugenfirer, line 163: Can you please describe how you hit the problem that requires the change?
Current code checks how many queues are active and if for some reason the queue address is already set.

YanVugenfirer: Also if the change is needed, also the comment for the code should change

YanVugenfirer (changes requested):

VirtIO/VirtIOPCILegacy.c
gaojunhao: > Can you please describe how you hit the problem that requires the change? Current code checks how many queues are active and if for some reason the queue address is already set.

I allocate a leagacy vsock device to windows, and windows start, then I install the vsock driver of virtio-win, the driver will has error and fail to install. The error is This device cannot start (Code 10) STATUS_DEVICE_POWER_FAILURE.
Then I use logman and wpa to record and analyze the trace, found the error trace: VIOSockRxVqInit virtio_query_queue_allocation(VIOSOCK_VQ_RX) failed. Then I go into the function virtio_query_queue_allocation->vio_legacy_query_vq_alloc, then find ioread32(vdev, vdev->addr + VIRTIO_PCI_QUEUE_PFN) return not 0 thats cause vio_legacy_query_vq_alloc return STATUS_NOT_FOUND, so it will affect the vsock driver install.

YanVugenfirer:
Let me check beginning of next week if this is not a problem on a host side.

gaojunhao:
> Let me check beginning of next week if this is not a problem on a host side.

Thanks for checking.  Firstly I think my change has problem, it will affect setup_vq, so it could not be modified. Secondly, vsock driver could not be install correctly, may change if (!NT_SUCCESS(status)) in VIOSockRxVqInit function to if (uRingSize == 0) like the deal in vioscsi ?

gaojunhao:
> Let me check beginning of next week if this is not a problem on a host side.

Thanks for checking, I think my modification has big problem, it will affect setup_vq, so it could not be modified. Then virtio_query_queue_allocation in VIOSockRxVqInit will return STATUS_NOT_FOUND, so it could not use virtio_query_queue_allocation interface from my view because it will check VIRTIO_PCI_QUEUE_PFN.

gaojunhao:
> Let me check beginning of next week if this is not a problem on a host side.

sorry for the chage of this patch, this patch shoud be change

viktor-prutyanov:
Hi @gaojunhao 

Could you please describe how do you run QEMU with legacy vsock device?
By the way, why don't you use modern vsock device?

YanVugenfirer:
Closing because of inactivity.

If still needed, please re-open.

PR #890, opened 2023-03-05
Fix Windows 10 targets

kostyanf14:


kostyanf14:
Close after discussion

Issue #867, opened 2022-12-14
viosock: calling select() in Windows guest results in BSOD


tuxxi:
**Describe the bug**
Running a server in Windows guest, connecting from a Linux host.
Calling `select()` on an accepted socket handle causes the VM to crash.

**To Reproduce**
Steps to reproduce the behaviour:
* Bind() address and Listen() for connections in VM guest
* Accept connection
* Call select() on socket
* Observe VM crash

It doesn't matter what parameters we pass in to select(), or whether the socket is in blocking or non-blocking mode.

Here is a minimal reproducible example. Note that error handling has been omitted for brevity
```c
#include <fcntl.h>
#include <stdio.h>
#include <stdlib.h>

#include <ws2spi.h>
#include <winsock2.h>

#include "vio_sockets.h"

#pragma comment(lib, "ws2_32.lib")

#define PORT 591

int main() {
    WSADATA wsaData = { 0 };
    WSAStartup(MAKEWORD(2, 2), &wsaData);
    ADDRESS_FAMILY af = ViosockGetAF();
    SOCKET sock = socket(af, SOCK_STREAM, 0);

    struct sockaddr_vm addr;
    struct sockaddr_vm peer_addr;

    memset(&addr, 0, sizeof(struct sockaddr_vm));
    addr.svm_family = af;
    addr.svm_port = PORT;
    addr.svm_cid = VMADDR_CID_ANY;

    bind(sock, (const struct sockaddr *)&addr, sizeof(struct sockaddr_vm));
    listen(sock, /*max connections=*/1);

    int peer_addr_size = sizeof(struct sockaddr_vm);
    int peer_fd = accept(sock, (struct sockaddr *)&peer_addr, &peer_addr_size);

    printf("Got client, CID: %d, port:%d\n", peer_addr.svm_cid, peer_addr.svm_port);
    
    FD_SET readSet;
    TIMEVAL timeout = {1, 0};

    size_t msg_len = 0;
    char buf[64];

    while(msg_len == 0) {
        FD_ZERO(&readSet);
        FD_SET(peer_fd, &readSet);

        select(0, &readSet, NULL, NULL, &timeout); // <----- fail here
        printf("select OK\n");
        if (FD_ISSET(peer_fd, &readSet)) {
            size_t recvd = recv(peer_fd, &buf[0], sizeof(buf), 0);
            if (recvd == WSAEWOULDBLOCK) {
                continue;
            }
[...]

PR #892, opened 2023-03-09
Fix calling select in guest results in BSOD

xuehuihui:


xuehuihui:
Solve this problem:
https://github.com/virtio-win/kvm-guest-drivers-windows/issues/867

YanVugenfirer (approved):

PR #896, opened 2023-03-13
RHBZ#2177673: fwcfg64: don't fail if unable to work with vmcoreinfo

viktor-prutyanov:
Previously, fwcfg64 driver was failing if fw_cfg device was present, but `etc/vmcoreinfo` entry wasn't. The driver shouldn't fail when device is OK. Anyway, `dump-guest-memory -w` will not work without vmcoreinfo.

### DIFF
--- a/fwcfg64/power.c
+++ b/fwcfg64/power.c
@@ -49,6 +49,9 @@ NTSTATUS FwCfgEvtDeviceD0Entry(IN WDFDEVICE Device,
 
     PAGED_CODE();
 
+    if (!ctx->kdbg)
+        return STATUS_SUCCESS;
+
     note->n_namesz = sizeof(VMCI_ELF_NOTE_NAME);
     note->n_descsz = DUMP_HDR_SIZE;
     note->n_type = 0;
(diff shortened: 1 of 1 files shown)


fwcfg64/power.c
```diff
@@ -120,13 +123,20 @@ NTSTATUS FwCfgEvtDevicePrepareHardware(IN WDFDEVICE Device,
                 "VMCoreInfo index is 0x%x", ctx->index);
     if (!ctx->index)
     {
-        return STATUS_BAD_DATA;
+        TraceEvents(TRACE_LEVEL_VERBOSE, DBG_PNP,
```
YanVugenfirer, line 126: Will the line 110 not fail?    
if (FWCfgCheckSig(ctx->ioBase) ||
        FWCfgCheckFeatures(ctx->ioBase, FW_CFG_VERSION_DMA) ||
        FWCfgCheckDma(ctx->ioBase))
    {
        return STATUS_BAD_DATA;
    }

viktor-prutyanov: @YanVugenfirer  It will not fail if fw_cfg is peresent

viktor-prutyanov: On modern QEMU it can't fail

viktor-prutyanov: Some very old versions can't do "DMA"

viktor-prutyanov:
rerun tests

YanVugenfirer (approved):

kostyanf14:
rerun tests

PR #916, opened 2023-05-24
viorng: Remove coinstaller from the INF

MartinDrab:
Direct registry manipulation is used instead of the co-installer. ¨This means, the crypto provider registration gets into effect after a reboot. Also, current registry changes might not be compatible with future versions of Windows, however, there seems to be no other way of installing the CNG provider through an INF file when rules for Universal Driver INF files are in effect.


YanVugenfirer:
@MartinDrab Thanks a lot!

On what OSes did you test those patches?

MartinDrab:
> @MartinDrab Thanks a lot!
> 
> On what OSes did you test those patches?

On Windows 10. But I can possibly add Windows 11 to my test environment in few days.

kostyanf14:
> On Windows 10. But I can possibly add Windows 11 to my test environment in few days.

@MartinDrab,
If you have any idea or solution on how to test the UMD part automatically, please ping me and I will extend our CI.

kostyanf14:
viofs test was run due to a root solution update. All other drivers tests are canceled.

MartinDrab:
> > On Windows 10. But I can possibly add Windows 11 to my test environment in few days.
> 
> @MartinDrab, If you have any idea or solution on how to test the UMD part automatically, please ping me and I will extend our CI.

Well, there is the `viorngtest.exe` application (it is part of the MSVS solution). It just reads random data from the CNG cryptographic provider installed by the viorng INF (driver). If it is successful, it returns exit code `0` and prints something like this:
```
D:\path\>viorngtest 1024
0: Microsoft Primitive Provider
1: QEMU VirtIO RNG Provider
Using it
Done OK
```
Otherwise it returns a nonzero exit code and prints an error message.

So, it can be tested like this:
- install the driver via the INF file,
- reboot,
- run `viorngtest 1024`.

The value of `1024` is size of the random data buffer to read from the driver, you may specify basically any value you like.

To add the VirtIo RNG device into the VM, `-device virtio-rng-pci` needs to be passed to QEMU (or something mentioned in [1]). If this is not possible, you can still test presence of the `QEMU VirtIO RNG Provider` provider since `viorngtest` will list it before failing.

[1]: https://wiki.qemu.org/Features/VirtIORNG

MartinDrab:
I also tested on Windows 8 and Windows 8.1 and the provider installs successfully as well. Can do Windows 11 soon.

kostyanf14:
Tested UDM on Server 2019, 2022 - works as expected.

viorng/viorng/viorng.inf
```diff
@@ -90,21 +89,16 @@ HKR,Parameters,DmaRemappingCompatible,0x00010001,1
...
 
-[VirtRng_Wdf_CoInstaller_AddReg]
-HKR,,CoInstallers32,0x00010000,"viorngci.dll,VirtRngCoInstaller", \
-    "WdfCoInstaller$KMDFCOINSTALLERVERSION$.dll,WdfCoInstaller"
 
-[VirtRng_Wdf_CoInstaller_CopyFiles]
-WdfCoInstaller$KMDFCOINSTALLERVERSION$.dll
-viorngci.dll
+; -------------------------------------
+; WDF
```
kostyanf14, line 100: Why did you remove the WDF coinstaller `WdfCoInstaller$KMDFCOINSTALLERVERSION$.dll` ?

MartinDrab: Since all work is done in the `AddReg` directive and there is no coinstaller, I saw no reason to include this in the INF file.

MartinDrab: Ah, I see, this is for the correct WDF version to install. I will bring it back.

kostyanf14:
`WdfCoInstaller$KMDFCOINSTALLERVERSION$.dll` is still removed in this patch 
@MartinDrab

MartinDrab:
> `WdfCoInstaller$KMDFCOINSTALLERVERSION$.dll` is still removed in this patch @MartinDrab

Yes I know. There is a bit of a trouble with Windows version before Windows 10. Universal INFs prohibit the `Coinstallers` section even if no custom coinstallers are specified (and the goal is only to ship the right WDF version with the driver). Of course, this is not an issue on Windows 10 (and later) where that is not necessary.

I suppose it should be OK to use the right WDF version on each of the supported version of Windows (i.e. the version that is always present there). I just need to check whether such versions are defined and the driver actually depends on them.

kostyanf14:
@MartinDrab 

>Coinstallers section even if no custom coinstallers are specified 


This is expected. We have this section in ALL drivers for ALL versions. For example fwcfg64 driver https://www.dropbox.com/sh/ieza0co6k1jez9b/AAD4TRbr8_pKJgJk0CPucs_Da?dl=0&file_subpath=%2Ffwcfg.inf&lst=&preview=fwcfg64-Win2022x64.zip

kostyanf14:
FWCfg INF https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/fwcfg64/fwcfg.inf#L80

kostyanf14:
rerun tests

PR #923, opened 2023-06-07
vioscsi: always kick VQ when buffers exhausted

JonKohler:
When VIRTIO_RING_F_EVENT_IDX is not enabled, we should kick the VQ even if VIRTQ_USED_F_NO_NOTIFY is set when all buffers are exhausted.

This will ensure that we always kick when we're full, not only *after* we are full. Do this by keying off of vq->num_unused <=1 in virtqueue_kick_prepare_split.

On the vioscsi side, after successfully adding a buffer to VQ, query virtqueue_kick_prepare under lock to ensure that we can get an accurate read of vq->num_unused.

nit: cleanup virtqueue_kick_prepare_split to only deal with event_suppression_related variables if that is enabled.


VirtIO/VirtIORing.c
```diff
@@ -349,16 +351,14 @@ static bool virtqueue_kick_prepare_split(struct virtqueue *_vq)
...
     KeMemoryBarrier();
 
-    wrap_around = (vq->num_added_since_kick >= (1 << 16));
-
-    old = (u16)(vq->master_vring_avail.idx - vq->num_added_since_kick);
-    new = vq->master_vring_avail.idx;
-    vq->num_added_since_kick = 0;
-
     if (_vq->vdev->event_suppression_enabled) {
+        wrap_around = (vq->num_added_since_kick >= (1 << 16));
```
JonKohler, line 355: While I'm at it, is this code right? Isn't 1<<16 ~65k? That seems absurdly large, wouldn't we want to be talking about wrapping around the size of the ring? or am I missing something?

sb-ntnx (changes requested):

vioscsi/helper.c
```diff
@@ -104,12 +105,11 @@ ENTER_FN_SRB();
...
         InsertTailList(&element->srb_list, &srbExt->list_entry);
         element->srb_cnt++;
+        notify = virtqueue_kick_prepare(adaptExt->vq[QueueNumber]);
     }
     VioScsiVQUnlock(DeviceExtension, MessageID, &LockHandle, FALSE);
-    if ( res >= 0){
-        if (virtqueue_kick_prepare(adaptExt->vq[QueueNumber])) {
-            virtqueue_notify(adaptExt->vq[QueueNumber]);
-        }
+    if (res >= 0 && notify) {
```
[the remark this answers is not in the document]
JonKohler: ah right. Morning brain fart, clearly the coffee has not kicked in yet. One second, I'll fix

JonKohler:
For posterity, the linux UAPI definition for `VRING_USED_F_NO_NOTIFY` very clearly states that the guest *will kick* when its out of buffers, so we should be good with this approach

```
/* The Host uses this in used->flags to advise the Guest: don't kick me when
 * you add a buffer.  It's unreliable, so it's simply an optimization.  Guest
 * will still kick if it's out of buffers. */
#define VRING_USED_F_NO_NOTIFY	1
```

vrozenfe:
@JonKohler 

It looks good to me.
Since the changes are not only in viostor  but also in virtio ring code we  need some time to check the impact on other drivers/stacks. 

Best regards,
Vadim.

JonKohler:
Thanks, Vadim. We're doing some internal testing on this as well. Basic functionality seems fine, performance appears to mostly be a wash to what @sb-ntnx was looking at before, which is what I'd expect since we're just optimizing around a corner case

Thinking about this more, maybe we should save the notify call into the error path if the datapath thread(s) have set NO_NOTIFY, as it is possible that when the last buffer is added (num_unused <= 1) that we get a completion immediately afterward, which would make the kick not really relevant. And we are already spec compliant in that we *will* kick when we're out of buffers already. 

Waiting a bit more internal testing before we say more one way or the other.

Worst case, I could peel this change back to just avoiding dealing with the event based variables when event IDX is not enabled, which would be a nice general improvement/cleanup.

YanVugenfirer:
@JonKohler Let me check next week. As far as I remember we already have a code kicking on a full ring in network driver. 
Are you disabling event indexing (VIRTIO_RING_F_EVENT_IDX) in your backend? I am just wondering if we are not dealing here with legacy case that not worth touching.

VirtIO/VirtIORing.c
ybendito: We need to look at it. Similar code exists in Linux kernel virtio_ring.c

ybendito:
@JonKohler Can you please explain a little more why this additional kick make sense at all? If the host sets no_nofity on used, it definitely will not do anything with these kicks (as it is busy with previous ones). Is there a _real_ need to to that except of the note in the virtio_ring.h?

PR #925, opened 2023-06-15
viosock: Fix rare race condition in VIOSockReadDequeueCb

MartinDrab:
This routine is used to match send() and recv() operations made on a given socket. This occurs in these situations:
- after a send() is recorded for the socket,
- when a new recv() request is added to an empty queue of such requests,

The routine may requeue a recv() request (e.g. because it finds not enough send() ones to fully satisfy it), which may trigger it being called recursively again and again. Thus, a mechanism to detect the recursion is implemented through the lInProgress variable. However, this may backfire in some rare cases, for example:
- the routine is called after recording a send() request, it is about to exit since the recv() queue is empty, simultaneously
- a recv() request just got into the queue, however, VIOSockReadDequeueCb does not do the matching since another instance of itself is still running.

This leads to send() and recv() operations wait infinitely for each other.

This branch implements new recursion detection mechanism that does not suffer from this problem. The routine now check whether it is not called recursively by one thread which prevents the recursion but nothing else.

### DIFF
--- a/viosock/sys/Rx.c
+++ b/viosock/sys/Rx.c
@@ -1297,7 +1297,6 @@ VIOSockReadDequeueCb(
     IN PSOCKET_CONTEXT  pSocket
 )
 {
-    static volatile LONG    lInProgress;
     PDEVICE_CONTEXT pContext = GetDeviceContextFromSocket(pSocket);
     WDFREQUEST      ReadRequest = WDF_NO_HANDLE;
     NTSTATUS        status = STATUS_SUCCESS;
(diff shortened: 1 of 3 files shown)


YanVugenfirer:
Hi @MartinDrab , thank you for sending the PR. I am currently travelling for KVMForum and will review the PR next week.

MartinDrab:
> Hi @MartinDrab , thank you for sending the PR. I am currently travelling for KVMForum and will review the PR next week.

No problem, take your time.

MartinDrab:
> I am trying to understand the flow more. The previous solution worked because it implemented the simplified lock inside the function. The problem was that the flag could flip just before the function exited, and then the function would be re-entrant. But now there is no lock protection at all for the new thread ID variable, and VIOSockReadProcessDequeueCb can be called from several places. Shouldn't the setting of thread ID be protected? Also, why is it not enough to rely on pSocket->RxLock?

Changes to the thread ID are protected by `pSocket->RxLock`. Actually, most of `VIOSockReadProcessDequeueCb` is protected by it. Goal of the thread ID here is just to prevent `VIOSockReadProcessDequeueCb` being called (through requeuing the recv() request) recursively by itslef. 

Speaking about `pSocket->RxLock` I think it does protect `VIOSockReadProcessDequeueCb` well enough. Even if it is called concurrently by multiple threads, the worst thing that can happen is that some of them end quickly because of the empty recv() queue.

Thus, my idea was just to prevent recursive calls of `VIOSockReadProcessDequeueCb`, but not prevent this routine being meaningfully called by multiple threads. 

Hope this helps. If not, I will try to elaborate more.

YanVugenfirer (approved):

PR #926, opened 2023-06-15
viosock: Use STATUS_IO_TIMEOUT instead of STATUS_TIMEOUT

MartinDrab:
The problem (or "problem") with `STATUS_TIMEOUT` is that it is classified as a success status code (the `NT_SUCCESS` macro returns `TRUE`). This problem does not appear in user mode since `viosocklib` converts `STATUS_TIMEOUT` to corresponding Win32 error code,

However, kernel mode components/drivers mostly use `NT_SUCCESS` macro to distinguish success from failure, thus, `STATUS_TIMEOUT` can be misleading for them. Also, some uses of the `VIOSockPendedRequestSetEx` in the viosock driver itself improperly handle `STATUS_TIMEOUT`.

This pull request modifies viosock driver to return `STATUS_IO_TIMEOUT` instead. This is an error code. `viosocklib` is also updated to translate this code to the timeout Win32 error.

### DIFF
--- a/viosock/lib/native.c
+++ b/viosock/lib/native.c
@@ -77,6 +77,7 @@ NtReadFile(
 #define STATUS_CONNECTION_RESET          ((NTSTATUS)0xC000020DL)
 #define STATUS_LOCAL_DISCONNECT          ((NTSTATUS)0xC000013BL)
 #define STATUS_HOST_UNREACHABLE          ((NTSTATUS)0xC000023DL)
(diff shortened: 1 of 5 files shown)


YanVugenfirer (approved):

YanVugenfirer:
Hi @MartinDrab, Approved, but I am waiting for CI to run. We have a large queue due to the PRs that involved changes in virtio library.

Issue #614, opened 2021-07-31
Baloon Serv


Svra44:

xiagao:
@Svra44 
Hi, I want to ask the detailed steps to reproduce this issue, in my env I didn't reproduce it.
Host: RHEL8
Guest:Win10-64(10.0.19043)
virtio-win-0.1-204

Steps:
1. Boot up vm with balloon device, machine type is q35.
2. Install balloon driver and start balloon service
blnsvr.exe -i
3. enable polling in qmp:
{"execute":"qmp_capabilities"}
{"execute": "qom-set","arguments": { "path":
"/machine/peripheral/balloon0","property": "guest-stats-polling-interval",
"value": 2 } }
{ "execute": "qom-get","arguments": { "path":
"/machine/peripheral/balloon0","property": "guest-stats" } }
4. Balloon memory via qmp
5. Check EVENT VIEWER in guest, didn't find any error log

Svra44:
@xiagao ,
There arent any specific steps to reproduce. I am seeing it on each of the VMs Ive checked so far. They would have been standard 10/Server2016 installs and then having virtio-win-gt-x64 0.1.173 installed.

Initially I identified the error because of a system having performance issues (running 0.1.173 blnsvr installed via virtio-win-gt-x64.msi I believe), and tracked the issue down to wmiprvse.exe CPU spikes which stopped once blnsvr service was stopped (& the 5858 errors stopped occurring).

Ive also tried uninstalling blnsvr and re-installing using 0.1.204 and have seen no change.
I am also seeing the same on a q35 VM, though most of my VMs are i440fx.

I setup a test VM using q35 with 21H1 19043.928 Windows 10 Pro N iso, copied "Balloon\w10\amd64\*" from virtio-win-0.1.204.iso to the C Drive and installed via "blnsvr.exe -i". I didnt see the errors occur, though as per below continuation, that may just be because I didnt wait long enough.
I then uninstalled blnsvr.exe and installed Balloon Service via virtio-win-gt-x64 and began seeing the errors on this new VM, though they took about 5 minutes after initial upstart of blnsvr.exe to start appearing in Microsoft-Windows-WMI-Activity, and seem to occur less frequently(~1 minute in intervals) than some other VMs I have checked.
Checking the file SHA1 hashes virtio-win-gt-x64 installs the correct w10 files, however it doesnt include the .pdb files (though manually copying them over doesnt change the behaviour).
[...]

PR #939, opened 2023-07-16
Consume the entire enumerator before letting it go out of scope

driehuis:
This diff seems to fix issue #614. I have replaced blnsvr.exe with a version that contains this patch, and the events are gone. I tested that blnsvr.exe built without this patch does exhibit the behavior. The service is running without any visible issues, but I have not been able to do a meaningful test of its functionality.

I compared the usage of IWbemServices:ExecQuery in memstat.cpp with other C++ code that uses the same API. The only material difference I noticed was that the other code used the form
```
    while(1) {
        hres = pEnumerator->Next(WBEM_INFINITE, 1, &pclsObj, &uReturn);
        if(0 == uReturn) {
            break;
        }
    }
```
which guarantees that the enumerator is consumed in its entirety. Apparently, the API doesn't clean up its own context unless all items are consumed and throws a hissy fit later on, and quite likely only on a few of the calls.

Please review the patch carefully. I think I got the indentation right according to the "When in Rome" rule, but I'm not positive. The patch is paranoid; the check for enumerator==NULL can probably be eliminated, and it could be made shorter.

I did not run the test suite.

### DIFF
--- a/Balloon/app/memstat.cpp
+++ b/Balloon/app/memstat.cpp
@@ -233,5 +233,15 @@ BOOL CMemStat::Update()
         m_Stats[idx++].val = 0;
     }
 
+    // Consume the enumerator
+    while (enumerator != NULL && !FAILED(enumerator->Next(
+        WBEM_INFINITE,
+        1L,
+        &memory,
+        &retcnt))) {
+            if (retcnt == 0)
+                break;
+    }
+
     return TRUE;
 }


YanVugenfirer:
Keeping the PR as a reminder

YanVugenfirer:
We have a Jira to track this bug: https://issues.redhat.com/browse/RHEL-12116

Closing the PR for now

PR #940, opened 2023-07-17
[viostor] fix multi-queue support in case if number of CPUs is not matching number on virtual queues

vrozenfe:
https://issues.redhat.com/browse/RHEL-465 - viostor driver

### DIFF
--- a/viostor/viostor.inx
+++ b/viostor/viostor.inx
@@ -97,7 +97,7 @@ HKR, "Parameters", "BusType", %REG_DWORD%, 0x00000001
 HKR, "Interrupt Management",, 0x00000010
 HKR, "Interrupt Management\MessageSignaledInterruptProperties",, 0x00000010
 HKR, "Interrupt Management\MessageSignaledInterruptProperties", MSISupported, 0x00010001, 1
-HKR, "Interrupt Management\MessageSignaledInterruptProperties", MessageNumberLimit, 0x00010001, 256
+HKR, "Interrupt Management\MessageSignaledInterruptProperties", MessageNumberLimit, 0x00010001, 257
 HKR, "Interrupt Management\Affinity Policy",, 0x00000010
 HKR, "Interrupt Management\Affinity Policy", DevicePolicy, 0x00010001, 5
 HKR, "Interrupt Management\Affinity Policy", GroupPolicy, 0x00010001, 1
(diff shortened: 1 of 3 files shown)


viostor/viostor.inx
```diff
@@ -97,7 +97,7 @@ HKR, "Parameters", "BusType", %REG_DWORD%, 0x00000001
 HKR, "Interrupt Management",, 0x00000010
 HKR, "Interrupt Management\MessageSignaledInterruptProperties",, 0x00000010
 HKR, "Interrupt Management\MessageSignaledInterruptProperties", MSISupported, 0x00010001, 1
-HKR, "Interrupt Management\MessageSignaledInterruptProperties", MessageNumberLimit, 0x00010001, 256
+HKR, "Interrupt Management\MessageSignaledInterruptProperties", MessageNumberLimit, 0x00010001, 257
```
YanVugenfirer, line 100: Why not increase it to the maximum?
https://github.com/virtio-win/kvm-guest-drivers-windows/blob/5c54cacb2ed054f9737fdadb84b3e4cd235a9928/NetKVM/netkvm-base.txt#L37

viostor/virtio_stor_hw_helper.c
```diff
@@ -43,6 +43,15 @@
                       pa = va ? StorPortGetPhysicalAddress(DeviceExtension, NULL, va, &len).QuadPart : 0; \
                     }
 
+#define MESSAGENUMBER_TO_QUEUE() { \
+                    MessageId = param.MessageNumber; \
+                    QueueNumber = MessageId - 1; \
+                    if (QueueNumber >= adaptExt->num_queues) { \
+                        QueueNumber %= adaptExt->num_queues; \
+                        MessageId += 1; \
```
YanVugenfirer, line 51: Why would this happen if the number of queues is limited during init?

YanVugenfirer (approved):
Approved, with a comment that maybe it is worth increasing the max number of possible MSI interrupts to maximum

viostor/virtio_stor_hw_helper.c
vrozenfe: It is a kind of sanity check. I was not able create such situation yet, but if you hotplug a vCPU and crate an application which creates a thread, set the thread affinity to this new CPU and then run a disk IO operation , then
CPU index (queue number) in this case will be bigger than actual number of queues. Unless you don't specify  some more num_queues in the qemu command line, of cause. 

Cheers,
Vadim.

PR #947, opened 2023-07-28
viosock: select IO control should return with 'Information' in success

sochinpub:
When select called to poll events of vsock socket, there is a large chance of failure with WSAEINVAL code, especially with timeout param. The trace events gives "Invalid output len" output, which help us to find this bug, that IOCTL_SELECT should always return with correct 'BytesReturned' in success.


YanVugenfirer (approved):

YanVugenfirer:
Thanks for sending the fix!

PR #948, opened 2023-07-28
viosock: fix one fatal recv data error bug

sochinpub:
This patch fixs one bug, that multiple recv calls always return the same data. From upper layer, it seems like Stream data pointer never moves. One simple multiple-recvs test case can reproduce this easily.


YanVugenfirer (approved):

YanVugenfirer:
Thanks for sending the fix!

YanVugenfirer (commented):
Looks like a pattern of errors: https://github.com/virtio-win/kvm-guest-drivers-windows/pull/948/commits/026415dbf5a5f9f60de530f2698bfaef307b9596

PR #951, opened 2023-08-03
viosock: make sure to get a correct computation

sochinpub:
Assign the corresponding RxCb's attribute to zero only after it is not used, otherwise some computational errors may cause unexpected behivor.


YanVugenfirer (approved):
https://github.com/virtio-win/kvm-guest-drivers-windows/commit/026415dbf5a5f9f60de530f2698bfaef307b9596

PR #952, opened 2023-08-03
viosock: fix one live lock issue

sochinpub:
using traceview to see tracing events, it turns out the windows thread id handle returned by PsGetCurrentThreadId() is really an integer, this means a NULL handle can be equal to a zero value thread ID(handle is 0), thus cause one infinit wait live lock issue which I actually encounters. Anyway, use a non-positive integer value than this handle seems to be safer.

### DIFF
--- a/viosock/sys/Rx.c
+++ b/viosock/sys/Rx.c
@@ -1309,7 +1309,7 @@ VIOSockReadDequeueCb(
 
     TraceEvents(TRACE_LEVEL_VERBOSE, DBG_READ, "--> %s\n", __FUNCTION__);
 
-    if (pSocket->RxProcessingThreadId == PsGetCurrentThreadId())
+    if (pSocket->RxProcessingThreadId == (LONG64)PsGetCurrentThreadId())
     {
         TraceEvents(TRACE_LEVEL_INFORMATION, DBG_READ, "Another instance of VIOSockReadDequeueCb already running, stop Cb dequeue\n");
         return FALSE; //one running instance allowed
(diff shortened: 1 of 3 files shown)


sochinpub:
image

YanVugenfirer (commented):
Is it an actual lock you encountered?

Because if the fix is theoretical, then it looks to be actually wrong:
1. Checkout the user pace counterpart of PsGetCurrentThreadId documentation [GetCurrentThreadId](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentthreadid) - "Note that no thread identifier will ever be 0."
2. HANDLE defined as  ```typedef PVOID HANDLE;```  so it looks wrong to use -1 as an error for PVOID type.

sochinpub:
> Is it an actual lock you encountered?
> 
> Because if the fix is theoretical, then it looks to be actually wrong:
> 
> 1. Checkout the user pace counterpart of PsGetCurrentThreadId documentation [GetCurrentThreadId](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentthreadid) - "Note that no thread identifier will ever be 0."
> 2. HANDLE defined as  `typedef PVOID HANDLE;`  so it looks wrong to use -1 as an error for PVOID type.

It's not theoretical, can you see the picture I provided above ?
I actually print this specific handle out by using %lld, the tracing log shows "Another instance of .... Cb dequeue(%lld) ", 
which means the handle is exactly the same with the Thread ID 0 which you can see  from the picture also.

1) VIOSockReadDequeueCb() can be called from this following call stack:
VIOSockRxVqProcess->VIOSockRxPktHandleConnected->VIOSockReadProcessDequeueCb->VIOSockReadDequeueCb
but may found empty read queue at last.
2) meanwhile VIOSockReadDequeueCb can be called from this following call stack either(from any thread context):
 VIOSockReadSocketState -> VIOSockReadProcessDequeueCb -> VIOSockReadDequeueCb) trigger by a upper layer recv call, 
but may exit immediately, thus cause the recv application wait infinitely.

No data send from the send side anymore.

YanVugenfirer (approved):

PR #954, opened 2023-08-04
viosock: remove an extra minus in select timeout computation

sochinpub:
VIOSockTimerStart() already contains a minux symbol, using an extra minux in VIOSockSelect() is needless.


YanVugenfirer (commented):
Looks like this code https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/viosock/sys/viosock.h#L869 worked always in case of Select in unintended manner. Did you check that after the change there are no side effects?

YanVugenfirer (approved):

PR #957, opened 2023-08-10
viosock: fix NtReadFile/NtWriteFile; fix sending large packet sg list

Jing118:
This PR fixes 2 issues:

1. Wait on a Event instead of the socket itself after calling NtWriteFile()/NtReadFile(), otherwise WaitForSingleObject() may return before the write/read completes, thus returning wrong status to caller. This issue can be reproduced by having a sender thread and a receiver thread running on a socket at the same time, and sometimes send() may return 0 bytes sent, even though the other side can receive the packet.

2. If the first element in SgList is not 4KB, there will be 17 element in SgList for 64KB data, which will cause a stack buffer overrun BSOD. This issue can be reproduced by sending more then 64KB data in one send().

### DIFF
--- a/viosock/lib/native.c
+++ b/viosock/lib/native.c
@@ -255,15 +255,22 @@ VIOSockWriteFile(
     if (lpNumberOfBytesWritten)
         *lpNumberOfBytesWritten = 0;
 
-    status = NtWriteFile((HANDLE)s, NULL, NULL, NULL,
+    HANDLE evt = CreateEvent(NULL, FALSE, FALSE, NULL);
+    if (!evt) 
+    {
+        return FALSE;
+    }
+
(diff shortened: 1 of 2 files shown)


YanVugenfirer:
Hi @Jing118 ,

1. Please split the patches.
2. Please add Signed-off-by line to the end of the commit message. Check contribution guidelines: https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/contributing.md
3.  For the TX fix, it is a bad practice to allocate memory on stack for SG in any case. I suggest to make a real fix that will allocate memory properly (also with consideration of probable usage of IOMMU) and not workaround. 

Thanks,
Yan.

YanVugenfirer (changes requested):
Please change according to my previous comment.

PR #959, opened 2023-08-15
viosock: fix NtReadFile/NtWriteFile

Jing118:
Wait on a Event instead of the socket itself after calling NtWriteFile()/NtReadFile(), otherwise WaitForSingleObject() may return before the write/read completes, thus returning wrong status to caller. This issue can be reproduced by having a sender thread and a receiver thread running on a socket at the same time, and sometimes send() may return 0 bytes sent, even though the other side can receive the packet.

### DIFF
--- a/viosock/lib/native.c
+++ b/viosock/lib/native.c
@@ -214,14 +214,21 @@ VIOSockDeviceControl(
     if(lpBytesReturned)
         *lpBytesReturned = 0;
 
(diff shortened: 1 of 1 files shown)


YanVugenfirer (changes requested):

YanVugenfirer (approved):
Waiting for CI to run

YanVugenfirer:
rerun tests

PR #960, opened 2023-08-15
viosock: fix recv stuck after process terminated

Jing118:
Like the previous PR #953, we should also process the RxPktList in VIOSockReadCleanupCb , otherwise the same stuck issue may occur after the process terminated, when all buffers in RxCbList return to RxCbBuffers in VIOSockReadCleanupCb.


YanVugenfirer (approved):

YanVugenfirer (changes requested):

YanVugenfirer (approved):

PR #963, opened 2023-08-15
viosock: fix CurrentCb BytesToRead setting

Jing118:
We should set pCurrentCb->BytesToRead to 0 before calling VIOSockRxCbPushLocked(). Consider following scenario:
1. thread 1 calls VIOSockRxCbPushLocked(), add CurrentCb to RxCbBuffers
2. thread 2 calls VIOSockRxPktInsert(), get the CurrentCb from RxCbBuffers, and uses virtqueue_add_buf() to add the CurrentCb to RxVq
3. thread 2 calls VIOSockRxVqProcess(), get the CurrentCb from virtqueue_get_buf(), and ready to be read from user recv() call
4. thread 1 sets pCurrentCb->BytesToRead to 0, so it modifies the CurrentCb used by thread 2

### DIFF
--- a/viosock/sys/Rx.c
+++ b/viosock/sys/Rx.c
@@ -1445,7 +1445,8 @@ VIOSockReadDequeueCb(
 
             if (!(pRequest->Flags & MSG_PEEK))
             {
-
+                ULONG uBytesToRead = pCurrentCb->BytesToRead;
+                pCurrentCb->BytesToRead = 0;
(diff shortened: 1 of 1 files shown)


YanVugenfirer (changes requested):

YanVugenfirer (approved):
Please rebase

Jing118:
Rebased.

PR #967, opened 2023-08-17
viogpu: Correctly copy the DXGKRNL_INTERFACE structure

MartinDrab:
During its DxgkDdiStartDevice routine, the viogpudo driver is provided with an DXGKRNL_INTERFACE structure containing mostly pointers to various WDDM interface functions that it may need to use. The driver copies this structure to its device-specific block of memory. However, the driver expected the system would provide the same version of the structure it uses, however, this is not necessarily true. Such behavior resulted in copying memory past the end of the system-provided structure which triggered the crash.

Fixed version of the driver copies only bytes really occupied by the DXGKRNL_INTERFACE structure (stored in its Size member).

I initially noticed this problem with `qxldod` (https://gitlab.freedesktop.org/spice/win32/qxl-wddm-dod/-/merge_requests/2) but `viogpu` uses the same code here.


vrozenfe:
looks good. 
Just please remove qxl and qxldod statement from the trace,

Best,
Vadim.

MartinDrab:
> looks good. Just please remove qxl and qxldod statement from the trace,
> 
> Best, Vadim.

Ah, I forgot to change them. It should be OK now.

vrozenfe (approved):

PR #1005, opened 2023-12-01
vioserial : host connection status set to false when driver exit

54shady:
In some scenario such as suspend the guest, the host will notify connect close, but the status can not be sync to guest driver,
which will cause a bug while the guest write the seiral before the host connect happend. so it's better just let the connection status to be false rather then true. because the next boot(resume) will enable the connection again

### DIFF
--- a/vioserial/sys/Port.c
+++ b/vioserial/sys/Port.c
@@ -1454,6 +1454,7 @@ VIOSerialPortEvtDeviceD0Exit(
         __FUNCTION__, Port->PortId, TargetState - WdfPowerDeviceD0);
 
     Port->Removed = TRUE;
+    Port->HostConnected = FALSE;
 
     VIOSerialDisableInterruptQueue(GetInQueue(Port));
 


YanVugenfirer:
Hi @54shady 

Thank you for the patch!
Did you test both S3 (sleep) and S4 (hibernate) scenarios?

6-dehan:
> OK from my side.
> 
> Pending Red Hat QE review.

yes, @YanVugenfirer  I'm on it. but I got a littler problem with setting S3/S4 power state for windows VM. 
And could you please provide more testing details you've operated @54shady ?

6-dehan:
image

6-dehan:
Update:
Currently, S3/S4(sleep/hibernate) power mode can be executed as we expected. before asking confirmation, let me give a brief explanation of how it goes now.

The test steps are as the following:
1. install the internal prewhql drivers (viogpudo/vioserial) on win2022
2. send data from host to guest via virtio-serial channel and receive from guest, at the same time, simultaneously perform hibernation operation `shutdown -h`
3. the actual result from the host is the connection disconnected while the VM hibernated successfully. 
4.  re-boot the VM and all result as you can see the pic
image

54shady:
> Update: Currently, S3/S4(sleep/hibernate) power mode can be executed as we expected. before asking confirmation, let me give a brief explanation of how it goes now.
> 
> The test steps are as the following:
> 
> 1. install both of the the internal prewhql drivers (viogpudo/vioserial) and the driver applied this patch on win2022
> 2. send data from host to guest via virtio-serial channel and receive from guest, at the same time, simultaneously perform hibernation operation `shutdown -h`
> 3. the actual result from the host is the connection disconnected while the VM hibernated successfully.
> 4. re-boot the VM and all result as you can see the pic
>    image
> 
> So they got the same results. maybe there are some misunderstandings or wrong operation, therefore, could you please describe more on it @54shady

Hello,

Test as below:

1. bootup virtual machine and enable the connect the from host side, which will make guest driver HostConneced = 1
2. in Guest, do S3/S4 sleep, for S4 (which is the bug lies in)  the host side will disconnect, because qemu quit
3. bootup (wakeup or resume)  the Guest,  this time, don't enable the connection from the Host side, run app in Guest to write the serial( it's better let the write failed immediately )

6-dehan:
> Hello,
> 
> Test as below:
> 
> 1. bootup virtual machine and enable the connect the from host side, which will make guest driver HostConneced = 1
what I got is only 'Is Connected = True', looks like what you said 'HostConnected' :
'Device manager' --> "Ports (COM & LPT)" --> check "Properties" --> Navigate to "Details" --> then choose 'Is connected' 
> 2. in Guest, do S3/S4 sleep, for S4 (which is the bug lies in)  the host side will disconnect, because qemu quit
Yeah, it happened. 
> 3. bootup (wakeup or resume)  the Guest,  this time, don't enable the connection from the Host side, run app in Guest to write the serial( it's better let the write failed immediately )
but I didn't hit this one. data sent well.

6-dehan:
>what's you host side application?

not an application at all, just a python script. 
> The original code which will keep the connection status in HostConneded = 1 which is the key why the write can get through

how and where did you get 'HostConnected' on windows?

54shady:
> > what's you host side application?
> 
> not an application at all, just a python script.
> 
> > The original code which will keep the connection status in HostConneded = 1 which is the key why the write can get through
> 
> how and where did you get 'HostConnected' on windows?

the python script read/write the which file on the host? when trigger the S4, the script still running?

the  HostConnected I refer, is in the code as below:
VIOSerialWillWriteBlock(
    IN PVIOSERIAL_PORT port
)
{
    BOOLEAN ret = FALSE;

    TraceEvents(TRACE_LEVEL_INFORMATION, DBG_PNP,"--> %s\n", __FUNCTION__);
    if (!port->HostConnected) <================================================ This line here
    {
        return TRUE;
    }

6-dehan:
> > 
> > how and where did you get 'HostConnected' on windows?
> 
> the python script read/write which file on the host? when trigger the S4, the script still running?

on the host side surely it disconnected. but guest will send data successfully after re-wake up. but I'll change the way to write data from guest,  will update later. 
> 
> the HostConnected I refer, is in the code as below: VIOSerialWillWriteBlock( IN PVIOSERIAL_PORT port ) { BOOLEAN ret = FALSE;
> 
> ```
> TraceEvents(TRACE_LEVEL_INFORMATION, DBG_PNP,"--> %s\n", __FUNCTION__);
> if (!port->HostConnected) <================================================ This line here
> {
>     return TRUE;
> }
> ```
from the code level. alright, I thought about the system level that we can see via cmd or 'device manager' stuff.

6-dehan:
> 
> 1. boot vm
>    PyScript----> qemu----> guest
> 2. do S4 in guest, so the PyScript should be  in the status disconnect to  qemu
>    PyScript----X------> qemu----> guest
> 3. no resume the virtual machine
>    qemu-----> guest
> 4. do write in guest
>    qemu<-----write----guest,          // the PyScript is not connected, so it's write should get failed

yes, as same as what you did, if I didn't resume the virtual machine, I got the write failed. but no matter with our driver or your driver.  In addition to preventing some vioserial potential damage, is there any obvious functional difference if we apply your patch?

YanVugenfirer:
@54shady do you have a real life scenario for this case?

YanVugenfirer:
I need to look at virtio spec also to see if this scenario is covered. 
I am not sure, silently not doing anything is a better solution.

54shady:
> @54shady do you have a real life scenario for this case?

the spice share directory is the real case

yes, this issue can be handle in application,  for example application delay the write before the host really connected, or just do try write timeout 

maybe this is not covered in virtio spec. hardcode the HostConnected = False in code is better then do nothing, this fix is not a MUST, it's a MUCH better chose one

54shady:
> > @54shady do you have a real life scenario for this case?
> 
> the spice share directory is the real case
> 
> yes, this issue can be handle in application, for example application delay the write before the host really connected, or just do try write timeout
> 
> maybe this is not covered in virtio spec. hardcode the HostConnected = False in code is better then do nothing, this fix is not a MUST, it's a MUCH better chose

54shady:
> ifference if we apply

the core different is : the HostConnected status in the code sync the real (represent the real current status)  in the above scenario

YanVugenfirer:
I am going to merge. But in my opinions this is a virtio spec behaviour omission that should have being resolve with guest side also notifying VIRTIO_CONSOLE_PORT_OPEN 0 in PM scenarios.

PR #1008, opened 2023-12-12
Add Windows 11 target

kostyanf14:
Update build scripts to use EWDK 11 21H2 for Win10 target and EWDK 11 22H2 for Win11 target


ybendito:
I see that in the projects Win11/x86 configuration were added. Of course we can remove them later but each such removal will request the change in the root virtio-win.sln (change project ref config to 'Win11/x64/nobuild') and will trigger full rebuild.
If possible I would suggest to make the change in root SLN file, then specific projects can be changed without triggering full rebuild.

kostyanf14:
@YanVugenfirer 

The build part is finished. SDV in progress.

ybendito:
+1 for me. There are some auxilliary 32-bit executables that will be good to have for test purposes only. But in case of need the binaries for Win10 can be used.

YanVugenfirer (approved):

PR #1011, opened 2023-12-13
netkvm: remove query for OID_GEN_TRANSMIT_QUEUE_LENGTH

ybendito:
The response to the query is calculated incorrectly. We can make it correct but I think that support for this OID is not required, it is not in the list of mandatory OIDs. In general the NDIS has the same knowledge about how many packets are queued by the miniport.


YanVugenfirer (approved):
Let's wait for CI

ybendito:
rerun tests

YanVugenfirer (approved):

PR #1016, opened 2023-12-24
Rx queue size adaptive (up to 1024) + minor fixes

ybendito:
https://issues.redhat.com/browse/RHEL-19430

### DIFF
--- a/NetKVM/CoInstaller/netkvmco-vs2015.vcxproj
+++ b/NetKVM/CoInstaller/netkvmco-vs2015.vcxproj
@@ -176,33 +176,13 @@
     <ClCompile Include="regaccess.cpp" />
     <ClCompile Include="regparam.cpp" />
     <ClCompile Include="tstrings.cpp" />
-    <ResourceCompile Include="@(RcSourceFiles)" Exclude="@(ResourceCompile)" />
-    <Midl Include="@(IdlSourceFiles)" Exclude="@(Midl)" />
-    <MessageCompile Include="@(McSourceFiles)" Exclude="@(MessageCompile)" />
-    <MASM Include="@(AsmSourceFiles)" Exclude="@(MASM)" />
-    <GenerateBmf Include="@(MofSourceFiles)" Exclude="@(GenerateBmf)" />
+    <ResourceCompile Include="netco.rc"/>
   </ItemGroup>
-  <!-- Set default environment variables, e.g. for stampinf -->
   <ItemGroup>
     <BuildMacro Include="SDK_INC_PATH">
       <Value>$(KIT_SHARED_INC_PATH)</Value>
       <EnvironmentVariable>true</EnvironmentVariable>
     </BuildMacro>
-  </ItemGroup>
-  <ItemGroup>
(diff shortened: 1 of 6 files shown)


YanVugenfirer (changes requested):
I frankly disagree with this change. 

The convention is new->create or new->init.  Having different naming for the essentially "create\init" creates confusion. 
BTW: this convention exists as a result of "new" being both allocator and initialiser that might fail. I agree it is awkward for the cases when init alway return TRUE, but again - better then different naming in every class.

YanVugenfirer (changes requested):
I suggest to break PR to several topic related PRs

YanVugenfirer (changes requested):
Please create separate PR

YanVugenfirer:
I suggest breaking into 3 PRs and creating Jiras for each one of them (one already exists):
1. Changing RX queue size default
2. Statistics (please update the WiKi and add wiki link to commit message)
3. Filtering changes

I also don't think we need to push to master the primary\secondary allocations to master as it doesn't do anything now. Better keep it in the topical branch.

ybendito:
> I frankly disagree with this change.
> 
> The convention is new->create or new->init. Having different naming for the essentially "create\init" creates confusion. BTW: this convention exists as a result of "new" being both allocator and initialiser that might fail. I agree it is awkward for the cases when init alway return TRUE, but again - better then different naming in every class.

I'm not aware of such convention. This very depends on the type of the object, for some of objects it is sometimes hard to initialize them in a constructor. For these this is really needed. We have various types of objects and for many of them we do not have any create(). I specifically removed such "creates" that do not do anything. Probably for these object I need to pass the requires parameters in the constructor, this is less trivial but possible.

ybendito:
Win11-22H2 failure: E2EPerf test Test Case: __COUNTER__ | 88888 | Receiver failed to coalesce the minimum percentage of received packets (70%). Only 0% of receives were coalesced.

YanVugenfirer (approved):

ybendito:
@kostyanf14 @YanVugenfirer 
The RSS test failure:
image
AFAICS this failure should be covered by Filter 5442
[UpdateFilters.txt](https://github.com/virtio-win/kvm-guest-drivers-windows/files/13955581/UpdateFilters.txt)
Are we sure we have uo-2-date filters on the controller?
Is it possible to add the exact date of UpdateFilters.sql file of the controller to the logs?

kostyanf14:
@ybendito 

I checked filters SQL.
```
# grep -e 5442 UpdateFilters.sql 
-- Inserting filter 5442 v12.
SELECT @FilterId = Id FROM Filter WHERE FilterNumber = 5442 AND Version = 12
	VALUES(5442, 12, 1, 1, 1, 1, 0, 0, @TestCommandLineId, 'HLK: NDIS 6.0 Tests fail with LibProtocolDriver: Install protocol driver failed.', 'The following HLK NDIS 6.0 Tests may fail with "LibProtocolDriver: Install protocol driver failed.":
```

I think we can add some info about filters.

PR #1021, opened 2024-01-04
netkvm: filter RX packets as early as possible

ybendito:
https://issues.redhat.com/browse/RHEL-20575

Currently the driver get RX descriptor from the rx virtQ, analyses it (L2/L3), calculates hash and delivers to the proper RSS queue. Later it extract the packet from the RSS queue and checks whether it passes RX filters.
Now we move the filtering to the earliest stage - just after basic analysis.
Note that in case of vhost there are no receive filters configured by the host and typically the host does not work in promiscuous mode, so it is redundant to calculate the hash and trigger the RSS queue for packets that should be dropped.

### DIFF
--- a/NetKVM/Common/ParaNdis_Common.cpp
+++ b/NetKVM/Common/ParaNdis_Common.cpp
@@ -1462,63 +1462,6 @@ VOID ParaNdis_OnShutdown(PARANDIS_ADAPTER *pContext)
     pContext->m_StateMachine.NotifyShutdown();
 }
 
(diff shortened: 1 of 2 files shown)


ybendito:
Win11 CentOS - only glitch free
2019 Fedora - RSS test failed, looks like internal problem of HCK (25 instances of the protocol loaded, the last one - can't load)
Win22 Fedora - only Glitch free
@kostyanf14 for completeness I'd run the RSS send receive test only.

YanVugenfirer (approved):

PR #1023, opened 2024-01-14
[vioscsi] add support for tracing SRB processing time 

vrozenfe:
https://issues.redhat.com/browse/RHEL-19188

### DIFF
--- a/vioscsi/vioscsi.h
+++ b/vioscsi/vioscsi.h
@@ -334,9 +334,7 @@ typedef struct _ADAPTER_EXTENSION {
     UCHAR                 rev_id[4 + 1];
     BOOLEAN               reset_in_progress;
     ACTION_ON_RESET       action_on_reset;
-#if (NTDDI_VERSION >= NTDDI_WINTHRESHOLD)
     ULONGLONG             fw_ver;
-#endif
 } ADAPTER_EXTENSION, * PADAPTER_EXTENSION;
 
 #ifndef PCIX_TABLE_POINTER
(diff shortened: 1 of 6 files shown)


vioscsi/vioscsi.h
```diff
@@ -60,6 +60,7 @@ typedef struct VirtIOBufferDescriptor VIO_SG, *PVIO_SG;
 
 #define REGISTRY_MAX_PH_BREAKS           "PhysicalBreaks"
 #define REGISTRY_ACTION_ON_RESET         "VioscsiActionOnReset"
+#define REGISTRY_RESP_TIME_LIMIT         "ResponseTime"
```
JonKohler, line 63: Nice! If I'm reading this right, this would be to debug IO that took over X amount of time, which is dictated by the ResponseTime reg key. If that's the only goal, would it make sense to do something like a TraceResponseTime or LogOutlierResponseTime ?

JonKohler (commented):
Great timing as this is something I was scratching my head on just recently. One small question inline

vioscsi/vioscsi.h
vrozenfe: Sure, I will change "ResponseTime" to "TraceResponseTime" shortly.
And yes, the main purpose of the patch is to make it easier for the end users and field engineers to 
troubleshoot  vioscsi/backend timing issues in production environments. 
Best,
Vadim.

YanVugenfirer (approved):

vioscsi/vioscsi.h
YanVugenfirer: @JonKohler Are you OK with the change?

JonKohler: yea, that is OK with me. Thanks for making this happen!

Issue #1012, opened 2023-12-15
packet loss and solutions


80886:
**Describe**

When a TCP-based application enables the TCP_NODELAY option by `setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)` (disables TCP Nagle), if the TCP Server does not receive packets as efficiently as possible, the operating system's protocol stack's special behavior when receiving packets can temporarily lead to the exhaustion of the RX Descriptors, and resulting in packet loss.

**To Reproduce**

Create two same VMs:

- 4 cores
- 8GB RAM
- A virtio network card (queues:2, queue_size:1024)
- Install Windows Server 2019 Datacenter and apply the latest update

On VM1:

1. Run `python.exe ./server.py 8000`.

On VM2:
1. First, run `python.exe ./client.py 256 $VM1_IP 8000`.
2. Then, in another command prompt window, run `ping $VM1_IP -t`.

(server.py and client.py can be found in the Additional context section.)

**Expected behavior**

When pinging VM1 from VM2, severe packet loss should not occur.

**VM:**
 - Windows Server 2019 Datacenter(with update Nov,2023)
 - Can be reproduced with both the latest master build and the latest stable version(virtio-win-0.1.240.iso) of the Netkvm driver.

**Additional context**

****server.py:****
```python
import asyncio
import time
import random
import os
import sys

async def handle_echo(reader, writer):
    try:
        while True:
            data = await reader.read(1024)
            if not data:
                break
            await asyncio.sleep(0.2)
    except Exception as e:
        print(f'Error in handle_echo: {str(e)}')
    finally:
        if not writer.is_closing():
                writer.close()
                await writer.wait_closed()

async def main():
    server = await asyncio.start_server(
        handle_echo, '0.0.0.0', int(sys.argv[1]))

    addr = server.sockets[0].getsockname()
    print(f'Serving on {addr}')

    async with server:
        try:
            await server.serve_forever()
        except Exception as e:
            print(f'Error in server: {e}')

asyncio.run(main())

```

****client.py:****
```python
import sys
import socket
import asyncio
import os
import psutil

def is_vms_greater_than_1G():
    pid = os.getpid()

    proc = psutil.Process(pid)

    mem_info = proc.memory_info()

    return mem_info.vms > (1024 * 1024 * 1024)

msg = b"test"

async def tcp_echo_client(id, host, port):
    loop = asyncio.get_running_loop()

    while True:
        try:
[...]

PR #1028, opened 2024-01-16
[netkvm] add a workaround for rx packet loss

80886:
Implemented a workaround for rx packet loss encountered with TCP_NODELAY. By default, this feature is turned off to minimize the impact on CPU performance and network bandwidth.

To activate the feature, go to the network card properties and set the MinRxBufferPercent to a value above 0. A good starting point is to set MinRxBufferPercent to 25% and then monitor for any packet loss. If packet loss continues, incrementally adjust the value to 50%, then 75%, and if necessary, up to the maximum of 100%.

In extreme cases, it may be required to not only adjust the  MinRxBufferPercent but also to increase the number of queues and the size of each queue (queue_size).

It is important to note that this feature is only applicable to rx packet loss.

For more details, see [this thread](https://github.com/virtio-win/kvm-guest-drivers-windows/issues/1012).

### DIFF
--- a/NetKVM/Common/ParaNdis-RX.h
+++ b/NetKVM/Common/ParaNdis-RX.h
@@ -22,6 +22,8 @@ public:
         ReuseReceiveBufferNoLock(pBuffersDescriptor);
     }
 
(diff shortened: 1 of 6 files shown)


ybendito (approved):
IMO everything seems OK

YanVugenfirer (approved):

PR #1029, opened 2024-01-16
[vioserial] Update MSI interrupt limit

kostyanf14:
Resolves: https://issues.redhat.com/browse/RHEL-21792


YanVugenfirer:
Just to add some comment - maybe if CI is still not running, even in the commit message: up until now when several ports were used only one MSI interrupt would be used.

PR #1031, opened 2024-01-20
Additional metrics for NetKVM

ybendito:


ybendito:
Win11-22H2 Failed E2E perf test:
One of iterations (receive 64K TCP) fails with:
**Receiver failed to coalesce the minimum percentage of received packets (70%). Only 0% of receives were coalesced.**
Note that previous iteration (receive 128K TCP) finished successfully and coalesced 100% of packets.
In the next iteration (receive 4K TCP) also no problem happened, 98% coalesced (why not 100%, I do not know)
In the problematic iteration the adapter did not receive anything at all:
**Receiver failed to achieve minimum link utilization of 80%. Actual utilization 0.00%**
**Average Packets per Interrupt 0.0000 is less than the acceptable minimum 1.00**
So, the question is why in this specific iteration we do not receive (or do not send) anything
There is some problem with the test that should be understood.
And because this is CI - I think we need to add some workaround also - for example, label this test as "repeat on fail" and indeed repeat it if failed.

YanVugenfirer (approved):

PR #1033, opened 2024-01-22
RHEL-22354: netkvm: request to respawn the DPC if TX queue is not empty

ybendito:
In the crash of Mini6RSSSendRecv test found a situation when there is one packet is still in the SendQueue when the protocol driver waits for unfinished packet completion. Try to solve it by requesting additional DPC also in case when the TX queue is not empty (currently we request it only when there are completed packets in Tx queue and we can't enable interrupts for the queue).

### DIFF
--- a/NetKVM/Common/ParaNdis_TX.cpp
+++ b/NetKVM/Common/ParaNdis_TX.cpp
@@ -782,6 +782,9 @@ bool CParaNdisTX::DoPendingTasks(CNBL *nblHolder)
                 m_VirtQueue.ProcessTXCompletions(nbToFree);
                 bRestartQueueStatus = SendMapped(true, completedNBLs);
             }
+
+            bRestartQueueStatus |= !m_SendQueue.IsEmpty();
+
         } else
         {
             // the call initiated by Send(), we can give up


YanVugenfirer:
[2c_Mini6RSSSendRecv (Multi-Group Win8+)](https://docs.microsoft.com/en-us/windows-hardware/test/hlk/testref/6ce7b1b3-c5f2-4fe0-99f4-dbf55971411b) still fails

kostyanf14:
> [2c_Mini6RSSSendRecv (Multi-Group Win8+)](https://docs.microsoft.com/en-us/windows-hardware/test/hlk/testref/6ce7b1b3-c5f2-4fe0-99f4-dbf55971411b) still fails

There is no dump on Win11. The test just failed. 

On Win2022 we got a backchannel communication error and dump. I will upload it.

kostyanf14:
HCK-CI/NetKVM-Win2019x64 NDISTest 6.5 - [2 Machine] - OffloadLSO hangs on with error message box

image

YanVugenfirer:
This PR did not succeed to solve the issue.

Tracking in https://issues.redhat.com/browse/RHEL-22354

PR #1036, opened 2024-01-26
Viosock: Take Care of Recv Requests Stucks After Socket Shutdown

### LINKED ISSUE #1020: Viosock : recv stuck in client, while the socket is being SHUTDOWN from host
**Describe the bug**

on client side, stuck while call the recv, even using the latest driver code

**To Reproduce**
Run attachment application as below description

[server.txt](https://github.com/virtio-win/kvm-guest-drivers-windows/files/13807564/server.txt)
mv server.txt server.c and compile it to run on linux host
run server.c  on the Linux host

[client.txt](https://github.com/virtio-win/kvm-guest-drivers-windows/files/13807538/client.txt)
mv client.txt client.cpp and compile it to run in windows guest

**Expected behavior**

on client side, recv can return after the host SHUTDOWN the connection

**Screenshots**

on server ,  the socket is being close
image

on client windows,  the recv stuck:
image

**Host:**
seems no related to the host and qemu version and command line options etc.
 - Disto:  ubuntu 
 - Kernel 6.2.0-31-generic
 - QEMU   emulator version 7.0.50

 - QEMU 
 - qemu-system-x86_64 
        ...
        -device vhost-vsock-pci,guest-cid=3 \
    ...
 

**VM:**
 - Windows  win10 18363.592
 - vsock
 - the latest vsock driver

the problem i found when i comment out the server.txt line 192, which sleep 10 second after close the socket
what if i comment out the line 193, which do a write on host side, the recv stuck issue on client side is disappear

MartinDrab:
When an other end (peer) of the socket reports a shutdown, the `recv` request queue is not processed. This means that some recv calls may get stuck forever. This pull request changes the behavior -- when a shutdown from its peer is received, the recv queue gets processed and all requests that would get stuck are completed with `STATUS_SUCCESS` (which usually ends in receiving zero bytes).

This should fix #1020. I tested with
- client in WIndows VM, server on the host,
- - both client and server within the same Windows VM.

### DIFF
--- a/viosock/sys/Rx.c
+++ b/viosock/sys/Rx.c
@@ -1383,7 +1383,21 @@ VIOSockReadDequeueCb(
         }
         else
         {
-            bRequeue = TRUE;
+            // The socket is connected, however, it might
+            // receive a shutdown notification from the other end.
+            // In such a case, we need to complete the recv request with success,
+            // possibly reporting zero data received. If the shutdown is here,
+            // the request must not be requeued since that may cause it being
+            // in the queue forever.
+            status = VIOSockStateValidate(pSocket, FALSE);
+            if (!NT_SUCCESS(status))
+            {
+                bStop = TRUE;
+                if (status == STATUS_LOCAL_DISCONNECT ||
+                    status == STATUS_REMOTE_DISCONNECT)
+                    status = STATUS_SUCCESS;
+            }
+            else bRequeue = TRUE;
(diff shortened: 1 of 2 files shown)


YanVugenfirer (approved):

YanVugenfirer:
@MartinDrab Thanks!
Waiting for CI to run before merging

PR #1040, opened 2024-02-08
IVSHMEM: Allow multiple processes to map the same shared memory

polloloco:
Previously there was a `shmemMap` member inside the `DEVICE_CONTEXT` which effectively restricted the mapping to a single process only. I've moved that pointer (that holds the virtual address of the mapping in the usermode process) to a new `FILE_CONTEXT` struct. For each `CreateFile` call in usermode, a `WDFFILEOBJECT` is created in the kernel. The `WDFDEVICE` `EvtDeviceFileCreate` callback then allocates the file context in the file object.

When a usermode application requests a map via ioctl, we extract the file object (and then file context) from the `WDFREQUEST`. Instead of updating the `shmemMap` member in the device context (this was before this pull request), now we update that address in the file context, which is unique to each handle to the driver.

With these changes, one or more applications can map the same memory into their virtual address space. One application could even map the region twice, but only from two driver handles (two calls to `CreateFile`).


ivshmem/Queue.h
```diff
@@ -4,6 +4,7 @@ NTSTATUS IVSHMEMQueueInitialize(_In_ WDFDEVICE Device);
 
 EVT_WDF_IO_QUEUE_IO_DEVICE_CONTROL IVSHMEMEvtIoDeviceControl;
 EVT_WDF_IO_QUEUE_IO_STOP           IVSHMEMEvtIoStop;
+EVT_WDF_DEVICE_FILE_CREATE		   IVSHMEMEvtDeviceFileCreate;
```
YanVugenfirer, line 7: Looks like an indentation issue.

ivshmem/test/ivshmem-test.cpp
```diff
@@ -124,24 +124,24 @@ int __cdecl main()
 		}
 		TEST_PASS();
 
-		TEST_START("Mapping more then once fails");
-		if (DeviceIoControl(devHandle, IOCTL_IVSHMEM_REQUEST_MMAP, NULL, 0, &map, sizeof(IVSHMEM_MMAP), &ulReturnedLength, NULL))
+		TEST_START("Mapping more then once from the same handle fails");
+		if (DeviceIoControl(devHandle, IOCTL_IVSHMEM_REQUEST_MMAP, &config, sizeof(IVSHMEM_MMAP_CONFIG), &map, sizeof(IVSHMEM_MMAP), &ulReturnedLength, NULL))
```
YanVugenfirer, line 128: The line is too long even for my taste.

YanVugenfirer:
There are line-ending mismatches in the device.c file. Please fix.

YanVugenfirer:
There are line-ending issues in the device.c. It will be great if you fix them too.
Thanks.

ivshmem/Queue.c
```diff
@@ -172,15 +202,20 @@ VOID IVSHMEMEvtDeviceFileCleanup(_In_ WDFFILEOBJECT FileObject)
     }
     KeReleaseSpinLock(&deviceContext->eventListLock, oldIRQL);
 
-    if (!deviceContext->shmemMap)
+    if (!fileContext->shmemMap)
+        return;
+
+    MmUnmapLockedPages(fileContext->shmemMap, deviceContext->shmemMDL);
+    fileContext->shmemMap = NULL;
+    /*if (!deviceContext->shmemMap)
```
YanVugenfirer: I prefer to squash the commits of adding and removing the comments.

YanVugenfirer (commented):
Mostly minor cosmetic comments

YanVugenfirer:
@polloloco Are you planning to fix the comments?

Thanks, 
Yan.

YanVugenfirer (approved):

PR #1041, opened 2024-02-13
viogpu: Fix recursive includes

davispuh:
Currently there are several includes that cause recursive including - it still works because of include guards but it's not good idea for this to happen in first place and sometimes it can cause quite puzzling errors.

Some examples:
* helper.h -> viogpu_pci.h -> helper.h
* helper.h -> viogpu_idr.h -> helper.h

### DIFF
--- a/viogpu/common/baseobj.cpp
+++ b/viogpu/common/baseobj.cpp
@@ -1,4 +1,5 @@
 #include "baseobj.h"
+#include "viogpu.h"
 
 _When_((PoolType & NonPagedPoolMustSucceed) != 0,
     __drv_reportError("Must succeed pool allocations are forbidden. "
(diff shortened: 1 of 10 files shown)


vrozenfe (approved):
Thanks.
Vadim.

YanVugenfirer (changes requested):
I'm afraid I have to disagree on the ideological level with the change. If anything should be changed, then it is a virtio-related stuff that should be removed from the helper.h 
```
    #include "viogpu_pci.h"
     #include "viogpu_idr.h"
```
and added to viogpu.h. And then only virtiogpu.h to be included in the virtiopgu\graphics related headers.

davispuh:
I actually agree but that's larger change, I updated PR.

YanVugenfirer (approved):

YanVugenfirer:
rerun tests

PR #1042, opened 2024-02-14
viogpu: Decouple DBG from WPP

davispuh:
Previously `!DBG` meant event tracing which can be confusing.

Now with this PR it's possible to build these combinations:
* `DBG 1` - debug build
* `EVENT_TRACING 1` - WPP build (default)
* neither

### DIFF
--- a/viogpu/viogpudo/viogpudo.cpp
+++ b/viogpu/viogpudo/viogpudo.cpp
@@ -35,7 +35,7 @@
 #include "viogpum.h"
 #include "edid.h"
 
-#if !DBG
+#if defined(EVENT_TRACING)
 #include "viogpudo.tmh"
 #endif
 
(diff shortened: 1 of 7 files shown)


vrozenfe (approved):
Looks good.
Thank you.
Vadim.

viogpu/viogpudo/viogpudo.cpp
```diff
@@ -3411,6 +3411,8 @@ BOOLEAN VioGpuAdapter::CreateCursor(_In_ CONST DXGKARG_SETPOINTERSHAPE* pSetPoin
     UINT resid, format, size;
     VioGpuObj* obj;
     PAGED_CODE();
+    (void)pSetPointerShape; // unused
```
YanVugenfirer, line 3414: Sorry, why those two are added in this commit?

kostyanf14:
@YanVugenfirer Yes, two DF tests failing is a new one. I restarted the tests but it failed again in both PRs. Need to investigate the reason. Error occurs only in Win2022.

viogpu/viogpudo/viogpudo.cpp
davispuh: Because those variables are not used. They are only referenced with `DbgPrint` but when you build non-debug build where it's no-op you'll get compiler warning so by this we silence that warning.

kostyanf14:
rerun tests

viogpu/viogpudo/viogpudo.cpp
YanVugenfirer: The correct way to do it is to use UNREFERENCED_PARAMETER macro

```
UNREFERENCED_PARAMETER(pSetPointerShape);
UNREFERENCED_PARAMETER(pCurrentMode);
```

Like here: 
https://github.com/virtio-win/kvm-guest-drivers-windows/blob/ec4a51758f1d53281c8a4d78956239ba74c2aaa4/vioscsi/virtio_pci.c#L123

YanVugenfirer (changes requested):

PR #1043, opened 2024-02-14
Fix git executable flag

davispuh:
Source code files shouldn't be executable but only `*.{bat,cmd}` files.

Currently a lot of files are marked executable in git. This PR fixes that so that only `.bat`/`.cmd` are.

Note that on Windows this doesn't make any difference because of Git's `filemode = false` setting so it always ignores mode.


YanVugenfirer:
We are now adding clang formate CI. It will fix those issues

PR #1063, opened 2024-03-21
Set TargetOSVersion in INF for Windows 11

kostyanf14:
HLK for Server 2025 fail INF Verification with the error:

The syntax 'DIRID 13 (CopyFiles)' was introduced in OS version 10.0.16299, but DDInstall sections utilizing the syntax will install on earlier OS versions. Those DDInstall sections should be restricted to only install on 10.0.16299 or higher using a TargetOSVersion decoration.

As Windows 10 before build 16299 is EOL and we don't support it. This is ok, to have DIRID 13 in Windows 10 INF without TargetOSVersion decoration.

Resolve: https://issues.redhat.com/browse/RHEL-1214


YanVugenfirer (approved):

ybendito:
> As Windows 10 before build 16299 is EOL and we don't support it

What about Server 2016 which is in extended support?

YanVugenfirer (approved):

PR #1066, opened 2024-03-24
Viosock: Implement Advanced Calls for the WSK Interface

MartinDrab:
This PR implements more advanced calls for the WSK interface:
- `WskSocketConnect` (`WskSocket` + `WskBind` + `WskConnect` in one call),
- `WskConnectEx` (`WskConnect` + `WskSend`),
- `WskSendEx` (the same as `WskSend` for our socket case),
- `WskReceiveEx` (the same as `WskReceive` for our socket case).

The test driver has been updated to test also these calls.

**NOTE:** WSK callbacks (namely the accept and receive ones) are still not implemented since their implementation requires modification of the Socket driver whereas the current goal was to implement the WSK interface without any need to modify the Socket driver (apart from fixing various bugs).


viosock/wsk/socket.c
```diff
@@ -1017,6 +1017,7 @@ VioWskListen(
     _Inout_ PIRP     Irp
 )
 {
+    ULONG Backlog = 128;
```
YanVugenfirer, line 1026: I suggest not to use magic numbers. Please use const. 
Not critical comment.

YanVugenfirer (approved):

PR #1094, opened 2024-05-06
netkvm: fix affinity of poll handlers

ybendito:
https://issues.redhat.com/browse/RHEL-33334
Fix wrong affinity assignment.

### DIFF
--- a/NetKVM/wlh/ParaNdis_Poll.cpp
+++ b/NetKVM/wlh/ParaNdis_Poll.cpp
@@ -42,27 +42,37 @@ static void UpdatePollAffinities(PPARANDIS_ADAPTER pContext)
     DPrintf(0, "updated #%d affinities\n", done);
 }
 
-void ParaNdisPollSetAffinity(PARANDIS_ADAPTER* pContext, const CCHAR* Indices, CCHAR Size)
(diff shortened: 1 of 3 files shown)


NetKVM/wlh/ParaNdis_Poll.cpp
```diff
@@ -42,27 +42,37 @@ static void UpdatePollAffinities(PPARANDIS_ADAPTER pContext)
...
-                {
-                    poll->m_ProcessorNumber = number;
-                    poll->m_UpdateAffinity = true;
-                    needUpdate = true;
-                }
+                needUpdate = true;
             }
         }
     }
+
```
YanVugenfirer, line 75: BTW, why was NdisFreeIoWorkItem not called before the update?

ybendito: This does not matter when to free it in this specific case. In general (rule of the thumb) - the allocation may contain additional data needed during the processing, so it can be deleted when not needed.

YanVugenfirer (approved):

ybendito:
@YanVugenfirer The test 2c_Mini6RSSSendRecv (Multi-Group Win8+) on 2025 ran and passed (thanks @kostyanf14 )
The test 'Static tools' passes after populating RoMetadata.dll files from studio machine on client

PR #1118, opened 2024-07-24
[CI-NO-BUILD] Add interruption handler to CollectSystemInfo

6-dehan:
Add an interruption handler. When the script is not finished running, the generated file name could be marked with the
ongoing process, such as Collecting_Status.txt. and it would be delete after completion. In this way, when the script is waiting to be interrupted, it will be clearer to the userwhich directory was interrupted.

Solves: https://issues.redhat.com/browse/RHEL-50169
Signed-off-by: Dehan Meng <demeng@redhat.com>

### DIFF
--- a/Tools/debug/CollectSystemInfo.ps1
+++ b/Tools/debug/CollectSystemInfo.ps1
@@ -186,9 +186,12 @@ function Write-InformationToArchive {
 $timestamp = Get-Date -Format 'yyyy-MM-dd_HH-mm-ss'
 $folderName = "SystemInfo_$timestamp"
 $folderPath = Join-Path -Path (Get-Location) -ChildPath $folderName
+$progressFile = "$folderPath\Collecting_Status.txt"
 New-Item -Path $folderPath -ItemType Directory | Out-Null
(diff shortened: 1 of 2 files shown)


YanVugenfirer (changes requested):
Hi @6-dehan,

I have several comments:
1. Administrative comment. If possible indicated Jira either in committee message or in the commit title.
For example: 
<pre>[CI-NO-BUILD] RHEL-50169:Add interruption handler to CollectSystemInfo</pre>

Or: 

<pre>Add an interruption handler. When the script is not finished
running, the generated folder name could be marked with the
ongoing process, such as SystemInfo_202x-0xx_xx-xx-xx_pending or
SystemInfo_202x-0x-xx_xx-xx-xx_collecting. In this way, when the
script is waiting to be interrupted, it will be clearer to the
user which directory was interrupted.

Solves: https://issues.redhat.com/browse/RHEL-50169</pre>

2. I don't think that renaming directory will leaf toa good user experience. People don't have patience and will enter the directory while logs are created. In this case you will not be able to rename it in the end of the process.

YanVugenfirer (approved):

PR #1124, opened 2024-08-05
[CI-NO-BUILD] Classify original folders and solve a failure

6-dehan:
Splite original summary folder to Log and Dump folders for data classification.
And solve a problem that archive big file encounter failure.

Solves: https://issues.redhat.com/browse/RHEL-52721
Signed-off-by: Dehan Meng demeng@redhat.com

### DIFF
--- a/Tools/debug/CollectSystemInfo.ps1
+++ b/Tools/debug/CollectSystemInfo.ps1
@@ -43,6 +43,17 @@ param (
     [switch]$Help
 )
 
+Add-Type -AssemblyName 'System.IO.Compression.FileSystem'
+
+function Compress-Files {
+    param (
+        [string]$SourcePath,
+        [string]$DestinationPath
+    )
+
+    [System.IO.Compression.ZipFile]::CreateFromDirectory($SourcePath, $DestinationPath)
+}
+
 function Show-Help {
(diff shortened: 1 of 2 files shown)


Tools/debug/CollectSystemInfo.ps1
```diff
@@ -260,4 +282,7 @@ try {
 }
 
 Remove-Item -Path $progressFile -ErrorAction SilentlyContinue
-Write-InformationToArchive
+Write-InformationToArchive -FolderPath $folderPath -SubFolderPath $logfolderPath -ArchiveFileName $logfolderName
+if ($IncludeSensitiveData) {
+    Write-InformationToArchive -FolderPath $folderPath -SubFolderPath $dumpfolderPath -ArchiveFileName $dumpfolderName
```
YanVugenfirer, line 287: Will it work correctly if there is no dump file collected?

YanVugenfirer: Yes, I just want to be sure that there will be no errors in case if there are no dump files.

6-dehan: uh, if so, the answer gonna be sure. there is no error if there are no dump files. folder and archive zip package will be empty then.

YanVugenfirer (approved):

PR #1130, opened 2024-08-17
virtio-wdf: extend DMA API to support also RX transfer

ybendito:
Prerequisite for https://issues.redhat.com/browse/RHEL-11405 Added device-to-host transactions API.


ybendito:
Note: the change is only in Virtio-WDF, not in VirtIOLib, so  viogpu,netkvm,viostor,vioscsi  are not affected by the change.

YanVugenfirer (approved):

PR #1145, opened 2024-09-12
viostor: adapt to size_max

Atchess:
according to the definitions of NumberOfPhysicalBreaks and MaximumTransferLength in the Microsoft documentation, as well as the calculation methods provided in commit cfbf9345 for these two variables, in a single io, the number of segments  will not exceed seg_max, and the total size of the io will not exceed seg_max*size_max. therefore, when building io, dirver can split any sgement larger than size_max into smaller requests.

fix https://github.com/virtio-win/kvm-guest-drivers-windows/pull/753/commits

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -1263,6 +1263,8 @@ VirtIoBuildIo(
     ULONG                 dummy;
     ULONG                 sgElement;
     ULONG                 sgMaxElements;
+    ULONG                 sgLength;
+    ULONG                 sgOffset;
     PADAPTER_EXTENSION    adaptExt;
     PSRB_EXTENSION        srbExt;
     PSTOR_SCATTER_GATHER_LIST sgList;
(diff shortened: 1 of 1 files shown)


vrozenfe:
@Atchess 
Thanks a lot for your patch.
Sorry for asking this question, but can you please help me to understand the use case for this patch? 
I mean  can we test it for a real life qemu based scenario or is it supposed to be used for some "custom" 
DMA engine cases?

Best,
Vadim.

Atchess:
@vrozenfe 
Hi, for software-emulated devices, size_max may not be significant. However, for virtio-blk devices implemented in hardware, the size of the buffer is limited, making the size_max feature essential.

And the virtio specification defines this feature, so I think it is reasonable for drivers to implement it in accordance with the specification.

Best,
Qi.

vrozenfe:
@Atchess 
Yes, this is what I thought. Make sense for some sort of FPGAs . 
Out of curiosity, are you working with something like that? 

Best,
Vadim.

vrozenfe (approved):

Atchess:
@vrozenfe

Yes, I am working on related tasks.

Additionally, VIRTIO_F_ORDER_PLATFORM is also important for hardware-based devices, which is necessary to ensure that the device gets the correct descriptor. Although x86 viostor does not ack to this feature, it guarantees memory order in the code logic. Do we have plans to ack this feature in the future?

Best,
Qi.

vrozenfe:
@Atchess 

I would suggest opening a feature request to add VIRTIO_F_ORDER_PLATFORM support to virtio-win drives.
In my understanding it doesn't require some big changes in virtio ring code but then we need your or somebodies
else help to test the changes on a real HW.

Cheers,
Vadim.

Atchess:
@vrozenfe 

Yes, we don't need change the virtio ring code, the origin packed ring code already have right barriers.
such as virtqueue_add_buf_packed code:
```
    if (va_indirect && vq->num_free > 0) {
        desc = va_indirect;
        for (i = 0; i < descs_used; i++) {
            desc[i].flags = i < out ? 0 : VRING_DESC_F_WRITE;
            desc[i].addr = sg[i].physAddr.QuadPart;
            desc[i].len = sg[i].length;
        }
        vq->packed.vring.desc[head].addr = phys_indirect;
        vq->packed.vring.desc[head].len = descs_used * sizeof(struct vring_packed_desc);
        vq->packed.vring.desc[head].id = id;

        KeMemoryBarrier();
        vq->packed.vring.desc[head].flags = VRING_DESC_F_INDIRECT | vq->avail_used_flags;

        DPrintf(5, "Added buffer head %i to Q%d\n", head, vq->vq.index);
        head++;
        if (head >= vq->packed.vring.num) {
            head = 0;
            vq->packed.avail_wrap_counter ^= 1;
            vq->avail_used_flags ^=
                1 << VRING_PACKED_DESC_F_AVAIL |
                1 << VRING_PACKED_DESC_F_USED;
        }
        vq->packed.next_avail_idx = head;
        /* We're using some buffers from the free list. */
        vq->num_free -= 1;
        vq->num_added += 1;

        vq->free_head = vq->packed.desc_state[id].next;

        /* Store token and indirect buffer state. */
        vq->packed.desc_state[id].num = 1;
        vq->packed.desc_state[id].data = opaque;
        vq->packed.desc_state[id].last = id;

    } else {
```

And we have had some tests on our hardware/fpga, it works very well.

I will submit a PR later.

YanVugenfirer (approved):

PR #1147, opened 2024-09-14
virtio: add VIRTIO_F_ORDER_PLATFORM feature

Atchess:
VIRTIO_F_ORDER_PLATFORM may be required for devices implemented in hardware.
However when adding buffers in vring KeMemoryBarrier is used which prevents both the compiler and the processor from moving operations across the barrier. So just add this feature in guest_feature.

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -683,6 +683,10 @@ RhelSetGuestFeatures(
         guestFeatures |= (1ULL << VIRTIO_BLK_F_WRITE_ZEROES);
     }
 
+    if (CHECKBIT(adaptExt->features, VIRTIO_F_ORDER_PLATFORM)) {
+        guestFeatures |= (1ULL << VIRTIO_F_ORDER_PLATFORM);
+    }
+
     if (!NT_SUCCESS(virtio_set_features(&adaptExt->vdev, guestFeatures))) {
         RhelDbgPrint(TRACE_LEVEL_FATAL, " virtio_set_features failed\n");
         return;
--- a/VirtIO/linux/virtio_config.h
+++ b/VirtIO/linux/virtio_config.h
@@ -67,6 +67,12 @@
 /* This feature indicates support for the packed virtqueue layout. */
 #define VIRTIO_F_RING_PACKED            34
 
+/*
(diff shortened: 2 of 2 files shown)


viostor/virtio_stor.c
```diff
@@ -683,6 +683,10 @@ RhelSetGuestFeatures(
         guestFeatures |= (1ULL << VIRTIO_BLK_F_WRITE_ZEROES);
     }
 
+    if (CHECKBIT(adaptExt->features, VIRTIO_F_ORDER_PLATFORM)) {
+        guestFeatures |= (1ULL << VIRTIO_F_ORDER_PLATFORM);
```
YanVugenfirer, line 687: @ybendito, we should probably add it to the virtio-net too.

vrozenfe (approved):
Looks good.
Thanks,
Vadim.

viostor/virtio_stor.c
ybendito: @YanVugenfirer probably, to all the drivers, as the spec says:
A device MAY fail to operate further if VIRTIO_F_ORDER_PLATFORM is offered but not accepted. A device MAY operate in a slower emulation mode if VIRTIO_F_ORDER_PLATFORM is offered but not accepted
IMO, the drivers already use proper barriers or at least pretend to do so.

vrozenfe: In my understanding at the moment we don't honour VIRTIO_F_ORDER_PLATFORM flag in our code in any case. We don't have a notion of "weak barrier" as Linux does it is always "strong" and turned into lock operation. Which should be fine in case of real HW, but might be unnecessary in case of  multi-queue and proper IRQL handling where we should not hit out of order execution. But in any case, in my understanding if an underlying adaptor reports VIRTIO_F_ORDER_PLATFORM  flag it supposed that the driver would accept it if this feature supported by driver..

Please correct me if I'm wrong.
 
Cheers,
Vadim.

Atchess: @YanVugenfirer @kostyanf14 I may submit a PR for netkvm at first, I haven't read the code of other drivers, so maybe later.

PR #1152, opened 2024-09-19
[vioscsi] DPC Fixes

benyamin-codez:
Corrects:

1. A reference to a ghost routine for HW_PASSIVE_INITIALIZE_ROUTINE
2. Some missing braces in VioScsiInterrupt()

### DIFF
--- a/vioscsi/vioscsi.c
+++ b/vioscsi/vioscsi.c
@@ -66,7 +66,7 @@ HW_ADAPTER_CONTROL   VioScsiAdapterControl;
 HW_UNIT_CONTROL      VioScsiUnitControl;
 HW_INTERRUPT         VioScsiInterrupt;
 HW_DPC_ROUTINE       VioScsiCompleteDpcRoutine;
-HW_PASSIVE_INITIALIZE_ROUTINE         VioScsiIoPassiveInitializeRoutine;
+HW_PASSIVE_INITIALIZE_ROUTINE         VioScsiPassiveInitializeRoutine;
 HW_MESSAGE_SIGNALED_INTERRUPT_ROUTINE VioScsiMSInterrupt;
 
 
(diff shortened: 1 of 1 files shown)


vioscsi/vioscsi.c
```diff
@@ -66,7 +66,7 @@ HW_ADAPTER_CONTROL   VioScsiAdapterControl;
 HW_UNIT_CONTROL      VioScsiUnitControl;
 HW_INTERRUPT         VioScsiInterrupt;
 HW_DPC_ROUTINE       VioScsiCompleteDpcRoutine;
-HW_PASSIVE_INITIALIZE_ROUTINE         VioScsiIoPassiveInitializeRoutine;
```
JonKohler, line 69: Huh, I'm surprised it lets this compile!

Some passing notes:
https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nf-storport-storportenablepassiveinitialization#remarks

> A miniport must call the StorPortEnablePassiveInitialization routine only from within [HwStorInitialize](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nc-storport-hw_initialize). Otherwise, StorPortEnablePassiveInitialization will return FALSE and the [HwStorPassiveInitializeRoutine](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nc-storport-hw_passive_initialize_routine) routine will not execute.

https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/vioscsi/vioscsi.c#L854
Calls StorPortEnablePassiveInitialization with pointer to VioScsiPassiveInitializeRoutine from within VioScsiHwInitialize as mandated by doc above

https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nc-storport-hw_passive_initialize_routine 
Indicates that this would be run after initialization? Its unclear to me how this even gets called

benyamin-codez: It might be the prior prep in `VioScsiHwInitialize()` helps in some way...

https://github.com/virtio-win/kvm-guest-drivers-windows/blob/22d0908a28f382d02d8d3a10a0ee0631af480fdf/vioscsi/vioscsi.c#L780-L789

I think it most likely StorPort just accepts the routine specified with `StorPortEnablePassiveInitialization()` at L854 as you mentioned, and this overrides the callback previously defined by `HW_PASSIVE_INITIALIZE_ROUTINE`.

vrozenfe (approved):
@benyamin-codez 

Thank you !
Vadim.

PR #1150, opened 2024-09-19
[vioscsi] Fix SendSRB notify regression

benyamin-codez:
Related issues:
#756
#623
#907
...
[likely others]

Regression was introduced in #684 between https://github.com/virtio-win/kvm-guest-drivers-windows/commit/7dc052d55057c2a4dc2f64639d2f5ad6de34c4af and https://github.com/virtio-win/kvm-guest-drivers-windows/commit/fdf56ddfb44028b5295d2bdf20a30f5d90b2f13e.
Specifically due to commits https://github.com/virtio-win/kvm-guest-drivers-windows/commit/f1338bb7edd712bb9c3cf9c9b122e60a3f518408 and https://github.com/virtio-win/kvm-guest-drivers-windows/commit/fdf56ddfb44028b5295d2bdf20a30f5d90b2f13e (considered together).

~~Prior to the regression, we did not issue `virtqueue_notify()` ~~outside~~ inside spinlock. This behaviour has been restored.~~

Prior to the regression, we issued `virtqueue_kick_prepare()` inside spinlock. This behaviour has been restored.

_Note:_ We also do not want to issue `virtqueue_notify()` for a corrupt or failed buffer so this has been removed where `virtqueue_add_buf()` does not return SUCCESS. The `virtqueue_add_buf()` routine will return 0 on SUCCESS or otherwise a negative number, usually -28 (ENOSPC).

Freedom for Windows guests once held captive...! 8^D

cc: @vrozenfe @JonKohler @sb-ntnx @frwbr @foxmox

Related external issues (at least confounded by this regression):

https://bugzilla.proxmox.com/show_bug.cgi?id=4295
https://bugzilla.proxmox.com/show_bug.cgi?id=4295
https://bugzilla.kernel.org/show_bug.cgi?id=199727
...
[likely others]

PR #1153, opened 2024-09-19
[vioscsi] SendSRB minor refactor

benyamin-codez:
The `virtqueue_add_buf()` routine will return 0 on SUCCESS or otherwise a negative number, usually -28 (`ENOSPC`), i.e no space for buffer. An inline comment is added for edification.

Here we refactor `virtqueue_add_buf()` handling to only process the SRB if `virtqueue_add_buf()` returns SUCCESS. Presently, other positive return codes would be processed.

Here we define `VQ_ADD_BUFFER_SUCCESS` in `vioscsi\helper.h`.

To ensure valid data is reported we also move the trace event on failure to above the call to `CompleteRequest()` and wrap the line for improved readability.

### DIFF
--- a/vioscsi/helper.c
+++ b/vioscsi/helper.c
@@ -55,7 +55,7 @@ SendSRB(
     ULONG               status = STOR_STATUS_SUCCESS;
     UCHAR               ScsiStatus = SCSISTAT_GOOD;
     ULONG               MessageID;
-    INT                 add_buffer_req_status = VQ_ADD_BUFFER_SUCCESS;
+    INT                 res = VQ_ADD_BUFFER_SUCCESS;
     PREQUEST_LIST       element;
     ULONG               vq_req_idx;
 ENTER_FN_SRB();
@@ -104,12 +104,12 @@ ENTER_FN_SRB();
 
     VioScsiVQLock(DeviceExtension, MessageID, &LockHandle, FALSE);
     SET_VA_PA();
-    add_buffer_req_status = virtqueue_add_buf(adaptExt->vq[QueueNumber],
+    res = virtqueue_add_buf(adaptExt->vq[QueueNumber],
                                               srbExt->psgl,
                                               srbExt->out, srbExt->in,
                                               &srbExt->cmd, va, pa);
 
-    if (add_buffer_req_status == VQ_ADD_BUFFER_SUCCESS) {
+    if (res == VQ_ADD_BUFFER_SUCCESS) {
         element = &adaptExt->processing_srbs[vq_req_idx];
         InsertTailList(&element->srb_list, &srbExt->list_entry);
         element->srb_cnt++;
@@ -128,7 +128,7 @@ ENTER_FN_SRB();
         StorPortBusy(DeviceExtension, 10);
         RhelDbgPrint(TRACE_LEVEL_WARNING, 
                 " Could not put an SRB into a VQ due to error %i. To be completed with SRB_STATUS_BUSY. QueueNumber = %lu, SRB = 0x%p, Lun = %d, TimeOut = %d.\n", 
-                add_buffer_req_status, QueueNumber, srbExt->Srb, SRB_LUN(Srb), Srb->TimeOutValue);
+                res, QueueNumber, srbExt->Srb, SRB_LUN(Srb), Srb->TimeOutValue);
         CompleteRequest(DeviceExtension, Srb);
     }
 
--- a/vioscsi/helper.c
+++ b/vioscsi/helper.c
@@ -54,10 +54,10 @@ SendSRB(
     STOR_LOCK_HANDLE    LockHandle = { 0 };
     ULONG               status = STOR_STATUS_SUCCESS;
     UCHAR               ScsiStatus = SCSISTAT_GOOD;
-    ULONG MessageID;
-    int res = 0;
+    ULONG               MessageID;
+    INT                 add_buffer_req_status = VQ_ADD_BUFFER_SUCCESS;
     PREQUEST_LIST       element;
-    ULONG               index;
+    ULONG               vq_req_idx;
 ENTER_FN_SRB();
 
     if (!Srb)
@@ -93,7 +93,7 @@ ENTER_FN_SRB();
     }
 
     MessageID = QUEUE_TO_MESSAGE(QueueNumber);
-    index = QueueNumber - VIRTIO_SCSI_REQUEST_QUEUE_0;
+    vq_req_idx = QueueNumber - VIRTIO_SCSI_REQUEST_QUEUE_0;
 
     if (adaptExt->reset_in_progress) {
         RhelDbgPrint(TRACE_LEVEL_FATAL, " Reset is in progress, completing SRB 0x%p with SRB_STATUS_BUS_RESET.\n", Srb);
@@ -104,13 +104,13 @@ ENTER_FN_SRB();
 
     VioScsiVQLock(DeviceExtension, MessageID, &LockHandle, FALSE);
     SET_VA_PA();
-    res = virtqueue_add_buf(adaptExt->vq[QueueNumber],
-        srbExt->psgl,
-        srbExt->out, srbExt->in,
-        &srbExt->cmd, va, pa);
+    add_buffer_req_status = virtqueue_add_buf(adaptExt->vq[QueueNumber],
+                                              srbExt->psgl,
+                                              srbExt->out, srbExt->in,
+                                              &srbExt->cmd, va, pa);
 
-    if (res == VQ_ADD_BUFFER_SUCCESS) {
-        element = &adaptExt->processing_srbs[index];
+    if (add_buffer_req_status == VQ_ADD_BUFFER_SUCCESS) {
+        element = &adaptExt->processing_srbs[vq_req_idx];
         InsertTailList(&element->srb_list, &srbExt->list_entry);
         element->srb_cnt++;
     }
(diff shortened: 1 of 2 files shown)


vioscsi/helper.c
```diff
@@ -109,7 +109,7 @@ ENTER_FN_SRB();
         srbExt->out, srbExt->in,
```
JonKohler, line 109: On virtqueue_add_buf_split(), can you fix up the comments above the function, so that you say returns VQ_ADD_BUFFER_SUCCESS on success?

vioscsi/helper.c
```diff
@@ -109,7 +109,7 @@ ENTER_FN_SRB();
...
         &srbExt->cmd, va, pa);
 
-    if (res >= 0) {
+    if (res == VQ_ADD_BUFFER_SUCCESS) {
```
JonKohler, line 112: both the split and packed variant either return -ENOSPC or zero, which made me think - I don't think ENOSPC is technically the right thing to do here. 

I'd imagine we should actually use ENOMEM or ENOBUFS, as that describes what is happening better than ENOSPC, which is supposed to be used for disk space, no? (ref https://www.gnu.org/software/libc/manual/html_node/Error-Codes.html)

vioscsi/helper.c
```diff
@@ -120,13 +120,16 @@ ENTER_FN_SRB();
             virtqueue_notify(adaptExt->vq[QueueNumber]);
         }
     } else {
+        // virtqueue_add_buf() returned -28 (ENOSPC), i.e. no space for buffer, or some other error
```
JonKohler, line 123: See my comment above. ENOSPC means no disk space. ENOBUFS means no buffers and ENOMEM is can't alloc mem

vioscsi/helper.c
```diff
@@ -120,13 +120,16 @@ ENTER_FN_SRB();
...
         virtqueue_notify(adaptExt->vq[QueueNumber]);
         ScsiStatus = SCSISTAT_QUEUE_FULL;
         SRB_SET_SRB_STATUS(Srb, SRB_STATUS_BUSY);
         SRB_SET_SCSI_STATUS(Srb, ScsiStatus);
         StorPortBusy(DeviceExtension, 10);
+        RhelDbgPrint(TRACE_LEVEL_WARNING,
```
JonKohler, line 129: We're concerned there could be at any point any other sort of error, wouldn't we want to put "res" into this log statement?

vioscsi/helper.c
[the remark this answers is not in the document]
benyamin-codez: Nice catch..!
To `VQ_ADD_BUFFER_SUCCESS` methinks...
I'll clean up that init table too...

benyamin-codez: I could but that would be crossing over into [virtio], and then we would need to probably add `VQ_ADD_BUFFER_SUCCESS` into `virtio_ring.h` instead which on reading the head of the file might be a little fraught. I also don't see any maintainer in CODEOWNERS.

vioscsi/helper.c
benyamin-codez: Although usually applying to disk devices, ENOSPC refers to _any_ device. The mnemonic ENOSPC refers to ERROR_NO_SPACE. AFAIK, this is returned from the hypervisor without translation rather than something we set. Typically ENOMEM == ENOBUFS in GNU (per your link) and these relate to allocation of virtual memory, whereas virtio sees a _device_ responding with "I have no space for that".

vioscsi/helper.c
benyamin-codez: Yes, absolutely... Will do... 8^d

benyamin-codez:
I'll drop another commit here in coming hours but thought it worth mentioning that I'm expecting the following gets clobbered by PR #1150, so won't do anything with that...

https://github.com/virtio-win/kvm-guest-drivers-windows/blob/281bb7b3c5e216fb08449ec05cfe748462ab1ff5/vioscsi/helper.c#L116-L121

benyamin-codez:
Have pushed the additional commit...
Switching this from draft to review...

benyamin-codez:
Check will fail per:

> I'll drop another commit here in coming hours but thought it worth mentioning that I'm expecting the following gets clobbered by PR #1150, so won't do anything with that...
> 
> https://github.com/virtio-win/kvm-guest-drivers-windows/blob/281bb7b3c5e216fb08449ec05cfe748462ab1ff5/vioscsi/helper.c#L116-L121

...for undeclared identifier `res` (as expected).

Let me know if that's ok or you need something else from me.

JonKohler (commented):
LGTM at first glance, thanks for the cleanup

kostyanf14:
Hi @benyamin-codez, please fix build errors.

benyamin-codez:
@kostyanf14 

Thanks I will revert the offending changes and put them in a new PR to review after other merges...

benyamin-codez:
@JonKohler 

Jon, I thought I'd better check the modded trace works properly...

I found we do actually set `-ENOSPC` in `virtqueue_add_buf_split()` and `virtqueue_add_buf_packed()` per, e.g. :

https://github.com/virtio-win/kvm-guest-drivers-windows/blob/22d0908a28f382d02d8d3a10a0ee0631af480fdf/VirtIO/VirtIORing.c#L263

However, it returns `-1` per:

https://github.com/virtio-win/kvm-guest-drivers-windows/blob/22d0908a28f382d02d8d3a10a0ee0631af480fdf/VirtIO/osdep.h#L20

I'm still of the view that `ENOSPC` is appropriate here.

@YanVugenfirer

Per the above, do you know if there was any reason we define `ENOSPC` as 1 in `osdep.h` rather than 28...?
I see it was implemented in 2014 per 264f1b6c86f5eeca5e2c9fbd24e3de8dbd0bee1d.

In Windows, 0x1 is `ERROR_INVALID_FUNCTION` and 0x1C (28) is `ERROR_OUT_OF_PAPER`...
Given why and how we are using it, alternatives might include one of the following:
```
ERROR_ACCESS_DENIED     0x5
ERROR_NOT_ENOUGH_MEMORY 0x8
ERROR_NOT_READY         0x15 (21)
ERROR_BUSY              0xAA (170)
```
...but these would require single line changes to VirtIO ( both `VirtIORing.c` and `VirtIORing-Packed.c`).

AFAICT, these are the only files in the tree which reference `ENOSPC`, but deleting the definition from `osdep.h` will result in `undeclared identifier` compilation errors. Perhaps I should include a commit to change it to 28 in this PR (or maybe a separate PR would be best)?

An alternative might be for me to use:

```c
RhelDbgPrint(TRACE_LEVEL_WARNING, 
             " Could not put an SRB into a VQ due to error %s. To be completed with SRB_STATUS_BUSY. QueueNumber = %lu, SRB = 0x%p, Lun = %d, TimeOut = %d.\n", 
             (res == -1) ? "ENOSPC" : "UNKNOWN", QueueNumber, srbExt->Srb, SRB_LUN(Srb), Srb->TimeOutValue);

```
Any advice on how to proceed would be appreciated.

cc: @vrozenfe

YanVugenfirer:
@benyamin-codez I will take a look. But I am travelling for the KVM Forum now and it might take me some time.

JonKohler:
> @benyamin-codez I will take a look. But I am travelling for the KVM Forum now and it might take me some time. 

Glad you could make it. Travel logistics to Brno are brutal from where I live, I wanted to go this year but yikes, it's tough!

vioscsi/helper.c
benyamin-codez: Per my comment below, we do actually set `-ENOSPC`, e.g.:

https://github.com/virtio-win/kvm-guest-drivers-windows/blob/22d0908a28f382d02d8d3a10a0ee0631af480fdf/VirtIO/VirtIORing.c#L263

vrozenfe (approved):

benyamin-codez:
@vrozenfe
@YanVugenfirer 

Yan & Vadim, 

I've set this back to draft with a view to replacing it with a single commit PR.

I will rebase against master and add the changes I had reverted back in.

On the ENOSPC / SCSISTAT_QUEUE_FULL trace, I think it best to use:

```c
        RhelDbgPrint(TRACE_LEVEL_WARNING, 
                     " Could not put an SRB into a VQ due to error %s (%i). To be completed with SRB_STATUS_BUSY. QueueNumber = %lu, SRB = 0x%p, Lun = %d, TimeOut = %d.\n", 
                     (res == -ENOSPC) ? "ENOSPC" : "UNKNOWN", res, QueueNumber, srbExt->Srb, SRB_LUN(Srb), Srb->TimeOutValue);
```
Let me know if you want a PR to change the definition of `ENOSPC` in `\VirtIO\osdep.h` from 1 to 28 (or xyz) ...

cc: @JonKohler

vrozenfe:
> 
> Let me know if you want a PR to change the definition of `ENOSPC` in `\VirtIO\osdep.h` from 1 to 28 (or xyz) ...
> 

As for me, I think it would be nice to keep the same definitions as in Linux.

Best,
Vadim.

YanVugenfirer:
> > Let me know if you want a PR to change the definition of `ENOSPC` in `\VirtIO\osdep.h` from 1 to 28 (or xyz) ...
> 
> As for me, I think it would be nice to keep the same definitions as in Linux.
> 
> Best, Vadim.

@benyamin-codez I agree with Vadim. Better be consistence with Linux.

benyamin-codez:
Single commit PR is #1162.

PR for `ENOSPC` change is #1163.

Closing this one...

PR #1162, opened 2024-10-07
[vioscsi] SendSRB minor refactor

benyamin-codez:
The `virtqueue_add_buf()` routine will return 0 on `SUCCESS` or otherwise a negative number, usually -28 (`ENOSPC`), i.e no space for buffer. An inline comment is added for edification. Refactored `virtqueue_add_buf()` handling to only process the SRB if `virtqueue_add_buf()` returns `SUCCESS`. Presently, other positive return codes would be processed.

Variable renames:
`res` to `add_buffer_req_status`
`index` to `vq_req_idx`
`MessageID` to `MessageId`
 -- Also cleaned up variable init table

New definitions:
`VQ_ADD_BUFFER_SUCCESS` is defined in `vioscsi\helper.h` header file.
 -- Used to initialize `add_buffer_req_status` variable. (Credit: @JonKohler)

To ensure valid data is reported we also move the trace event on failure to above the call to `CompleteRequest()` and wrap the line for improved readability.
 -- Add `add_buffer_req_status` content to trace output on error (Credit: @JonKohler)

### DIFF
--- a/vioscsi/helper.h
+++ b/vioscsi/helper.h
@@ -49,6 +49,8 @@
 #define CACHE_LINE_SIZE 64
 #define ROUND_TO_CACHE_LINES(Size)  (((ULONG_PTR)(Size) + CACHE_LINE_SIZE - 1) & ~(CACHE_LINE_SIZE - 1))
 
+#define VQ_ADD_BUFFER_SUCCESS 0
+
 #include <srbhelper.h>
 
 // Note: SrbGetCdbLength is defined in srbhelper.h
--- a/vioscsi/helper.h
+++ b/vioscsi/helper.h
@@ -40,6 +40,7 @@
 
 #include "osdep.h"
 #include "virtio_pci.h"
+#include "virtio_ring.h"
 #include "vioscsi.h"
 
 #define CHECKBIT(value, nbit) virtio_is_feature_enabled(value, nbit)
@@ -48,8 +49,6 @@
 
 #define CACHE_LINE_SIZE 64
 #define ROUND_TO_CACHE_LINES(Size)  (((ULONG_PTR)(Size) + CACHE_LINE_SIZE - 1) & ~(CACHE_LINE_SIZE - 1))
-
-#define VQ_ADD_BUFFER_SUCCESS 0
 
 #include <srbhelper.h>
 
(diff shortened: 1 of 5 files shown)


vioscsi/helper.h
```diff
@@ -49,6 +49,8 @@
 #define CACHE_LINE_SIZE 64
 #define ROUND_TO_CACHE_LINES(Size)  (((ULONG_PTR)(Size) + CACHE_LINE_SIZE - 1) & ~(CACHE_LINE_SIZE - 1))
 
+#define VQ_ADD_BUFFER_SUCCESS 0
```
JonKohler, line 52: Can you have this be the value that gets returned too? That would help glue this all together, so if you did a git grep in the future you would see it super crisply

JonKohler (changes requested):
One small nit: use the new definition on the return path too, and it will glue it all together

vioscsi/helper.h
benyamin-codez: I was trying to avoid touching anything VirtIO (in this case  in `VirtIO\VirtIORing.c` and `VirtIO\VirtIORing-Packed.c`)... Is that the MO for the repo? It seemed to me we avoid making changes there if possible.

JonKohler: I guess if I was splitting hairs, I'd have this in two commits. First being adding that to virtio common code in general, and a second commit to plumb into scsi code. Just my 2 cents

benyamin-codez: Thanks Jon. I've added a second commit to add the definition to VirtIO instead and updated the relevant VirtIO routines to use the new definition.

YanVugenfirer (approved):

YanVugenfirer:
@kostyanf14 Can you take a look please on the build error:

```
       (ApiValidator target) -> 
         C:\EWDK11\Program Files\Windows Kits\10\build\WindowsDriver.common.targets(1812,5): error MSB4175: The task factory "XamlTaskFactory" could not be loaded from the assembly "C:\EWDK11\Program Files\Microsoft Visual Studio\2019\BuildTools\MSBuild\Current\Bin\Microsoft.Build.Tasks.Core.dll". Could not find file 'C:\Windows\TEMP\0ovbab01\0ovbab01.dll'. [C:\workspace\VirtIO-EWDK-11-24H2-SDV\viogpu\viogpusc\viogpusc.vcxproj]


    0 Warning(s)
    1 Error(s)

```

kostyanf14:
rerun tests

benyamin-codez:
@kostyanf14
Is there something up with the checks?
I'm not seeing those failures in my tests...

vrozenfe (approved):

PR #1163, opened 2024-10-07
[VirtIO] Redefine ENOSPC to common Linux errno

benyamin-codez:
The `VirtIO\osdep.h` header file defines `ENOSPC` as 1. On Windows this equates to `ERROR_INVALID_FUNCTION`.

This commit redefines `ENOSPC` to use the more typical Linux definition of 28.

### DIFF
--- a/VirtIO/osdep.h
+++ b/VirtIO/osdep.h
@@ -17,7 +17,7 @@
 #include <ntddk.h>
 
 #if !defined(ENOSPC)
-#define ENOSPC 1
+#define ENOSPC 28
 #endif
 
 #if !defined(__cplusplus) && !defined(bool)


benyamin-codez:
It's worth mentioning that AFAICT the only place in the tree that uses this definition of `ENOSPC` are the `virtqueue_add_buf()` routines in `VirtIO\VirtIORing.c` and `VirtIO\VirtIORing-Packed.c`.

JonKohler (commented):
LGTM

YanVugenfirer:
> It's worth mentioning that AFAICT the only place in the tree that uses this definition of `ENOSPC` are the `virtqueue_add_buf()` routines in `VirtIO\VirtIORing.c` and `VirtIO\VirtIORing-Packed.c`.

Yes, that's how it looks like.

YanVugenfirer (approved):
Thanks!

PR #1165, opened 2024-10-09
[viomem][https://issues.redhat.com/browse/RHEL-35647] Fixed BSOD, whi…

marekkedzierski:
…ch sometimes appeared when the user tried to shrink memory. An array holding memory ranges was overwritten.

### DIFF
--- a/viomem/sys/viomem.c
+++ b/viomem/sys/viomem.c
@@ -628,6 +628,66 @@ LONGLONG FindConsecutivePagesCountMDL(PPFN_NUMBER Pages, LONGLONG Remaining)
 }
 
 //
+// Function calculates and returns count of memory ranges from MDL.
+//
+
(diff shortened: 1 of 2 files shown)


viomem/sys/viomem.c
```diff
@@ -627,6 +627,66 @@ LONGLONG FindConsecutivePagesCountMDL(PPFN_NUMBER Pages, LONGLONG Remaining)
 	return index;
 }
 
+//
+// Function calculates and returns count of memory ranges from MDL.
+//
+
+ULONG GetMemoryRangesCountFromMdl(PMDL Mdl)
+{
+	PPFN_NUMBER pfnNumbers = 0;
```
YanVugenfirer, line 636: I think this name is very confusing. It looks like it is going to hold the amount of PFNs and not the pointer to the array.

YanVugenfirer (approved):

PR #1170, opened 2024-10-15
[viomem][https://issues.redhat.com/browse/RHEL-31880] Optimized STATE request processing.

marekkedzierski0:
PR created after applying changes after code review:
- removed the old definition of the function, which was kept in the source 
  file surrounded with #if #else #end)
- added a sanity check to limit the number of blocks that can be processed 
  (member virtio_mem_req_state::nb_blocks is a type of DWORD, so the limit is 
  up to 65535 blocks); if the number of blocks is greater than 65535, the function fails)

### DIFF
--- a/viomem/sys/viomem.c
+++ b/viomem/sys/viomem.c
@@ -1567,18 +1567,99 @@ BOOLEAN SendStateRequest(
 //		   during system resets". So this code is for hypothetical scenario 
 //		   because after reset plugged_size will be always set to zero.
 //
+// The optimization is based on replacing simple loop iteration with a query 
+// and processing a set of blocks at once. The performance depends on block 
+// distribution in a given range, i.e., 'holes' between blocks.
+//
+// It can be described with simplified (recursive) code:
+//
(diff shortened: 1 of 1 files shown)


viomem/sys/viomem.c
```diff
@@ -1620,15 +1723,137 @@ BOOLEAN SynchronizeDeviceAndDriverMemory(IN WDFOBJECT Device,
...
+					//
+					// It should not happen because a device will always respond.
+					//
+
+					return FALSE;
+				}
+			}
+		}
+		if (st)
+			ExFreePoolWithTag(st, VIRTIO_MEM_POOL_TAG);
```
YanVugenfirer, line 1851: I suggest to use "{" also with single line "if".

YanVugenfirer: I suggest doing it as part of the code-style refactoring task.

YanVugenfirer:
Waiting for CI to merge

YanVugenfirer (approved):

PR #1175, opened 2024-10-23
[vioscsi] Reduce spinlock management complexity

benyamin-codez:
Spinlocks are now primarily of the DpcLock type, without which the driver will not properly operate. Fallback to an InterruptLock type spinlock is limited, perhaps even cosmetic.

This commit removes the VioScsiVQLock() and VioScsiVQUnlock() wrapper routines to use the native functions instead, i.e. StorPortAcquireSpinLock() and StorPortReleaseSpinLock(). VioScsiVQLock() contained a superfluous MSI-X and QueueNumber check and VioScsiVQUnlock() did not release non-DpcLock type spinlocks. Removing the wrapper routines also resolves these two issues.

Refactoring includes:
* Renaming of ProcessQueue() to ProcessBuffer(). Whilst technically more accurate, this also provides space for a new ProcessQueue() routine, via which the HW_MESSAGE_SIGNALLED_INTERRUPT_ROUTINE and HW_INTERRUPT routines can then be consolidated.
* Where necessary, changing BOOLEAN isr to STOR_SPINLOCK LockMode
* CamelCase / conformity correction of MessageID to MessageId
* Conformity correction of QueuNum to QueneNumber
* Removal of unnecessary use of MsgID and MessageID or MessageId in area of focus (11 instances of MessageID remain in other vioscsi.c routines)
* Mnemonic rename of:
  * CompletePendingRequests() to CompletePendingRequestsOnReset()
  * index to vq_req_idx
* Syntactic correction of if statement (added braces)

### DIFF
--- a/vioscsi/helper.c
+++ b/vioscsi/helper.c
@@ -53,9 +53,9 @@ SendSRB(
     ULONG               QueueNumber = VIRTIO_SCSI_REQUEST_QUEUE_0;
     BOOLEAN             notify = FALSE;
     STOR_LOCK_HANDLE    LockHandle = { 0 };
+    PVOID               LockContext;
     ULONG               status = STOR_STATUS_SUCCESS;
     UCHAR               ScsiStatus = SCSISTAT_GOOD;
-    ULONG               MessageId;
     INT                 add_buffer_req_status = VQ_ADD_BUFFER_SUCCESS;
     PREQUEST_LIST       element;
     ULONG               vq_req_idx;
@@ -94,8 +94,8 @@ ENTER_FN_SRB();
         return;
     }
 
-    MessageId = QUEUE_TO_MESSAGE(QueueNumber);
     vq_req_idx = QueueNumber - VIRTIO_SCSI_REQUEST_QUEUE_0;
+    LockContext = &adaptExt->dpc[vq_req_idx];
 
     if (adaptExt->reset_in_progress) {
         RhelDbgPrint(TRACE_LEVEL_FATAL, " Reset is in progress, completing SRB 0x%p with SRB_STATUS_BUS_RESET.\n", Srb);
@@ -104,7 +104,7 @@ ENTER_FN_SRB();
         return;
     }
 
-    VioScsiVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
+    StorPortAcquireSpinLock(DeviceExtension, DpcLock, LockContext, &LockHandle);
     SET_VA_PA();
     add_buffer_req_status = virtqueue_add_buf(adaptExt->vq[QueueNumber],
                                               srbExt->psgl,
@@ -127,7 +127,7 @@ ENTER_FN_SRB();
                      (add_buffer_req_status == -ENOSPC) ? "ENOSPC" : "UNKNOWN", add_buffer_req_status, QueueNumber, srbExt->Srb, SRB_LUN(Srb), Srb->TimeOutValue);
         CompleteRequest(DeviceExtension, Srb);
     }
-    VioScsiVQUnlock(DeviceExtension, MessageId, &LockHandle, FALSE);
+    StorPortReleaseSpinLock(DeviceExtension, &LockHandle);
     if (notify){
         virtqueue_notify(adaptExt->vq[QueueNumber]);
     }
@@ -495,51 +495,6 @@ ENTER_FN();
     EventNode->sg.physAddr = StorPortGetPhysicalAddress(DeviceExtension, NULL, &EventNode->event, &fragLen);
     EventNode->sg.length   = sizeof(VirtIOSCSIEvent);
     return SynchronizedKickEventRoutine(DeviceExtension, (PVOID)EventNode);
-EXIT_FN();
-}
-
-VOID
-//FORCEINLINE
-VioScsiVQLock(
-    IN PVOID DeviceExtension,
-    IN ULONG MessageID,
-    IN OUT PSTOR_LOCK_HANDLE LockHandle,
-    IN BOOLEAN isr
-    )
-{
-    PADAPTER_EXTENSION  adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
-    ULONG               QueueNumber = MESSAGE_TO_QUEUE(MessageID);
-    ENTER_FN();
-
-    if (!isr) {
-        if (adaptExt->msix_enabled) {
-            // Queue numbers start at 0, message ids at 1.
-            NT_ASSERT(MessageID > VIRTIO_SCSI_REQUEST_QUEUE_0);
-            if (QueueNumber >= (adaptExt->num_queues + VIRTIO_SCSI_REQUEST_QUEUE_0)) {
-                QueueNumber %= adaptExt->num_queues;
-            }
-            StorPortAcquireSpinLock(DeviceExtension, DpcLock, &adaptExt->dpc[QueueNumber - VIRTIO_SCSI_REQUEST_QUEUE_0], LockHandle);
-        }
-        else {
-            StorPortAcquireSpinLock(DeviceExtension, InterruptLock, NULL, LockHandle);
-        }
-    }
-EXIT_FN();
-}
-
-VOID
-//FORCEINLINE
-VioScsiVQUnlock(
-    IN PVOID DeviceExtension,
-    IN ULONG MessageID,
-    IN PSTOR_LOCK_HANDLE LockHandle,
-    IN BOOLEAN isr
-    )
-{
-ENTER_FN();
-    if (!isr) {
-        StorPortReleaseSpinLock(DeviceExtension, LockHandle);
-    }
 EXIT_FN();
 }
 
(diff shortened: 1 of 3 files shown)


JonKohler (commented):
LGTM at first blush, will defer to Vadim for a deeper review

benyamin-codez:
After exploring various blends of different spinlock types, I was minded to remove the spinlock wrappers entirely.

This approach also has a minor performance improvement.

For reference:

The full spinlock manager I used: https://github.com/benyamin-codez/kvm-guest-drivers-windows/blob/bc85387690c04baaf21adad52b2ee827c3f740f3/vioscsi/helper.c#L853-L956
An essential-only version: https://github.com/benyamin-codez/kvm-guest-drivers-windows/blob/bc85387690c04baaf21adad52b2ee827c3f740f3/vioscsi/helper.c#L806-L851

vrozenfe:
@benyamin-codez 

Thanks a lot.
Can you please check it briefly on the following configurations:
- num_queues=1 (from the qemu command line)
- and  MSISupported Registry key set to "0" 
Both these configurations are quite rare nowadays and mostly mean the HW malfunction. But we still need to 
make sure that the driver is functional if the underling virtio HBA is a kind of "broken".

Best regards,
Vadim.

benyamin-codez:
@vrozenfe 
> Can you please check it briefly on the following configurations:
> 
>     * num_queues=1 (from the qemu command line)
>     * and  MSISupported Registry key set to "0"
Thanks Vadim.

I am pleased to confirm this commit works both with and without MSI-X using the registry control method via the `MSISupported` value in the `ENUM\PCI` registry tree.

I have been explicitly testing `num_queues=1` by using the `vCPU=1` method. I am pleased to confirm this method also works whether MSI-X is enabled or disabled. I trust this is sufficient.

I also tested `MSISupported=0` with `vCPU>1` which results in `num_queues=1` and successful operation.

Regarding my comments about calls to `InterruptLock` type spinlocks being largely cosmetic, I should note I'm not sure how this commit or master would work when `!adaptExt->dpc_ok`... My understanding is that it will fail at `HW_INITIALIZE`, per: https://github.com/virtio-win/kvm-guest-drivers-windows/blob/162fe67bf4837045f4715a81fb5fd95a61801203/vioscsi/vioscsi.c#L854-L856 As such, should we remove the calls to the remaining `InterruptLock` type spinlocks?

If so, I would propose doing so in a different PR, perhaps following consolidation of the `HW_INTERRUPT` and `HW_MESSAGE_SIGNALLED_INTERRUPT_ROUTINE` codepaths (which I intended on raising after this PR was merged). The reason for this is that `VioScsiInterrupt()` currently contains a branch to call an `InterruptLock` type spinlock. The consolidation then relies on the sole remaining `InterruptLock` type spinlock in `DispatchQueue()`.

The natural progression of such refactoring would be to remove conditional checks for `adaptExt->dpc_ok`. Perhaps this could form part of an additional PR to remove the remaining `InterruptLock` type spinlock...?

Please let me know your thoughts on this and if you have any other concerns.

Best regards,
Ben

cc: @YanVugenfirer

vrozenfe:
> I am pleased to confirm this commit works both with and without MSI-X using the registry control method via the `MSISupported` value in the `ENUM\PCI` registry tree.
> 
> I have been explicitly testing `num_queues=1` by using the `vCPU=1` method. I am pleased to confirm this method also works whether MSI-X is enabled or disabled. I trust this is sufficient.

Thanks, 
we also need to test  vCPU > 1 and num_queues = 1 configuration because in this case we can expect ISR and DPC to be executed on different CPUs simultaneously.  

> 
> I also tested `MSISupported=0` with `vCPU>1` which results in `num_queues=1` and successful operation.
> 
> Regarding my comments about calls to `InterruptLock` type spinlocks being largely cosmetic, I should note I'm not sure how this commit or master would work when `!adaptExt->dpc_ok`... My understanding is that it will fail at `HW_INITIALIZE`, per:
> 
> https://github.com/virtio-win/kvm-guest-drivers-windows/blob/162fe67bf4837045f4715a81fb5fd95a61801203/vioscsi/vioscsi.c#L854-L856
> 
that !adaptExt->dpc_ok condition is a kind of tricky one. IIRC, we hit this case when resuming from Suspend and/or Hibernation 

Best,
Vadim.

benyamin-codez:
@vrozenfe 
> Thanks, we also need to test vCPU > 1 and num_queues = 1 configuration because in this case we can expect ISR and DPC to be executed on different CPUs simultaneously.

We know that when `MSISupported=0` that `num_queues=1` via:
https://github.com/virtio-win/kvm-guest-drivers-windows/blob/162fe67bf4837045f4715a81fb5fd95a61801203/vioscsi/vioscsi.c#L529-L532 So using `MSISupported=1` and `num_queues=1` via QEMU command line with `vCPU>1` results in `VioScsiInterrupt()` only issuing DPCs whilst `adaptExt->dpc_ok=TRUE` via:
https://github.com/virtio-win/kvm-guest-drivers-windows/blob/162fe67bf4837045f4715a81fb5fd95a61801203/vioscsi/vioscsi.c#L1057-L1067 The `InterruptLock` type spinlock (which I have not tested) would only be used in the `else` branch, i.e. when `IsCrashDumpMode=TRUE` or when `adaptExt->dpc_ok=FALSE` (or `!adaptExt->dpc_ok`).

In any case - to answer your question - this configuration also tests ok with this commit.
> that !adaptExt->dpc_ok condition is a kind of tricky one. IIRC, we hit this case when resuming from Suspend and/or Hibernation

Very tricky indeed from what I can tell..! 8^d

In fact, I cannot see how this would work when all other spinlocks are the `DpcLock` type. In my previous testing, issuing `InterruptLock` type spinlocks when `adaptExt->dpc_ok=TRUE` usually clobbered in-flight DPCs.

I think this is worth exploring, but in a future PR as previously suggested.

Best regards,
Ben

cc: @YanVugenfirer

kostyanf14:
rerun tests

benyamin-codez:
@vrozenfe, @YanVugenfirer 

>>Thanks, we also need to test vCPU > 1 and num_queues = 1 configuration because in this case we can expect ISR and DPC to be executed on different CPUs simultaneously.
>
>In any case - to answer your question - this configuration also tests ok with this commit.

If I'm not mistaken, this one should be good to go too, per https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1181#issuecomment-2466981842.

Perhaps it's worthwhile running one more CI retest?

Best regards,
Ben

cc: @kostyanf14

benyamin-codez:
@kostyanf14 

> Perhaps it's worthwhile running one more CI retest?

What do you think? Or is this one right to go too?

kostyanf14:
rerun tests

kostyanf14:
@benyamin-codez 

`NO-SDV` means we skip DVL.XML generation, so `Static Tools Logo Test` will fails. This is expected. 
This was done just to reduce the build time.

benyamin-codez:
@kostyanf14 

> `NO-SDV` means we skip DVL.XML generation, so `Static Tools Logo Test` will fails. This is expected.
This was done just to reduce the build time.

Yep, that was pretty clear from the Jenkins config you shared the other day.

I was more concerned with the `Flush Test` and `Disk Verification (LOGO)` on Server2022 and the latter on Server2025.
I could only find the same 0x80070002 errors in the logs, as found in https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1181#issuecomment-2466981842.

In that PR, you stated it was safe to proceed. However, this PR is much more consequential, and having those failures is a concern. Are the failures a known issue which can be disregarded? Or are you able to provide some insight as to why they are occurring, e.g. are there BSODs?

With kind regards,
Ben

YanVugenfirer:
@vrozenfe is it OK to merge?

vrozenfe (approved):
Looks good.
Thank you.
Vadim.

YanVugenfirer:
@benyamin-codez sorry, can you rebase please?

kostyanf14:
@benyamin-codez I stopped CI, as PR is blocked from merging by GitHub anyway.

benyamin-codez:
@YanVugenfirer @kostyanf14 

Will that work, or have I cooked it..?

It certainly seems rebased now, but I see it still says `This branch cannot be rebased safely`...

FYI, I used:
```
git checkout master 
git pull
git checkout SpinLocks-minimalist 
git rebase master
git pull --rebase
git push
```
Is that right, or should I have used another technique?

Let me know if you need a fresh PR instead.

benyamin-codez:
@YanVugenfirer @kostyanf14 

This doesn't look right to me at all...
Please advise alternate technique or should I just do a fresh PR against master?

benyamin-codez:
@vrozenfe @YanVugenfirer @kostyanf14 

I'm still getting `This branch cannot be rebased safely` following another rebase...

Here is a fresh PR against master: #1196

benyamin-codez:
Binary equivalent PR #1196 is reporting `This branch has no conflicts with the base branch when rebasing`...
Therefore, I will close this PR...

PR #1206, opened 2024-12-01
RHEL-68725: netkvm: Several fixes to prevent problems under driver verifier

ybendito:


YanVugenfirer (approved):
Approved. But I suggest to change commit messages to comply with Jira automation.

PR #1215, opened 2024-12-07
[vioscsi] DriverEntry() improved tracing

benyamin-codez:
Show the _RegistryPath_.
Show the _CrashDump Mode_.
Show `StorPortInitialize()` return value (including LONG).
Show `NTDDI_VERSION`.
Also added `hwInitData.AddressTypeFlags`.
Added missing `NTDDI` definitions in `vioscsi.h`.

### DIFF
--- a/vioscsi/vioscsi.c
+++ b/vioscsi/vioscsi.c
@@ -350,6 +350,8 @@ DriverEntry(
 
     HW_INITIALIZATION_DATA hwInitData;
     ULONG                  initResult;
+    ANSI_STRING            aRegistryPath;
+    NTSTATUS               u2a_status;
 
 #ifdef EVENT_TRACING
     STORAGE_TRACE_INIT_INFO initInfo;
@@ -358,10 +360,20 @@ DriverEntry(
     InitializeDebugPrints((PDRIVER_OBJECT)DriverObject, (PUNICODE_STRING)RegistryPath);
 
     IsCrashDumpMode = FALSE;
-    RhelDbgPrint(TRACE_LEVEL_FATAL, " Vioscsi driver started...built on %s %s\n", __DATE__, __TIME__);
+    #if !defined(RUN_UNCHECKED) || defined(RUN_MIN_CHECKED)
+    RhelDbgPrint(TRACE_LEVEL_FATAL, " VIOSCSI driver started...built on %s %s\n", __DATE__, __TIME__);
+    #endif
     if (RegistryPath == NULL) {
         IsCrashDumpMode = TRUE;
+        #if !defined(RUN_UNCHECKED)
+        memset(&aRegistryPath, 0, sizeof(aRegistryPath));
+        u2a_status = RtlUnicodeStringToAnsiString(&aRegistryPath, RegistryPath, TRUE);
+        if (u2a_status == STATUS_SUCCESS) {
+            RhelDbgPrint(TRACE_LEVEL_VERBOSE, " RegistryPath : %s \n", aRegistryPath.Buffer);
+            RtlFreeAnsiString(&aRegistryPath);
+        }
         RhelDbgPrint(TRACE_LEVEL_INFORMATION, " Crash dump mode\n");
+        #endif
     }
 
     RtlZeroMemory(&hwInitData, sizeof(HW_INITIALIZATION_DATA));
@@ -394,7 +406,8 @@ DriverEntry(
     hwInitData.NumberOfAccessRanges     = PCI_TYPE0_ADDRESSES;
     hwInitData.MapBuffers               = STOR_MAP_NON_READ_WRITE_BUFFERS;
 
-    hwInitData.SrbTypeFlags = SRB_TYPE_FLAG_STORAGE_REQUEST_BLOCK;
+    hwInitData.SrbTypeFlags             = SRB_TYPE_FLAG_STORAGE_REQUEST_BLOCK;
+    hwInitData.AddressTypeFlags         = ADDRESS_TYPE_FLAG_BTL8;
 
     initResult = StorPortInitialize(DriverObject,
                                     RegistryPath,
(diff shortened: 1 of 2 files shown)


vioscsi/vioscsi.h
```diff
@@ -46,9 +46,26 @@ typedef struct VirtIOBufferDescriptor VIO_SG, *PVIO_SG;
...
+#define NTDDI_WINTHRESHOLD      0x0A000000  /* ABRACADABRA_THRESHOLD */
 #endif
 
+#ifndef NTDDI_WIN10_NI
+#define NTDDI_WIN10_NI          0x0A00000C
+#define NTDDI_WIN10_CU          0x0A00000D
+#endif
+
+#ifndef NTDDI_WIN11
+#define NTDDI_WIN11             NTDDI_WIN10_CO
```
YanVugenfirer, line 58: I think OS version defines should go to a separate file or to osdep.h

vioscsi/vioscsi.c
```diff
@@ -358,10 +360,20 @@ DriverEntry(
     InitializeDebugPrints((PDRIVER_OBJECT)DriverObject, (PUNICODE_STRING)RegistryPath);
 
     IsCrashDumpMode = FALSE;
-    RhelDbgPrint(TRACE_LEVEL_FATAL, " Vioscsi driver started...built on %s %s\n", __DATE__, __TIME__);
+    #if !defined(RUN_UNCHECKED) || defined(RUN_MIN_CHECKED)
```
YanVugenfirer, line 363: Why are the printouts conditional? This is not hot path, print always.

vioscsi/vioscsi.c
```diff
@@ -394,7 +406,8 @@ DriverEntry(
     hwInitData.NumberOfAccessRanges     = PCI_TYPE0_ADDRESSES;
     hwInitData.MapBuffers               = STOR_MAP_NON_READ_WRITE_BUFFERS;
 
-    hwInitData.SrbTypeFlags = SRB_TYPE_FLAG_STORAGE_REQUEST_BLOCK;
+    hwInitData.SrbTypeFlags             = SRB_TYPE_FLAG_STORAGE_REQUEST_BLOCK;
+    hwInitData.AddressTypeFlags         = ADDRESS_TYPE_FLAG_BTL8;
```
YanVugenfirer, line 410: Not related to trace.

benyamin-codez:
@YanVugenfirer 

Many thanks for the review, Yan...!

Unfortunately I've run out of time for now...
I'll come back to this one after pushing the new commits for the build stuff.
Hopefully before too long.

Best regards,
Ben

vioscsi/vioscsi.c
benyamin-codez: No.
IIRC, this was introduced in Win8.1.
I would have added this to my WIP when I added all missing init options. Would you prefer it in a separate PR?

vioscsi/vioscsi.c
benyamin-codez: @YanVugenfirer 
I have started a discussion related to this over [here](https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1214#issuecomment-2543969865).

vioscsi/vioscsi.h
benyamin-codez: @YanVugenfirer 
Can do, but it's worth mentioning this is historically where `ABRACADABRA_THRESHOLD` has been defined.
If you prefer, I can implement it in, say `ntddi_ver.h`...
Let me know.

vioscsi/vioscsi.c
YanVugenfirer: Yes, please.

benyamin-codez: Done..!

vioscsi/vioscsi.c
benyamin-codez: I'll be updating this per PR #1228, but note this particular event is never logged with ETW as it hasn't loaded yet, i.e. this is for DBG-enabled builds only.

vioscsi/vioscsi.h
benyamin-codez: Will extract all NTDDI version definitions to `ntddi_ver.h`...

benyamin-codez:
@YanVugenfirer @vrozenfe @kostyanf14 

This one should be right to go again.
Did you want to re-run any checks before switching from draft to _Ready for review_...?

I've fixed the problem breaking compilation (my bad), so the `HCK-CI [EWDK]` check should be right to go again. 8^d
The build completes a full SDV build for all platforms and all targets without errors...
... so should be good-to-go for other `HCK-CI/vioscsi` checks too.

All appears to work as expected in both Win10 and Win11.

benyamin-codez:
@kostyanf14 @YanVugenfirer

I believe those check failures are all known issues, yes...?

YanVugenfirer:
@benyamin-codez [Disk Verification (LOGO)](https://docs.microsoft.com/en-us/windows-hardware/test/hlk/testref/f8f88b8a-ff9c-4cfb-9e95-6be2511e5510) and Flush tests failures are known issue.

vioscsi/vioscsi.c
```diff
@@ -418,11 +431,78 @@ DriverEntry(
     }
 #endif
 
-    RhelDbgPrint(TRACE_LEVEL_VERBOSE,
-                 " Initialize returned 0x%x\n", initResult);
-
+    #if !defined(RUN_UNCHECKED) || defined(RUN_MIN_CHECKED)
+    RhelDbgPrint(TRACE_LEVEL_NONE, " VIOSCSI driver starting...");
+    RhelDbgPrint(TRACE_LEVEL_NONE, " Built on %s at %s \n", __DATE__, __TIME__);
```
YanVugenfirer, line 435: Again, why not to print this alway?
This is not a "performance path," and it is degrading existing behavior.

YanVugenfirer:
Also, please squash the commits.

vioscsi/vioscsi.c
benyamin-codez: The mechanism provides a means to run without WPP which can result in an approx. 80-100 &mu;s latency improvement and an increase of 20-50k IOPS for 4KiB random I/O (with 32 queues, 16 threads on 4KiB segments). When WPP is turned off, these macros stop compile-time errors when building.

The mechanism is also quite useful during development to help determine the overhead of various tracing changes.

vioscsi/vioscsi.c
```diff
@@ -418,11 +431,78 @@ DriverEntry(
...
+        RtlFreeAnsiString(&aRegistryPath);
+    }
+    RhelDbgPrint(TRACE_LEVEL_INFORMATION, " Crash dump mode : %s \n", (IsCrashDumpMode) ? "ACTIVATED" : "NOT ACTIVATED");
+    RhelDbgPrint(TRACE_LEVEL_VERBOSE, " StorPortInitialize() returned : 0x%x (%lu) \n", initResult, initResult);    
+    #endif
+
+    #if !defined(RUN_UNCHECKED)    
+    switch (NTDDI_VERSION) {
+        case NTDDI_WIN10:
+            RhelDbgPrint(TRACE_LEVEL_VERBOSE, " NTDDI_VERSION : ABRACADABRA_THRESHOLD | Windows 10.0.10240 | 1507 | Threshold 1 \n");
```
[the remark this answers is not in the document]
benyamin-codez: > Move the switch/case to a separate function (...) to keep DriverEntry cleaner.

Done...!
8^d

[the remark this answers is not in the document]
benyamin-codez: I think this one is now resolved...!?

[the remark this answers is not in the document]
benyamin-codez: @YanVugenfirer @vrozenfe 

> @vrozenfe, what do you think?
>
> Even if we go for this:
> 1. Remove "ABRACADABRA_ :)
> 2. Move the switch/case to a separate function (compile under pre-processor conditions) to keep DriverEntry cleaner.

I've addressed both of these. Was there anything left to resolve here?
Perhaps we are just waiting for your view, Vadim?

vioscsi/vioscsi.c
benyamin-codez: One can also use `RUN_UNCHECKED` to completely exclude tracing as a cause of latency when debugging performance optimisations and other code changes unrelated to tracing.

After a few minutes, I find myself just looking past / through the compile-time macros. The lack of indent helps the brain ignore them when scanning the code base. Initially it is perhaps a little aesthetically displeasing, but then one adjusts and barely notices.

benyamin-codez:
Ok...
So I've pushed a mod to the start of the function to ensure `IsCrashDumpMode` is set before doing any tracing.
If we are tracing, then we skip registry read if `IsCrashDumpMode` is TRUE, as it will be `NULL`.

YanVugenfirer:

PR #1212, opened 2024-12-05
[build] Fix and extend build scripting

benyamin-codez:
1. Fixed various syntax issues, e.g. unquoted variables
2. Unwrapped nested `IF` statements to ensure no jumping out of loops
3. Extended colour pallette to `stdout`
4. Introduced environment variable `SKIP_SDV_ACTUAL` to skip _Static Driver Verifier (SDV)_ during analysis builds. _CodeQL, Code Analysis (CA)_ and _DVL_ operations are still executed. Default is to perform the SDV build (unchanged behaviour).
5. Make actual SDV build conditional (`Win10_SDV` only) f)
6. Introduced managed CodeQL package cache and test suite versioning control. If the CodeQL binary is detected at `CODEQL_BIN` a CodeQL build commences, otherwise it is skipped (unchanged behaviour). The package cache versions are set in the `:config_ql_whcp` function (new). Test suite versioning is via git hash of the WDK Developer Supplemental Tools repository (new, depends on `git`). The versions are determined by the value of the `WHCP_LEVEL` variable, which can be either `WHCP_LEGACY` or `WHCP_24H2`. Provision for future WHCP versions is templated as `WHCP_NEXT`.
7. Introduced environment variable `CODEQL_OFFLINE_ONLY` to perform package cache and suite version checking but NOT download any needed updates. Exits on error expecting resolution. Requires `git`.
8. Introduced environment variable `CODEQL_RUN_BLIND` to retain legacy behaviour and perform no package cache or suite version checking. It presumes the prerequisites are present. Removes `git` dependency.
9. Pre-builds x86 `viosock` libraries when building `virtio-win.sln` or `viosock.sln` for `amd64` in the circumstances the required x86 libraries do not already exist.
10. If running in a Germanium EWDK environment, the script will instantiate a Cobalt EWDK environment to build Win10 SDV operations, and will return to Germanium EWDK when done.
11. Use Germanium EWDK to sign Win11 drivers.
12. Enforce `echo off` after calling `SetupBuildEnv.cmd`
13. Provide some extra EWDK details when `EnterpriseWDK=True`
14. Prettify last leg of build and provide SUCCESS / FAIL feedback.

PR #1232, opened 2024-12-24
[build] Use Germanium EWDK to sign Win11 drivers

benyamin-codez:
1. Use the Germanium EWDK Build Lab environment to sign Win11 drivers.
    We currently use the Cobalt EWDK to sign both Win10 and Win11 drivers.

Split from PR #1212.

### DIFF
--- a/build/signAll.bat
+++ b/build/signAll.bat
@@ -9,7 +9,7 @@ endlocal
 
 setlocal
 echo Loading Windows 11 build env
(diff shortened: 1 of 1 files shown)


kostyanf14 (approved):

YanVugenfirer (commented):
Let's stop using yet another MS name conversion and keep YeahHalf conversion for now (Win11 24H2 as example).

PR #1231, opened 2024-12-24
[build] Basic fixes and extentions to build scripting

benyamin-codez:
1. Fixed various syntax issues, including an unquoted variable in conditional test (Debug or Release log file name), superfluous backslashes, spaces between errorlevel tests and tested functions, MSBuild syntax consolidation, prefixing FOR do commands with @, etc.
 2. Unwrapped nested IF statements to ensure no jumping out of loops or other unexpected behaviours.
 3. Updated usage, including Win11 target and explanation regarding analysis builds (build.bat only).
 4. Added analysis build specific variables CODEQL_FAILED and SDV_FAILED to init area (build.bat only).
 5. Refactored fail code, removing return failure detection in favour of in-line failure detection.
 6. Updated labels to use snake style notation.
 7. Refactored end of build_arch function to avoide "goto :eof" and instead use ":build_arch_done" to ensure POPD occurs and ":build_arch_skip" to check for success before ending the local environment (build.bat only).
 8. Do conditional (if exists) removal of sdv and codeql_db folders (build.bat only).
 9. Additional commenting and stdout output (logging).
10. Remove superfluous newline spacing.
11. Removed packaging folder from build per PRs [#1209](https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1209) and [#1224](https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1224) (buildAll.bat only).
12. Consolidated exiting function (whether success or fail) to :leave (buildAll.bat only).

Split from PR #1212.
```
     .,ueeeeeu,
   e$$$$$P?$$$$$b
 :$$$$"      '$$$$c
 $$$$          $$$$c
 $$$6          $$$$$
 $$$F          $$$$"
 $$$F          $$$F
 $$$$         $$$F
 $$$$        J$$F                                   .ccCCCCCCCCc,.
 ?$$$        $$$$$$e                             ,CCCCCCCCCCCCCCCCCCc
 :$$$        `$$$$$"                            CCCCCCCCCCCCCC" CCCCCC.
 `$$$.                                         CCCCCCCCCCC" .,d.`CCCCCC.
  $$$t             .,uuuuu,.                  <CCC  . `"",$$$$$$L CCCCCC
  ?$$$          ,HHHHHHHHHHHHHb,               CCC $$$$$$$$$P"""" `CCCCC
  `$$$        4F"" .,uuuuuuu,,,,.              `CC '=-  "$$$="""$$r C><C
[...]

PR #1234, opened 2024-12-24
[build] Pre-build x86 viosock libraries as needed

benyamin-codez:
1. Pre-builds `x86` `viosock` libraries when building `virtio-win.sln` or `viosock.sln` for `amd64` in the circumstances the required `x86` libraries do not already exist.

Split from PR #1212.

### DIFF
--- a/build/build.bat
+++ b/build/build.bat
@@ -179,6 +179,13 @@ echo.
 pushd %BUILD_DIR%
 call "%~dp0SetVsEnv.bat" %TARGET_PROJ_CONFIG%
 
+rem Check for any prerequisite x86 libraries and build them if needed...
+call "%~dp0prebuild_x86_libs.bat" %BUILD_FILE% %BUILD_ARCH% "%BUILD_DIR%" %TARGET% %BUILD_FLAVOUR%
+IF ERRORLEVEL 1 (
+  set BUILD_FAILED=1
+  goto :build_arch_done
+)
+
 rem Split builds between Code Analysis and No-Analyis...
 if /I "!TAG!"=="SDV" (
   echo.
--- a/build/prebuild_x86_libs.bat
+++ b/build/prebuild_x86_libs.bat
@@ -1,0 +1,60 @@
+@echo off
+setlocal
+rem Check for x86 viosock libraries and build them if needed...
+call :do_viosock %*
+if ERRORLEVEL 1 (
+  endlocal
+  exit /B 1
+)
+endlocal
+goto :eof
+
+:do_viosock
+rem Lay up some variables
+set BUILD_FILE=%~1
+set BUILD_ARCH=%~2
+set BUILD_DIR=%~3
+set TARGET=%~4
+set BUILD_FLAVOUR=%~5
+set VIOSOCK_PREBUILD_X86_LIBS=
+set BUILD_FAILED=
+
+rem Which solutions need this?
+if "%BUILD_FILE%"=="virtio-win.sln" (
+  set VIOSOCK_PREBUILD_X86_LIBS=1
+)
+if "%BUILD_FILE%"=="viosock.sln" (
+  set VIOSOCK_PREBUILD_X86_LIBS=1
+)
+
+if "%VIOSOCK_PREBUILD_X86_LIBS%" EQU "1" (
+  rem Only proceed if we are building for x64
+  if %BUILD_ARCH%==x64 (
+    rem Check for x86 viosock libraries and build them if needed...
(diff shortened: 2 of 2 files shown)


build/build.bat
```diff
@@ -162,6 +165,42 @@ set TARGET_VS_CONFIG="%TARGET_PROJ_CONFIG% %BUILD_FLAVOR%|%BUILD_ARCH%"
 pushd %BUILD_DIR%
 call "%~dp0\SetVsEnv.bat" %TARGET_PROJ_CONFIG%
 
+rem Check for x86 viosock libraries and build them if needed...
+if "%BUILD_FILE%"=="virtio-win.sln" (
```
YanVugenfirer, line 169: Can he have this block as a separate batch file called form build.bat (prebuild_x86_vsock.bat)?
In the future, we will most probably have something similar for virtio-gpu as well.

YanVugenfirer (changes requested):

build/build.bat
benyamin-codez: We can, but it's only worthwhile if it's the whole thing, less some error checking on return.
This will suit inclusions for other drivers too, e.g. your mention of `virtio-gpu`.
The only downside to this is that it will always be called - for every build - until retired.

benyamin-codez: Okay, so I've dropped that into `prebuild_x86_libs.bat`.

benyamin-codez: Just one minor cleanup... 8^d

benyamin-codez: Ok, all sorted...

benyamin-codez:
@YanVugenfirer @kostyanf14 

This should probably have a new `HCK-CI [EWDK]` check as well as some `HCK-CI/viosock` checks too, yes?

build/build.bat
```diff
@@ -162,6 +165,42 @@ set TARGET_VS_CONFIG="%TARGET_PROJ_CONFIG% %BUILD_FLAVOR%|%BUILD_ARCH%"
...
+  set VIOSOCK_PREBUILD_X86_LIBS=1
+)
+if "%BUILD_FILE%"=="viosock.sln" (
```
YanVugenfirer: Sorry for being picky, but let's name the variable BUILD_SOLUTION instead of BUILD_FILE

YanVugenfirer (changes requested):

build/build.bat
benyamin-codez: @YanVugenfirer 

Thanks, Yan. That's an outdated section btw, `build\build.bat` only has this now:

https://github.com/virtio-win/kvm-guest-drivers-windows/blob/b93f4a30939f2a7f3346a55e1cecd312faf8a23f/build/build.bat#L165-L170

The `BUILD_FILE` variable is historically used throughout `build\build.bat` (12 instances). Moreover, it's perhaps worth pointing out that in practice `BUILD_FILE` more often refers to a project file (`*.vcxproj`) than a solution file (`*.sln`). As such, I'm not sure how appropriate it is to change it.

Also, if you do still want me to change it, I'm not sure if this is the appropriate PR to do so.

In this PR I could certainly change all instances (5) in `build\prebuild_x86_libs.bat`, and then perhaps raise another PR to change it in `build\build.bat` following merging of this and PR #1231. The alternative is to change it here and rebase PR #1231 once this one merges.

Please do let me know ASAP what you would prefer I do.

benyamin-codez:
@kostyanf14 

Did something break with the checks...?!?

YanVugenfirer:
@benyamin-codez Please rebase

YanVugenfirer:
@benyamin-codez Also, please squash the commits.

YanVugenfirer (approved):

YanVugenfirer:
Waiting for CI to wake up.

benyamin-codez:
> Waiting for CI to wake up.

Did it die...?

YanVugenfirer:
> > Waiting for CI to wake up.
> 
> Did it die...?

Yes :(
There is a problem with HW on one of the CI servers. @kostyanf14 bringing it up.

kostyanf14:
rerun tests

benyamin-codez:
Thanks @kostyanf14.

Did you want to run any `HCK-CI/viosock` checks, esp. for `amd64` only (so the new script can run the build path)...?

YanVugenfirer:
> Thanks @kostyanf14.
> 
> Did you want to run any `HCK-CI/viosock` checks, esp. for `amd64` only (so the new script can run the build path)...?

Konstantin double-checked the build. Merging.

Thanks!

PR #1251, opened 2025-01-07
Rhel 69070 apply clang format to virt io driver

Jedoku:


kostyanf14:
@Jedoku What it is?  VirtIO already formated

kostyanf14:
@Jedoku What it is?  VirtIO already formated

PR #1255, opened 2025-01-09
RHEL-11854: exclude trace files from clang format

Jedoku:


.github/workflows/clang-format.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       matrix:
```
kostyanf14, line 9: `s/clang-fomart/clang-format/g`

kostyanf14 (changes requested):

kostyanf14 (approved):

YanVugenfirer (approved):

PR #1276, opened 2025-01-30
format try option   [CI-NO-BUILD] [SKIP-HCK-CI]

kostyanf14:


kostyanf14:
rerun tests

YanVugenfirer:
Looks like it made things worse in some cases. Still looking for solution

PR #1284, opened 2025-02-03
[CI-NO-BUILD] [build] Extend ANSI colour palette to stdout

benyamin-codez:
1. Extended ANSI colour palette to `stdout`
2. Update `echo` prints to use ANSI colour palette
3. Includes self-deleting scheduled task to clean `build_log.txt` of colour palette artefacts

Split from PR #1212.


benyamin-codez:
@kostyanf14

Hope you're well.

Did you have anything further for me to consider wrt this one?

Best regards,
Ben

kostyanf14:
rerun tests

buildAll.bat
```diff
@@ -89,15 +90,38 @@ if not exist "%dvl_file%" (
...
+set "_c_Grn="40;92m""
+set "_c_Yel="40;93m""
+set "_c_Cyn="40;96m""
+set "_c_Wht="40;37m""
+goto :eof
+
 :leave
+if exist ".\build_log.txt" (
+  call schtasks /create /tn build_log_cleanup /tr "%comspec% /c %~dp0build\clean_build_log.bat" /sc ONCE /sd 01/01/1910 /st 00:00 1> nul 2>&1
+  call schtasks /run /tn build_log_cleanup 1> nul 2>&1
```
kostyanf14, line 124: I don't like the idea of triggering schtasks. This will work when the building is running on a clean Windows PC but what about docker or Windows containers? 

I think we don't need to cleanup files at all.

buildAll.bat
```diff
@@ -89,15 +90,38 @@ if not exist "%dvl_file%" (
...
+rem Get the ANSI ESC character: 0x27
+for /f "tokens=2 usebackq delims=#" %%i in (`"prompt #$H#$E# & echo on & for %%i in (1) do rem"`) do @(
+  set z_esc=%%i
+)
+rem Prepare pallette
+set "_c_Red="40;91m""
+set "_c_Grn="40;92m""
+set "_c_Yel="40;93m""
+set "_c_Cyn="40;96m""
+set "_c_Wht="40;37m""
```
kostyanf14, line 118: Maybe it is better to name these constants more like general log functions: debug, info(_c_Grn), warn(_c_Yel), error (_c_Red), etc.

benyamin-codez: That's a great idea. Methinks a level of abstraction here is worth the extra lines for helpers...
I will implement as `:log_info`, `:log_warn`, `:log_error` and `:log_success`.

buildAll.bat
benyamin-codez: Ok, so we will run with the `SYSTEM` user. This hides the `clean_build_log.bat` so it will execute without GUI, this should be sufficient for most Docker and container configurations. It also avoids the need to instead use `powershell` to achieve the same outcome, as this requires the XML configuration to be properly escaped for it to work as expected.

It seems prudent to give a bit more time (+2s = 5s) for handles to close in the `powershell` context and remove some for the visual check for debugging purposes (-3s = 2s).

_NOTE: Some containers might need to be started with the `-enableTaskScheduler` parameter, but most setups should just work._

> I think we don't need to cleanup files at all.

It is very ugly otherwise, like morning after too many :beers: ... 8^D  Per:

```
[40;92mBuild for Q35 succeeded.[[[[0m

[40;96mProcessing DVL files to create Windows 10, version 1607, WIN10_RS1 COMPAT version...[[[[0m
[40;93mWARNING : No DVL files were found.[[[[0m

[40;92mBUILD COMPLETED SUCCESSFULLY.[[[[0m
```

Vs:

```
Build for Q35 succeeded.

Processing DVL files to create Windows 10, version 1607, WIN10_RS1 COMPAT version...
WARNING : No DVL files were found.

BUILD COMPLETED SUCCESSFULLY.
```

benyamin-codez:
Apologies for prior lack of rebase... 8^d

buildAll.bat
benyamin-codez: What do you think? Does that look better?

buildAll.bat
benyamin-codez: Any further thoughts on this?

kostyanf14: Let's remove this.
1. You specified the `build_log.txt` file which is definitely not part of the build. This file can be generated by Jenkins in our case.
2. Jenkins doesn't know anything about schtasks, so it can start collecting artifacts after `buildAll.bat` exits. 

You should create `build/clean_log.bat` which receives the log file name for cleanup and cleanup of this file. If someone writes a log to a file, he should also run clean_log.bat for the file that he created.

benyamin-codez: Thanks for the further feedback, Kostiantyn.

Re:
(1) Yes, I only catered for your Jenkins-based automation. My thoughts were that others can mimic the naming; &
(2) I previously discussed how the clean should complete before signing is finished. Alternatively, you could issue a sleeper to be sure, e.g.: `powershell Start-Sleep -Seconds 7` after building or signing and before the artefact collection occurs.

I can certainly make `build\clean_log.bat` accept a parameter for _filename_ so it can act in a manual and universal manner. Are you sure you don't want this automated? Is `schtasks` too sneaky? Are you considering that you will just call `build\clean_log.bat build_log.txt` before artefact collection (via Jenkins)?

kostyanf14: Let's start with a fully stable variant with the manual solution and then think about an automated variant.

benyamin-codez: I've introduced a new environment variable `AUTOCLRCLEAN_VIRTIO_WIN_BLD_LOG`, which corresponds to a log file you want cleaned of colour artefacts. If the variable is defined and the file exists then the clean operation is scheduled.

Also updated `build\clean_build_log.bat` with more comments, debug prints and the filename parameter, plus added check for `SYSTEM` user where we now skip flashing the `stdout` cleaning log output for 5 seconds (for debug purposes). Also made sure that `AUTOCLRCLEAN_VIRTIO_WIN_BLD_LOG` is both relative and absolute path safe and also file name (white)space safe.

So, this means that unless you pop `AUTOCLRCLEAN_VIRTIO_WIN_BLD_LOG` with a valid file (path and) name and create that file by some external mechanism, e.g. via Jenkins, then the task will not be created or run (the current behaviour).

I also checked manual operation, per, e.g. `build\clean_build_log.bat .\build_log.txt`

I trust that is acceptable. If so, enjoy..!!

YanVugenfirer:

PR #1287, opened 2025-02-03
[CI-NO-BUILD] [build] Introduce CodeQL Management Framework

benyamin-codez:
1. Introduced managed CodeQL binary, package cache and test suite versioning control. If the CodeQL binary is detected at `CODEQL_BIN` a CodeQL build commences, otherwise it is skipped (unchanged behaviour).
2. The package cache versions are set in the `:config_ql_whcp` function (new). Test suite versioning is via git hash of the `WDK Developer Supplemental Tools` repository (new, depends on `git`). The versions are determined by the value of the `WHCP_LEVEL` variable, which can be either `WHCP_LEGACY` or `WHCP_24H2`. Provision for future WHCP versions is templated as `WHCP_NEXT`.
3. Introduced environment variable `SKIP_SDV_ACTUAL` to skip Static Driver Verifier (SDV) during analysis builds. CodeQL, Code Analysis (CA) and DVL operations are still executed. Default is to perform the SDV build (unchanged behaviour).
4. Introduced environment variable `CODEQL_OFFLINE_ONLY` to perform binary, package cache and suite version checking but NOT download any needed updates. Exits on error expecting resolution. Requires `git`.
5. Introduced environment variable `CODEQL_RUN_BLIND` to retain legacy behaviour and perform no binary, package cache or suite version checking. It presumes the prerequisites are present. Removes `git` dependency.
6. Make actual SDV build conditional (perform for `Win10_SDV` only)
7. Uses built-in tools available from WIN10_RS3:
    (a) `curl` to download the CodeQL archive; and
    (b) `Powershell` to extract it.

Split from PR #1212.


YanVugenfirer:

PR #1196, opened 2024-11-20
NO-SDV [vioscsi] Reduce spinlock management complexity

benyamin-codez:
Spinlocks are now primarily of the `DpcLock` type, without which the driver will not properly operate. Fallback to an `InterruptLock` type spinlock is limited, perhaps even cosmetic.

This commit removes the `VioScsiVQLock()` and `VioScsiVQUnlock()` wrapper routines to use the native functions instead, i.e. `StorPortAcquireSpinLock()` and `StorPortReleaseSpinLock()`. `VioScsiVQLock()` contained a superfluous MSI-X and `QueueNumber` check and `VioScsiVQUnlock()` did not release non-`DpcLock` type spinlocks. Removing the wrapper routines also resolves these two issues.

Refactoring includes:
* Renaming of `ProcessQueue()` to `ProcessBuffer()`. Whilst technically more accurate, this also provides space for a new `ProcessQueue()` routine, via which the `HW_MESSAGE_SIGNALLED_INTERRUPT_ROUTINE` and `HW_INTERRUPT` routines can then be consolidated.
* Where necessary, changing BOOLEAN `isr` to `STOR_SPINLOCK` `LockMode`
* CamelCase / conformity correction of `MessageID` to `MessageId`
* Conformity correction of `QueuNum` to `QueneNumber`
* Removal of unnecessary use of `MsgID` and `MessageID` or `MessageId` in area of focus (11 instances of `MessageID` remain in other `vioscsi.c` routines)
* Mnemonic rename of:
  * `CompletePendingRequests()` to `CompletePendingRequestsOnReset()`
  * `index` to `vq_req_idx`
* Syntactic correction of `if` statement (added braces)

DUPLICATE of PR #1175 to produce clean rebased version.

PR #1294, opened 2025-02-16
[CI-NO-BUILD] Nutanix-ENG-741981 [vioscsi] Fix NMI in crashdump/hibernation pathway

benyamin-codez:
Purpose: To resolve regression introduced in PR #1196 / #1175 (fe07e50).

Background: We previously ignored calls for a spinlock with isr=TRUE in VioScsiVQLock() and VioScsiVQUnlock(). This was replaced with a call to InterruptLock in the (!IsCrashDumpMode && adaptExt->dpc_ok) = FALSE pathway. In testing, suspend/resume/hibernate did not use this pathway but instead issued DPCs. The InterruptLock was presumed to be used when IsCrashDumpMode=TRUE.

This preliminary fix proposes to use DpcLevelLock when adaptExt->dpc_ok=TRUE and InterruptLock when adaptExt->dpc_ok=FALSE.


kostyanf14:
@MartinCHarvey-Nutanix Please look at this fix

benyamin-codez:
> @MartinCHarvey-Nutanix Please look at this fix

He's on PTO for a bit.

benyamin-codez:
@kostyanf14 

Can we please rerun tests on this one? I'm confident it should be ready.

JonKohler:
This looks OK to me. @sb-ntnx can you give it a look?

benyamin-codez:
@kostyanf14

Maybe some `HCK-CI/vioscsi` checks too?
I'm curious whether any of those failing checks might now pass...

kostyanf14:
> @kostyanf14
> 
> Maybe some `HCK-CI/vioscsi` checks too? I'm curious whether any of those failing checks might now pass...

Yes, all of them are in the queue

benyamin-codez:
> @kostyanf14
> 
> Is the `Flush Test` failure expected in `HCK-CI/vioscsi-Win2025x64_host`...?

I note that this test passed in PR #1293...
...which - if valid - would mean @MartinCHarvey-Nutanix has found an additional fix beyond the regression...
... not that I can see what that might be.
Maybe he's worked around an upstream (StorPort) bug...?

cc: @JonKohler @sb-ntnx

benyamin-codez:
@JonKohler @sb-ntnx

My latest thoughts on this are:

Perhaps the init of `PVOID LockContext = NULL;`...
...and then when I later try `LockContext = &adaptExt->dpc[vq_req_idx];` ...
...that is actually a problem. 8^O

This was because I used `PVOID LockContext = NULL; // sanity check for LockMode = InterruptLock or StartIoLock` in my WIP. We can now probably get rid of this in all the places where a `DpcLock` type spinlock is unconditionally used.
This is what Martin did and the `Flush Test` seemed to work as expected.

My latest cut (which I just rebased here) and Martin's are getting very close to being the same. I guess I'm preferring mine over minor semantics. I'd like to retain a reference to `STOR_SPINLOCK` so the spinlock context is considered when calling the `ProcessBuffer()` function.

Is @MartinCHarvey-Nutanix back from PTO... It would be good to get his view on the differences. Anyone else?

@kostyanf14, could we please re-run the test to be sure?

benyamin-codez:
> @kostyanf14, could we please re-run the [`HCK-CI/vioscsi-Win2025x64_host`] test to be sure?

It will be very telling if it passes...

I note PR #1293 also had:

a)  `HCK-CI/vioscsi-Win2022x64` fail the `Static Tools Logo Test`, the `Flush Test` and `Disk Verification (LOGO)`.
b) `HCK-CI/vioscsi-Win11_24H2x64_host_uefi_viommu` fail `Disk Verification (LOGO)`.

Both of these also skipped some tests.
Are any of these failures unexpected?

kostyanf14:
@benyamin-codez 

Static Tools Logo Test unexpected. For some reason SDV crashed for vioscsi driver. 
Flush Test and Disk Verification (LOGO) - known issue

benyamin-codez:
@kostyanf14 

> Static Tools Logo Test unexpected. For some reason SDV crashed for vioscsi driver.
> Flush Test and Disk Verification (LOGO) - known issue

Thanks, Kostiantyn.

Just to be clear, was the `Flush Test` failure expected or unexpected in `HCK-CI/vioscsi-Win2025x64_host`...?
Can we please re-run  it to confirm my theory re using `PVOID LockContext = NULL` being a problem...?

kostyanf14:
Flush Test failure is expected in general. We try to emulate force off of VMs and sometimes HLK is not happy. If the test passed - this is good, if fail - not a bug.

benyamin-codez:
@JonKohler @sb-ntnx @MartinCHarvey-Nutanix @MartinCHarvey

In light of @kostyanf14's comments re the `Flush Test` above, the fact the same just passed, and a little time to consider other options, I am minded to leave those last edits as-is. The relevant functions in my WIP also had a `STOR_SPINLOCK` parameter, which I removed when I realised those functions would never use anything but `DpcLock` type spinlocks moving forward, but I obviously did leave some artefacts behind, which I guess should be cleaned up anyway.

I am now also minded to perhaps create a custom `STOR_SPINLOCK`-based enumeration, perhaps named `CUSTOM_STOR_SPINLOCK`, which could then include a `Skip_Locking=0` or `No_Lock=0` or `Invalid_Lock=0` spinlock type for use in this scenario, or which would otherwise mimic the Win11 24H2 enumeration (which uses `InvalidLock=0`).

Probably something like:

```c
typedef enum _CUSTOM_STOR_SPINLOCK {
    Skip_Locking = 0,
    No_Lock = 0,
#if defined(NTDDI_WIN11_GE) && (NTDDI_VERSION >= NTDDI_WIN11_GE)
    Invalid_Lock = InvalidLock,
#else
    Invalid_Lock = 0,    
#endif
    Dpc_Lock = DpcLock,
    StartIo_Lock = StartIoLock,
    Interrupt_Lock = InterruptLock,
    ThreadedDpc_Lock = ThreadedDpcLock,
    DpcLevel_Lock = DpcLevelLock
} CUSTOM_STOR_SPINLOCK;
```

I think it is highly likely a `ThreadedDpcLock`-based solution will be considered as some point in the not-to-distant future. I had some success with this in my WIP (an unpublished variant). For this reason, and as mentioned above, i.e. to help ensure the spinlock context is considered when calling the `ProcessBuffer()` function, I think we should use a `STOR_SPINLOCK`-based parameter rather than the `BOOLEAN` that Martin has used in the other PR.

I'll start putting that together, but in the meantime please do share any thoughts you have about the above.

Also, who discovered the regression?
I would like to give credit to Martin and whomever else might be deserving in the commit message.

benyamin-codez:
@JonKohler @sb-ntnx @MartinCHarvey-Nutanix @MartinCHarvey

Please find my new cut, crafted and pushed per the above post, for your consideration.

@kostyanf14 

Maybe we should try `HCK-CI/vioscsi-Win2022x64` this round. I had no SDV failures on my end.

benyamin-codez:
_P.S. :_
Small boo-boo in commit message:
Duplicated `ThreadedDpc_Lock = ThreadedDpcLock`, one should be `DpcLevel_Lock = DpcLevelLock`... 8^d

benyamin-codez:
@kostyanf14 

> Flush Test failure is expected in general. We try to emulate force off of VMs and sometimes HLK is not happy. If the test passed - this is good, if fail - not a bug.

wrt your power-loss emulation code for the `Flush Test`, is that open for peeking?
I'd imagine you have a QEMU-based solution, so maybe:
`qm stop <vmid> --overrule-shutdown 1 --timeout 0`
I was curious whether you might also be using `--keepActive 1` or even `--skiplock 1` (as su).
...or do you just use `kill -9 <vm_pid>`...?

benyamin-codez:
@kostyanf14 

Apologies for clobbering the previous `HCK-CI/vioscsi-Win2025x64_host` check...!! 8^d

kostyanf14:
> @kostyanf14
> 
> > Flush Test failure is expected in general. We try to emulate force off of VMs and sometimes HLK is not happy. If the test passed - this is good, if fail - not a bug.
> 
> wrt your power-loss emulation code for the `Flush Test`, is that open for peeking? I'd imagine you have a QEMU-based solution, so maybe: `qm stop <vmid> --overrule-shutdown 1 --timeout 0` I was curious whether you might also be using `--keepActive 1` or even `--skiplock 1` (as su). ...or do you just use `kill -9 <vm_pid>`...?

Please see https://github.com/HCK-CI/AutoHCK/ in general and https://github.com/HCK-CI/AutoHCK/pull/486 specific for this test

benyamin-codez:
@kostyanf14 

Looks like `HCK-CI/vioscsi-Win2025x64_host` passed `Flush Test` again but failed `Disk Verification (LOGO)` this time...

> Maybe we should try `HCK-CI/vioscsi-Win2022x64` this round. I had no SDV failures on my end.

Ready for this one?

benyamin-codez:
@MartinCHarvey-Nutanix @MartinCHarvey @sb-ntnx 

Thanks for the post, Martin.

> Have we addressed the case where this happens at very initial driver load and device start?

I have been unable to reproduce this issue, so I have presumed you are using something funky in your setup...
We did test per comments from [here](https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1175#issuecomment-2439905009) in PR #1175 (which later became PR #1196 where the regression was introduced).
Do you have a setup not covered in those comments?

> My understanding is that the ISR gets called after initial device addition, but before the DPC has been set up.

ISR = `VioScsiCompleteDpcRoutine()` which calls `ProcessBuffer()` with a `DpcLock` type spinlock.

iirc, the codepath for setting it up is:

`VioScsiHwInitialize()` [`HW_INITIALIZE` following `VioScsiFindAdapter()`]
...
... `StorPortEnablePassiveInitialization(DeviceExtension, VioScsiPassiveInitializeRoutine)`
...
`VioScsiPassiveInitializeRoutine()`  [`HW_PASSIVE_INITIALIZE_ROUTINE` following `HW_INITIALIZE`]

Where:
`VioScsiPassiveInitializeRoutine()`
... calls `StorPortInitializeDpc()` for each request queue...
... to create StorPort DPC objects [`STOR_DPC` (_Buffer, Lock_)] at each pointer `adaptExt->dpc[vq_idx]`...
... associating the ISR [`VioScsiCompleteDpcRoutine()`] with each StorPort DPC object...
... and then blindly sets `adaptExt->dpc_ok = TRUE` (as `StorPortInitializeDpc()` does not return a status).

Perhaps `adaptExt->dpc_ready` would be a mnemonic improvement upon `adaptExt->dpc_ok`.

If `StorPortEnablePassiveInitialization()` returns `FALSE`, i.e. the system does not support DPCs,  then the `VioScsiPassiveInitializeRoutine()` routine will never be called and `adaptExt->dpc_ok` will remain `FALSE`. The trouble with such a scenario is that we rely on DPCs and `DpcLock` everywhere else, so I'm not sure how the driver would work.

Is this the case with your setup? I ask because the DPC infrastructure should be well and truly set up, by the time the `VioScsiMSInterrupt()` is called by StorPort following `VioScsiStartIo()`. Why else would you hit this codepath...?

In the case DPCs are supported, and we happen to use a `STOR_SPINLOCK` type other than`DpcLock`, we will clobber any DPCs/`DpcLock` spinlocks, as the new spinlock will be issued above `DISPATCH_LEVEL`. In my testing, using a different spinlock type usually caused a BSOD (typ. `IRQL_NOT_LESS_OR_EQUAL`), and not a "hung" VM...

So historically, we have skipped the spinlock. This PR restores this previous behaviour.

Please let me know if you can still reproduce your issue even with this PR.
It is possible we might need [in `DispatchQueue()`]:

```c
    if (!adaptExt->dump_mode && adaptExt->dpc_ok)
    {
        NT_ASSERT(MessageId >= QUEUE_TO_MESSAGE(VIRTIO_SCSI_REQUEST_QUEUE_0));
        StorPortIssueDpc(DeviceExtension,
                         &adaptExt->dpc[MessageId - QUEUE_TO_MESSAGE(VIRTIO_SCSI_REQUEST_QUEUE_0)],
                         ULongToPtr(MessageId),
                         ULongToPtr(MessageId));
    }
    else if (adaptExt->dump_mode)
    {
        ProcessBuffer(DeviceExtension, MessageId, Skip_Locking);
    }
    else
    {
        ProcessBuffer(DeviceExtension, MessageId, Interrupt_Lock);
    }
```

...or some other variant + code in `ProcessBuffer()` for `InterruptLock` or other spinlock types... 8^d

benyamin-codez:
@MartinCHarvey-Nutanix @MartinCHarvey @sb-ntnx

The one variant I didn't try was:

```c
    if (!adaptExt->dump_mode && adaptExt->dpc_ok)
    {
        ...
    }
    else
    {
        ProcessBuffer(DeviceExtension, MessageId, Dpc_Lock);
    }
```
..., which in ***some*** circumstances, might actually work...

MartinCHarvey-Nutanix:
> @MartinCHarvey-Nutanix @MartinCHarvey @sb-ntnx
> 
> I have been unable to reproduce this issue, so I have presumed you are using something funky in your setup... 

I suspect this may be the case. I'm testing the case where there is a totally "bare" windows VM, which has a vioscsi device, and additionally (I guess), there are pending events / Io's / something on the device at the moment where a driver is very first installed.

> Do you have a setup not covered in those comments?

Possibly. Nutanix runs storage stuff on the host which is not "vanilla upstream".
 
> > My understanding is that the ISR gets called after initial device addition, but before the DPC has been set up.

> Perhaps `adaptExt->dpc_ready` would be a mnemonic improvement upon `adaptExt->dpc_ok`.

Yes, dpc_ok strikes me as a bit ambigous. I think a bit of instrumentation with some dpc_state variable might be a good idea.

> If `StorPortEnablePassiveInitialization()` returns `FALSE`, i.e. the system does not support DPCs

I'm willing to believe this generally shouldn't happen in most/all normal use cases. Possibly the dump/hiber path doesn't allow DPC's in the normal case. I'll get back to you on that.

> Is this the case with your setup? I ask because the DPC infrastructure should be well and truly set up, by the time the `VioScsiMSInterrupt()` is called by StorPort following `VioScsiStartIo()`. Why else would you hit this codepath...?

Hmmm. I notice there's a "virtio_device_ready" at the end of HwInitialize, but before passive initialization has completed, and strictly speaking (I haven't checked the virtio spec) we might not want to indicate fully ready until passive init has succeded or failed. Maybe some interrupt not strictly associated with a vm/client side request?

> Please let me know if you can still reproduce your issue even with this PR. It is possible we might need [in `DispatchQueue()`]:

Hmmm. OK. Some concrete evidence and instrumentation / tracing is required. Leave it with me, and I'll get back to you in the next day or two on that.

MartinCHarvey-Nutanix:
@benyamin-codez I have yet to check whether any of these issues also apply to virtio-stor.

benyamin-codez:
@MartinCHarvey-Nutanix

Thanks for the updates..!

I started getting `DRIVER_IRQL_NOT_LESS_OR_EQUAL` in `ntoskrnl.exe` at driver installation (in my `vioscsi` WIP) a few hours back following `clang-format` changes... 8^d

It's perhaps unrelated, but I remain curious as to why your environment is calling:
`VioScsiMSInterrupt() [HW_MESSAGE_SIGNALED_INTERRUPT_ROUTINE]`
...before `virtio_device_ready()`, or even before `VioScsiStartIo()` for that matter.

Did you find a cause for why the MSI is being raised so early...?

> ...move that virtio_device_ready call to the end of passive initialization.

In the event that passive initialisation doesn't occur, i.e. `adaptExt->dpc_ok = FALSE`, how do we ensure the  `virtio_device_ready()` routine still runs?

benyamin-codez:
@MartinCHarvey-Nutanix

> I started getting `DRIVER_IRQL_NOT_LESS_OR_EQUAL` in `ntoskrnl.exe` at driver installation (in my `vioscsi` WIP) a few hours back following `clang-format` changes... 8^d
>
> It's perhaps unrelated...

It was unrelated...
... 8^d
... ... 8^D

Any joy moving that `virtio_device_ready()` routine...?

I was thinking it might be the case that it is required where it is, ...
...i.e. before StorPort calls `HW_PASSIVE_INITIALIZE_ROUTINE` (i.e. `VioScsiPassiveInitializeRoutine()`) 
...i.e. in `HW_INITIALIZE` (i.e. `VioScsiHwInitialize()`)...

...so I'm very curious to see where you might call it from, and if that works.

MartinCHarvey-Nutanix:
@benyamin-codez Give me a mo to retest both normal usage and crashdump. I'll update my PR with a couple of commites, so you can cherry-pick one of them across.

MartinCHarvey:
Yeah,

https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nf-storport-storportenablepassiveinitialization

Is not particularly forthcoming about which old os versions don't support DPC's so I assume it's from some ancient version. Everything NT based does.

If you look at the code, failure of passive initialization results in failed HW Init, and I'm leaving that unchanged. It should be fine to delay the "ready" notification for a bit.

Still hoop jumping to get this to host a boot drive. Getting it to host the location of the dump file should be easier. Feedback RSN.


Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

benyamin-codez:
FYI, I'm prepping to close this in favour of Martin's PR #1293.
Refactoring for same and Just doing some testing to be sure...

PR #1298, opened 2025-02-17
[CI-NO-BUILD] [viostor] Programatically determine max_segments

benyamin-codez:
Refactors `VirtIoFindAdapter()` to:

1. Enable Registry lookup for `max_segments` in value "MaxPhysicalSegments"
2. Refactor existing options into proposed `vioscsi` scaffold
3. Calc sectors from geometry and segments from sectors
4. Choose from among best candidates and check within limits
5. Enforce legacy StorPort NOPB off-by-one behaviour and provide explanation
6. Improves relevant instrumentation

Note: This commit enables the use of max_segments greater than
      `info.seg_max` (max = 254) up to `MAX_PHYS_SEGMENTS` (presently 512).
      This commit also fixes a NOPB off-by-one regression.

PR #1300, opened 2025-02-17
[CI-NO-BUILD] [viostor] Refactor VirtIoHwInitialize()

benyamin-codez:
Refactors `VirtIoHwInitialize()` to improve instrumentation.

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -670,6 +670,7 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
     ULONG status = STOR_STATUS_SUCCESS;
     MESSAGE_INTERRUPT_INFORMATION msi_info = {0};
     PREQUEST_LIST element = NULL;
+    ULONG index = 0;
 
     adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
 
@@ -678,46 +679,224 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
     adaptExt->poolOffset = 0;
     adaptExt->stopped = FALSE;
 
+    if ((!adaptExt->dump_mode) && (adaptExt->num_queues > 1) && (adaptExt->perfFlags == 0))
+    {
+        perfData.Version = STOR_PERF_VERSION;
+        perfData.Size = sizeof(PERF_CONFIGURATION_DATA);
+
+        status = StorPortInitializePerfOpts(DeviceExtension, TRUE, &perfData);
+
+        RhelDbgPrint(TRACE_LEVEL_VERBOSE,
+                     " PERF: Perf Version = 0x%x, Flags = 0x%x, ConcurrentChannels = %d, FirstRedirectionMessageNumber "
+                     "= %d, LastRedirectionMessageNumber = %d, DeviceNode = %d\n",
+                     perfData.Version,
+                     perfData.Flags,
+                     perfData.ConcurrentChannels,
+                     perfData.FirstRedirectionMessageNumber,
+                     perfData.LastRedirectionMessageNumber,
+                     perfData.DeviceNode);
+    }
+    else
+    {
+        RhelDbgPrint(TRACE_LEVEL_WARNING, " PERF: StorPortInitializePerfOpts GET failed with status = 0x%x\n", status);
+    }
+    if ((status == STOR_STATUS_SUCCESS) && (CHECKFLAG(perfData.Flags, STOR_PERF_DPC_REDIRECTION)))
+    {
+        adaptExt->perfFlags = STOR_PERF_DPC_REDIRECTION;
+        if (CHECKFLAG(perfData.Flags, STOR_PERF_CONCURRENT_CHANNELS))
+        {
+            adaptExt->perfFlags |= STOR_PERF_CONCURRENT_CHANNELS;
+            perfData.ConcurrentChannels = adaptExt->num_queues;
+        }
+        if (CHECKFLAG(perfData.Flags, STOR_PERF_INTERRUPT_MESSAGE_RANGES))
+        {
+            adaptExt->perfFlags |= STOR_PERF_INTERRUPT_MESSAGE_RANGES;
+            perfData.FirstRedirectionMessageNumber = VIRTIO_BLK_MSIX_VQ_OFFSET;
+            perfData.LastRedirectionMessageNumber = perfData.FirstRedirectionMessageNumber +
+                                                    perfData.ConcurrentChannels - 1;
+        }
+        ASSERT(perfData.lastRedirectionMessageNumber < adaptExt->num_affinity);
+        if ((adaptExt->pmsg_affinity != NULL) && CHECKFLAG(perfData.Flags, STOR_PERF_ADV_CONFIG_LOCALITY))
+        {
+            RtlZeroMemory((PCHAR)adaptExt->pmsg_affinity, sizeof(GROUP_AFFINITY) * ((ULONGLONG)adaptExt->num_affinity));
+            adaptExt->perfFlags |= STOR_PERF_ADV_CONFIG_LOCALITY;
+            perfData.MessageTargets = adaptExt->pmsg_affinity;
+        }
+        if (CHECKFLAG(perfData.Flags, STOR_PERF_DPC_REDIRECTION_CURRENT_CPU))
+        {
+            adaptExt->perfFlags |= STOR_PERF_DPC_REDIRECTION_CURRENT_CPU;
+        }
+        if (CHECKFLAG(perfData.Flags, STOR_PERF_OPTIMIZE_FOR_COMPLETION_DURING_STARTIO))
+        {
+            adaptExt->perfFlags |= STOR_PERF_OPTIMIZE_FOR_COMPLETION_DURING_STARTIO;
+        }
+        if (CHECKFLAG(perfData.Flags, STOR_PERF_NO_SGL))
+        {
+            /* FIXME : We still use
+             *          StorPortGetScatterGatherList(), and
+             *          ConfigInfo->ScatterGather = TRUE,
+             *         so not sure why we are using STOR_PERF_NO_SGL here.
+             *         Does not enable anyway...
+             */
+            adaptExt->perfFlags |= STOR_PERF_NO_SGL;
+        }
+        perfData.Flags = adaptExt->perfFlags;
+        RhelDbgPrint(TRACE_LEVEL_VERBOSE,
+                     " PERF: Perf Version = 0x%x, Flags = 0x%x, ConcurrentChannels = %d, "
+                     "FirstRedirectionMessageNumber = %d, LastRedirectionMessageNumber = %d, DeviceNode = %d\n",
+                     perfData.Version,
+                     perfData.Flags,
+                     perfData.ConcurrentChannels,
+                     perfData.FirstRedirectionMessageNumber,
+                     perfData.LastRedirectionMessageNumber,
+                     perfData.DeviceNode);
+        RhelDbgPrint(TRACE_LEVEL_VERBOSE,
+                     " PERF: STOR_PERF_DPC_REDIRECTION flag is : %s \n",
+                     (CHECKFLAG(perfData.Flags, STOR_PERF_DPC_REDIRECTION)) ? "ON" : "OFF");
+        RhelDbgPrint(TRACE_LEVEL_VERBOSE,
+                     " PERF: STOR_PERF_CONCURRENT_CHANNELS flag is: %s \n",
+                     (CHECKFLAG(perfData.Flags, STOR_PERF_CONCURRENT_CHANNELS)) ? "ON" : "OFF");
+        RhelDbgPrint(TRACE_LEVEL_VERBOSE,
+                     " PERF: STOR_PERF_INTERRUPT_MESSAGE_RANGES flag is : %s \n",
+                     (CHECKFLAG(perfData.Flags, STOR_PERF_INTERRUPT_MESSAGE_RANGES)) ? "ON" : "OFF");
+        RhelDbgPrint(TRACE_LEVEL_VERBOSE,
+                     " PERF: STOR_PERF_ADV_CONFIG_LOCALITY flag is: %s \n",
+                     (CHECKFLAG(perfData.Flags, STOR_PERF_ADV_CONFIG_LOCALITY)) ? "ON" : "OFF");
+        RhelDbgPrint(TRACE_LEVEL_VERBOSE,
+                     " PERF: STOR_PERF_OPTIMIZE_FOR_COMPLETION_DURING_STARTIO flag is: %s \n",
+                     (CHECKFLAG(perfData.Flags, STOR_PERF_OPTIMIZE_FOR_COMPLETION_DURING_STARTIO)) ? "ON" : "OFF");
+        RhelDbgPrint(TRACE_LEVEL_VERBOSE,
+                     " PERF: STOR_PERF_DPC_REDIRECTION_CURRENT_CPU flag is : %s \n",
+                     (CHECKFLAG(perfData.Flags, STOR_PERF_DPC_REDIRECTION_CURRENT_CPU)) ? "ON" : "OFF");
+        RhelDbgPrint(TRACE_LEVEL_VERBOSE,
+                     " PERF: STOR_PERF_NO_SGL flag is : %s \n",
+                     (CHECKFLAG(perfData.Flags, STOR_PERF_NO_SGL)) ? "ON" : "OFF");
+
+        status = StorPortInitializePerfOpts(DeviceExtension, FALSE, &perfData);
+
+        if (status != STOR_STATUS_SUCCESS)
+        {
+            adaptExt->perfFlags = 0;
+            RhelDbgPrint(TRACE_LEVEL_ERROR,
+                         " PERF: StorPortInitializePerfOpts SET failed with status = 0x%x\n",
+                         status);
+        }
+        for (index = 0; index < adaptExt->num_affinity; ++index)
+        {
+            GROUP_AFFINITY vector_affinity = adaptExt->pmsg_affinity[index];
+            RhelDbgPrint(TRACE_LEVEL_VERBOSE,
+                         " PERF: MSI-X Vector %lu CPU Affinity : KAFFINITY Mask = %I64d, CPU Group = %lu \n",
+                         index,
+                         vector_affinity.Mask,
+                         vector_affinity.Group);
+        }
+    }
+
     while (StorPortGetMSIInfo(DeviceExtension, adaptExt->msix_vectors, &msi_info) == STOR_STATUS_SUCCESS)
     {
-        RhelDbgPrint(TRACE_LEVEL_INFORMATION, " MessageId = %x\n", msi_info.MessageId);
-        RhelDbgPrint(TRACE_LEVEL_INFORMATION, " MessageData = %x\n", msi_info.MessageData);
-        RhelDbgPrint(TRACE_LEVEL_INFORMATION, " InterruptVector = %x\n", msi_info.InterruptVector);
-        RhelDbgPrint(TRACE_LEVEL_INFORMATION, " InterruptLevel = %x\n", msi_info.InterruptLevel);
-        RhelDbgPrint(TRACE_LEVEL_INFORMATION,
-                     " InterruptMode = %s\n",
-                     msi_info.InterruptMode == LevelSensitive ? "LevelSensitive" : "Latched");
-        RhelDbgPrint(TRACE_LEVEL_INFORMATION, " MessageAddress = %I64x\n\n", msi_info.MessageAddress.QuadPart);
+        if (adaptExt->num_queues > 1)
+        {
+            RhelDbgPrint(TRACE_LEVEL_INFORMATION,
+                         " MSI-X Vector [MessageId] = %lu, MessageAddress = 0x%I64x, MessageData = %lu, "
+                         "InterruptVector = %lu, InterruptLevel = %lu, InterruptMode = %s, CPU Affinity : "
+                         "Mask = %I64d, Group = %lu \n",
+                         msi_info.MessageId,
+                         msi_info.MessageAddress.QuadPart,
+                         msi_info.MessageData,
+                         msi_info.InterruptVector,
+                         msi_info.InterruptLevel,
+                         msi_info.InterruptMode == LevelSensitive ? "LevelSensitive" : "Latched",
+                         adaptExt->pmsg_affinity[msi_info.MessageId].Mask,
+                         adaptExt->pmsg_affinity[msi_info.MessageId].Group);
+        }
+        else
+        {
+            RhelDbgPrint(TRACE_LEVEL_INFORMATION,
+                         " MSI-X Vector [MessageId] = %lu, MessageAddress = 0x%I64x, MessageData = %lu, "
+                         "InterruptVector = %lu, InterruptLevel = %lu, InterruptMode = %s \n",
+                         msi_info.MessageId,
+                         msi_info.MessageAddress.QuadPart,
+                         msi_info.MessageData,
+                         msi_info.InterruptVector,
+                         msi_info.InterruptLevel,
+                         msi_info.InterruptMode == LevelSensitive ? "LevelSensitive" : "Latched");
+        }
         ++adaptExt->msix_vectors;
     }
 
-    RhelDbgPrint(TRACE_LEVEL_INFORMATION, " Queues %d msix_vectors %d\n", adaptExt->num_queues, adaptExt->msix_vectors);
-    if (adaptExt->num_queues > 1 && ((adaptExt->num_queues + 1U) > adaptExt->msix_vectors))
+    if ((adaptExt->num_queues > 1) && ((adaptExt->num_queues + VIRTIO_BLK_MSIX_VQ_OFFSET) > adaptExt->msix_vectors))
     {
         adaptExt->num_queues = (USHORT)adaptExt->msix_vectors;
     }
 
-    if (adaptExt->msix_vectors >= (adaptExt->num_queues + 1U))
+    if (!adaptExt->dump_mode && adaptExt->msix_vectors > 0)
     {
-        /* initialize queues with a MSI vector per queue */
-        RhelDbgPrint(TRACE_LEVEL_INFORMATION, (" Using a unique MSI vector per queue\n"));
-        adaptExt->msix_one_vector = FALSE;
+        if (adaptExt->msix_vectors >= (adaptExt->num_queues + VIRTIO_BLK_MSIX_VQ_OFFSET))
+        {
+            /* initialize queues with a MSI vector per queue */
+            adaptExt->msix_one_vector = FALSE;
+        }
+        else
+        {
+            /* if we don't have enough vectors, use one for all queues */
+            adaptExt->msix_one_vector = TRUE;
+        }
     }
     else
     {
-        /* if we don't have enough vectors, use one for all queues */
-        RhelDbgPrint(TRACE_LEVEL_INFORMATION, (" Using one MSI vector for all queues\n"));
-        adaptExt->msix_one_vector = TRUE;
+        /* initialize queues with no MSI interrupts */
+        adaptExt->msix_enabled = FALSE;
     }
 
     if (!InitializeVirtualQueues(adaptExt))
     {
         LogError(DeviceExtension, SP_INTERNAL_ADAPTER_ERROR, __LINE__);
-
-        RhelDbgPrint(TRACE_LEVEL_FATAL, (" Cannot find snd virtual queue\n"));
+        RhelDbgPrint(TRACE_LEVEL_FATAL, " !!! - Failed to initialize the Virtual Queues - !!!\n");
         virtio_add_status(&adaptExt->vdev, VIRTIO_CONFIG_S_FAILED);
-        return ret;
+        return FALSE;
     }
+
+    RhelDbgPrint(TRACE_LEVEL_INFORMATION,
+                 " VirtIO Request Queues : %d, MSI-X Enabled : %s, MSI-X Use ONE Vector : %s, MSI-X Vectors "
+                 "[msix_vectors] : %d \n",
+                 adaptExt->num_queues,
+                 (adaptExt->msix_enabled) ? "YES" : "NO",
+                 (adaptExt->msix_one_vector) ? "YES" : "NO",
+                 adaptExt->msix_vectors);
+    RhelDbgPrint(TRACE_LEVEL_VERBOSE,
+                 " MSI-X Vector %lu | StorPort exclusive control \n",
+                 (VIRTIO_BLK_MSIX_VQ_OFFSET - 1));
+
+    for (index = 0; index < adaptExt->num_queues; ++index)
+    {
+        element = &adaptExt->processing_srbs[index];
+        InitializeListHead(&element->srb_list);
+        element->srb_cnt = 0;
+        RhelDbgPrint(TRACE_LEVEL_VERBOSE,
+                     " MSI-X Vector %lu | VIRTIO Request Queue %lu (index %lu) \n",
+                     (index + VIRTIO_BLK_MSIX_VQ_OFFSET),
+                     (index + VIRTIO_BLK_MSIX_VQ_OFFSET),
+                     index);
+    }
+
+    if (!adaptExt->dump_mode)
+    {
+        /* we don't get another chance to call StorPortEnablePassiveInitialization and initialize
+         * DPCs if the adapter is being restarted, so leave our datastructures alone on restart
+         */
+        if (adaptExt->dpc == NULL)
+        {
+            adaptExt->dpc = (PSTOR_DPC)VioStorPoolAlloc(DeviceExtension, sizeof(STOR_DPC) * adaptExt->num_queues);
+        }
+        if (!adaptExt->dpc_ok && !StorPortEnablePassiveInitialization(DeviceExtension, VirtIoPassiveInitializeRoutine))
+        {
+            RhelDbgPrint(TRACE_LEVEL_FATAL, " StorPortEnablePassiveInitialization FAILED\n");
+            virtio_add_status(&adaptExt->vdev, VIRTIO_CONFIG_S_FAILED);
+            return FALSE;
+        }
+    }
+
+    virtio_device_ready(&adaptExt->vdev);
 
     memset(&adaptExt->inquiry_data, 0, sizeof(INQUIRYDATA));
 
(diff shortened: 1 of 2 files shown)


benyamin-codez:
~Shares some definitions with PR #1298. Waiting for that to merge before rebasing.~

viostor/virtio_stor.c
```diff
@@ -678,165 +679,239 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+    memset(&adaptExt->inquiry_data, 0, sizeof(INQUIRYDATA));
 
-    return ret;
+    adaptExt->inquiry_data.ANSIVersion = 4;
+    adaptExt->inquiry_data.ResponseDataFormat = 2;
+    adaptExt->inquiry_data.CommandQueue = 1;
+    adaptExt->inquiry_data.DeviceType = DIRECT_ACCESS_DEVICE;
+    adaptExt->inquiry_data.Wide32Bit = 1;
+    adaptExt->inquiry_data.AdditionalLength = 91;
+    StorPortMoveMemory(&adaptExt->inquiry_data.VendorId, "Red Hat ", sizeof("Red Hat "));
```
MartinCHarvey-Nutanix, line 909: Oh yuck. Can we make sure the const strings are the same in both places, at least via #define ?

benyamin-codez: Hey...!
I just moved it from Line 722... 8^d
Where's the other place...? :flushed:

MartinCHarvey-Nutanix: Not to mention, this might have to be a strlen. Are you trying to copy a pointer, or copy the string data?

vrozenfe:
@benyamin-codez 
Can you please explain the purpose of this patch?
Thanks,
Vadim.

benyamin-codez:
@vrozenfe 

> Can you please explain the purpose of this patch?

The `StorPortInitializePerfOpts()` part is refactored to make sure the trace has proper KAFFINITY values. The present method instantiates the mask afterwards when it needs to be first, resulting in unset values.

The instrumentation is also improved.

We will also address the find per @MartinCHarvey-Nutanix's post above.
On that, isn't string data for Vendor, Device, etc. usually set elsewhere?

viostor/virtio_stor.c
benyamin-codez: Thinking I might split this to a new PR...
cc: @vrozenfe

MartinCHarvey-Nutanix:
> @vrozenfe
> 
> > We will also address the find per @MartinCHarvey-Nutanix's post above.
> > On that, isn't string data for Vendor, Device, etc. usually set elsewhere?

I should note on that, the code is probably OK - ie, it's moving the memory without the null terminator, and does not overwrite anything, hence the trailing space. However, someone needs to have a look through the docs somewhere to find out whether the the string should be null terminated, optionally null terminated, or padded with spaces to the end of the (pretty small) amount of space provided. Since "Red Hat" is only 7 chars long, there's room for a null terminator.

Additionally, we should put a compile time assert in on the length of the string so that when some vendor (i.e. us and many others) tried to change some #define without thinking about it, it either barfs, or safely truncates the string.

benyamin-codez:
> Since "Red Hat" is only 7 chars long, there's room for a null terminator.
> ...whether the the string should be null terminated, optionally null terminated, or padded with spaces to the end...

It's presently padded due to `VendorId[8]`.

With `"VirtIO"` (6) in `ProductId[16]`, so 10 to go.
... and "0001" (4) in `ProductRevisionLevel[4]` - so all good here. A hint these may not be null terminated.
... and "0001" (4) in `VendorSpecific[20]`, so 16 to go.

> ...we should put a compile time assert in on the length of the string...

We can also add informative comments to the macros (even though maybe not seen by anyone)...
Something like:
```c
// Set Vendor Information - Observe INQUIRYDATA struct length limits
#if !defined(VIRTIO_BLK_VENDOR_ID)
#define VIRTIO_BLK_VENDOR_ID     "Red Hat "  // UCHAR VendorId[8]
#define VIRTIO_BLK_PRODUCT_ID    "VirtIO"    // UCHAR ProductId[16]
#define VIRTIO_BLK_PROD_REV_LVL  "1001"      // UCHAR ProductRevisionLevel[4]
#define VIRTIO_BLK_VEND_SPECIFIC "1001"      // UCHAR VendorSpecific[20]
#endif
```
Although the question remains as to where best to locate these macros... I think maybe elsewhere in the tree...
@vrozenfe : Any advice?

benyamin-codez:
And then:
```c
StorPortMoveMemory(..., VIRTIO_BLK_VENDOR_ID, sizeof(((INQUIRYDATA){}).VendorId));
StorPortMoveMemory(..., VIRTIO_BLK_PRODUCT_ID, sizeof(((INQUIRYDATA){}).ProductId));
StorPortMoveMemory(..., VIRTIO_BLK_PROD_REV_LVL, sizeof(((INQUIRYDATA){}).ProductRevisionLevel));
StorPortMoveMemory(..., VIRTIO_BLK_VEND_SPECIFIC, sizeof(((INQUIRYDATA){}).VendorSpecific));
```
...?

benyamin-codez:
> Something like:
>```c
> // Set Vendor Information - Observe INQUIRYDATA struct length limits
> #if !defined(VIRTIO_BLK_VENDOR_ID)
> #define VIRTIO_BLK_VENDOR_ID     "Red Hat "  // UCHAR VendorId[8]
> ...
> #endif
> ```

So I presume these will all be `const lstr` and so null terminated. Therefore, we will likely need to trim.
In the past I've just used something much like `CopyBufferToAnsiString()` but left off adding `\0` after copy:
```c
USHORT CopyBufferAsUnterminatedString(void *_pDest, const void *_pSrc, const char delimiter, size_t _maxlength)
{
    PCHAR dst = (PCHAR)_pDest;
    PCHAR src = (PCHAR)_pSrc;
    USHORT _length = _maxlength;

    while (_length && (*src != delimiter))
    {
        *dst++ = *src++;
        --_length;
    };
    return _length;
}
```
That being the solution, we can just use `"Red Hat"` (or any other value) without the padding.
...and implement like this:
```c
#define VIRTIO_BLK_VENDOR_ID     "Red Hat"  // UCHAR VendorId[8]
CHAR trimmed_vendor_id[8] = {0};
CopyBufferAsUnterminatedString(&trimmed_vendor_id, &VIRTIO_BLK_VENDOR_ID, '\0', sizeof(((INQUIRYDATA){0}).VendorId));
StorPortMoveMemory(&adaptExt->inquiry_data.VendorId, trimmed_vendor_id, sizeof(((INQUIRYDATA){0}).VendorId));
```

benyamin-codez:
@MartinCHarvey-Nutanix 

The above will safely truncate the string rather than barf... ¯\\\_(ツ)\_/¯

@vrozenfe @YanVugenfirer 

If happy with that approach (you as well as Martin), let me know and I'll split it off to a new PR and let this one be... 8^d

benyamin-codez:
> However, someone needs to have a look through the docs somewhere to find out whether the the string should be null terminated, optionally null terminated, or padded with spaces to the end of the (pretty small) amount of space provided.

According to the doco they are padded with ASCII blanks equiv. to `0x20` or "space"...

Only the verbage for `ProductId` states:
> "The data shall be left-aligned within this field and the unused bytes filled with ASCII blanks."

I guess it is implied for the others...

YanVugenfirer:

benyamin-codez:
@MartinCHarvey-Nutanix 

Hey Martin,

I solved this Product and Vendor inquiry data issue per d5c5a71. Have a look if you're interested.

Best regards,
Ben

viostor/virtio_stor.c
```diff
@@ -981,11 +981,11 @@ VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
                              SRB_FUNCTION(Srb),
                              adaptExt->srb_cnt,
                              adaptExt->inqueue_cnt);
-                for (USHORT i = 0; i < adaptExt->num_queues; i++)
+                for (ULONG vq_req_idx = VIRTIO_BLK_REQUEST_QUEUE_0; vq_req_idx < adaptExt->num_queues; vq_req_idx++)
                 {
```
YanVugenfirer, line 1249: I like the spirit of the change, but not the letter of the change.
While it is true that the queue are used to process requests, we never call them request queue in the code (this is not also the letter of the spec). So frankly it is a little confusing to see a new term used. I think just "VIRTIO_BLK_QUEUE_0" and "vq_idx" is better.

viostor/virtio_stor.h
```diff
@@ -80,11 +80,13 @@ typedef struct VirtIOBufferDescriptor VIO_SG, *PVIO_SG;
 #define IO_PORT_LENGTH                     0x40
 #define MAX_CPU                            256u
 #define MAX_DISCARD_SEGMENTS               256u
+#define MIN_DISCARD_SECTOR_ALIGNMENT       8
 
-#define VIRTIO_BLK_QUEUE_LAST              MAX_CPU
-
+#define VIRTIO_BLK_MSIX_VQ_OFFSET          1
+#define VIRTIO_BLK_MSIX_VQ_1_VCTR_OFFSET   0
```
YanVugenfirer: Again, I am not sure about "offset" term here.

YanVugenfirer: Even more, I would just use VIRTIO_BLK_MSIX_CONFIG_VECTOR + 1 

VIRTIO_BLK_MSIX_VQ_1_VCTR_OFFSET - this is not control queue. This is configuration notification interrupt. It is different from control queue. So please don't add confusion.

viostor/virtio_stor.c
```diff
@@ -750,35 +750,53 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
                 if (CHECKFLAG(perfData.Flags, STOR_PERF_DPC_REDIRECTION))
                 {
                     adaptExt->perfFlags |= STOR_PERF_DPC_REDIRECTION;
+
+                    if (CHECKFLAG(perfData.Flags, STOR_PERF_INTERRUPT_MESSAGE_RANGES))
+                    {
+                        adaptExt->perfFlags |= STOR_PERF_INTERRUPT_MESSAGE_RANGES;
+                        perfData.FirstRedirectionMessageNumber = VIRTIO_BLK_REQUEST_QUEUE_0 + VIRTIO_BLK_MSIX_VQ_OFFSET;
```
YanVugenfirer: Sorry, this is adding confusion to simple assignment.
`perfData.FirstRedirectionMessageNumber = VIRTIO_BLK_FIRST_MSIX_VQ_IDX;` or something with similar meaning will be better.

viostor/virtio_stor.c
```diff
@@ -750,35 +750,53 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+                            perfData.MessageTargets = adaptExt->pmsg_affinity;
+                        }
+                    }
+
+                    if (CHECKFLAG(perfData.Flags, STOR_PERF_OPTIMIZE_FOR_COMPLETION_DURING_STARTIO))
+                    {
+                        adaptExt->perfFlags |= STOR_PERF_OPTIMIZE_FOR_COMPLETION_DURING_STARTIO;
+                    }
+
+                    if (CHECKFLAG(perfData.Flags, STOR_PERF_DPC_REDIRECTION_CURRENT_CPU))
```
YanVugenfirer: Why do we need this "if", if it does nothing

viostor/virtio_stor.c
```diff
@@ -688,22 +688,38 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+        ((adaptExt->num_queues + VIRTIO_BLK_REQUEST_QUEUE_0 + VIRTIO_BLK_MSIX_VQ_OFFSET) > adaptExt->msix_vectors))
     {
-        adaptExt->num_queues = adaptExt->msix_vectors;
+        if (adaptExt->msix_vectors == 1)
+        {
+            adaptExt->num_queues = 1;
+        }
+        else
+        {
+            adaptExt->num_queues = (adaptExt->msix_vectors - VIRTIO_BLK_REQUEST_QUEUE_0 - VIRTIO_BLK_MSIX_VQ_OFFSET);
```
YanVugenfirer: why?

just use "-1" here.

viostor/virtio_stor.c
```diff
@@ -688,22 +688,38 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
-    if (adaptExt->num_queues > 1 && ((adaptExt->num_queues + 1U) > adaptExt->msix_vectors))
+
+    /* edge case when queues exceed MSI vectors */
+    if (adaptExt->num_queues > 1 &&
+        ((adaptExt->num_queues + VIRTIO_BLK_REQUEST_QUEUE_0 + VIRTIO_BLK_MSIX_VQ_OFFSET) > adaptExt->msix_vectors))
     {
-        adaptExt->num_queues = adaptExt->msix_vectors;
+        if (adaptExt->msix_vectors == 1)
+        {
+            adaptExt->num_queues = 1;
```
YanVugenfirer, line 692: `adaptExt->msix_one_vector = TRUE;` do it here for clarity.

viostor/virtio_stor.c
```diff
@@ -688,22 +688,38 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+        {
+            /* if we don't have enough vectors, use one for all queues */
+            adaptExt->msix_one_vector = TRUE;
+        }
     }
     else
     {
-        /* if we don't have enough vectors, use one for all queues */
-        RhelDbgPrint(TRACE_LEVEL_INFORMATION, (" Using one MSI vector for all queues\n"));
-        adaptExt->msix_one_vector = TRUE;
```
YanVugenfirer, line 711: Are you removing adaptExt->msix_one_vector = TRUE; for dump mode?

viostor/virtio_stor.c
```diff
@@ -688,22 +688,38 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+        /* initialize queues with no MSI interrupts */
+        adaptExt->msix_enabled = FALSE;
```
YanVugenfirer, line 705: Let's simplify all the above block of queue/vector decision (pseudo code below)

If(adaptExt->dump_mode) {
 adaptExt->msix_one_vector = TRUE;
// what ever else needed for this case
} 
else if (adaptExt->msix_vectors == 1) {
   // Corner case when for some reason we have only one vector, in general this is misconfiguration in QEMU or some implementation fo virt-blk on the host/HW that is deficient or Windows OS gone crazy
 adaptExt->msix_one_vector = TRUE;
 adaptExt->num_queues = 1;
 adaptExt->first_vq_message = 0; // You don't have this now in the code, but instead use "offset" constant that is not true for this corner case. And use it, only in case  `adaptExt->msix_one_vector `= TRUE is not sufficient later.
} else {
  // all the rest, including the case where we have more queues than vectors.
}

viostor/virtio_stor.c
```diff
@@ -674,6 +674,95 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
     adaptExt->poolOffset = 0;
     adaptExt->stopped = FALSE;
 
+    if ((!adaptExt->dump_mode) && (adaptExt->num_queues > 1) && (adaptExt->perfFlags == 0))
```
YanVugenfirer: I think this is `adaptExt->num_queues > 1` redundant. No need to make special case for one queue if we have two MSI vectors. Using one queue with MSI shouldn't be special. My guess is that you see code difference only for historical reasons, as multi-queue was added after the driver code was stable.

YanVugenfirer: The better question here is not of we have move than 1 queue, but rather if we have more than 1 vector.

viostor/virtio_stor.c
```diff
@@ -681,14 +681,25 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
-                        " Perf Version = 0x%x, Flags = 0x%x, ConcurrentChannels = %d, FirstRedirectionMessageNumber = "
-                        "%d,LastRedirectionMessageNumber = %d\n",
-                        perfData.Version,
-                        perfData.Flags,
-                        perfData.ConcurrentChannels,
-                        perfData.FirstRedirectionMessageNumber,
-                        perfData.LastRedirectionMessageNumber);
+        if (status == STOR_STATUS_SUCCESS)
+        {
+            RhelDbgPrint(TRACE_LEVEL_INFORMATION,
```
YanVugenfirer, line 717: Sorry for opening pandora box, but the trace level usage is not so good here. 
I would suggest to always printout config info on the lowest of the levels. 
I mean, all the driver config and init printouts.

benyamin-codez:
@YanVugenfirer 

Thanks for the very prompt review, Yan. It's good to get some feedback.

I had meant to add an explanatory memorandum, but I had to go for a minor surgery and while I was out we had a major power outage for over six hours, including a UPS failure at our primary site - due to diesel bug...! So I'm just coming back to it now.

Perhaps let me get that memorandum out before addressing the points you raised in your review. Hopefully we can settle on something mutually satisfactory. I'm also minded to get the interrupt / locking PRs out ASAP too. So I'll do my best to balance the output. For now I'll mark this back as a draft and I'll do the same with the new PRs.

Best regards,
Ben

cc: @kevmw @vrozenfe

YanVugenfirer:
@benyamin-codez get well soon!

viostor/virtio_stor.c
```diff
@@ -981,11 +981,11 @@ VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
                              SRB_FUNCTION(Srb),
                              adaptExt->srb_cnt,
                              adaptExt->inqueue_cnt);
-                for (USHORT i = 0; i < adaptExt->num_queues; i++)
+                for (ULONG vq_req_idx = VIRTIO_BLK_REQUEST_QUEUE_0; vq_req_idx < adaptExt->num_queues; vq_req_idx++)
```
kevmw: `VIRTIO_BLK_QUEUE_0` doesn't exist yet in this commit, so this won't even build. It is only added by the next commit.

I don't really see the point of defining a constant for 0 as a lower bound either. I feel this change makes the code more complicated to read rather than easier; with the current code I can tell at a glance what it does without having to look up a define. 0 to num_something is the normal and expected thing.

So I'd prefer to just change the type in this line and otherwise leave it alone.

viostor/virtio_stor.h
kevmw: I don't think there is a point in defining constants that aren't actually constant because they depend on a condition. These constants shouldn't be used in many places, but we should have helpers to convert from QueueNumber to MessageId, and this helper can use comments to explain the open-coded constant that will be a lot clearer than abbreviations like VQ_1_VCTR.

If we do keep these constants, they need a comment at their definition like there is in the commit message, because how they relate to `msix_one_vector` is really not obvious from the name.

viostor/virtio_stor.h
```diff
@@ -80,11 +80,13 @@ typedef struct VirtIOBufferDescriptor VIO_SG, *PVIO_SG;
...
 #define VIRTIO_BLK_MSIX_CONFIG_VECTOR      0
-#define MIN_DISCARD_SECTOR_ALIGNMENT       8
+#define VIRTIO_BLK_REQUEST_QUEUE_0         0
```
kevmw: As mentioned in the previous commit, this is just the lower bound of a normal zero-based array. I don't think there is a reason why it would ever be something else than 0, and we don't have such a constant for other arrays either.

viostor/virtio_stor.c
```diff
@@ -816,27 +810,22 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
         }
         if ((adaptExt->dpc != NULL) && (adaptExt->dpc_ok == FALSE))
         {
-            ret = StorPortEnablePassiveInitialization(DeviceExtension, VirtIoPassiveInitializeRoutine);
+            if (!StorPortEnablePassiveInitialization(DeviceExtension, VirtIoPassiveInitializeRoutine))
+            {
+                virtio_add_status(&adaptExt->vdev, VIRTIO_CONFIG_S_FAILED);
+                return FALSE;
```
kevmw, line 937: Missing closing brace, this can't compile.

viostor/virtio_stor.c
kevmw: It's incorrect anyway for `msix_one_vector` (which is preexisting, but the clever looking constants only serve to hide the problem).

viostor/virtio_stor.c
```diff
@@ -750,35 +750,53 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
                 if (CHECKFLAG(perfData.Flags, STOR_PERF_DPC_REDIRECTION))
                 {
                     adaptExt->perfFlags |= STOR_PERF_DPC_REDIRECTION;
+
+                    if (CHECKFLAG(perfData.Flags, STOR_PERF_INTERRUPT_MESSAGE_RANGES))
```
kevmw: Not sure where to comment best about the change to check the prerequisite flags, so I'll just pick the first line.

We got `perfData.Flags` from StorPort, and if already a prerequisite for a flag weren't fulfilled, the flag wouldn't be set in the first place. So I think this only increases the nesting level for no gain.

viostor/virtio_stor.c
```diff
@@ -688,22 +688,38 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
     }
 
     RhelDbgPrint(TRACE_LEVEL_INFORMATION, " Queues %d msix_vectors %d\n", adaptExt->num_queues, adaptExt->msix_vectors);
-    if (adaptExt->num_queues > 1 && ((adaptExt->num_queues + 1U) > adaptExt->msix_vectors))
+
+    /* edge case when queues exceed MSI vectors */
+    if (adaptExt->num_queues > 1 &&
+        ((adaptExt->num_queues + VIRTIO_BLK_REQUEST_QUEUE_0 + VIRTIO_BLK_MSIX_VQ_OFFSET) > adaptExt->msix_vectors))
```
kevmw: If `VIRTIO_BLK_REQUEST_QUEUE_0` were actually a useful non-zero constant, I think you would have to subtract it rather than add it here (if `num_queues` were just the maximum queue number + 1) or leave it out completely (if `num_queues` is still the number of queues). Addition doesn't seem to make sense.

But overall this is just another example how these constants make relatively simple code hard to read and reason about.

viostor/virtio_stor.c
benyamin-codez: I'll try to explain my self better in a little while, but seeing as both you and Yan have mentioned it, I was essentially looking to consolidate the common code base  with `vioscsi`, where this would be `VIRTIO_SCSI_REQUEST_QUEUE_0` and a value of 2, these could then (at a later date) be reduced to `VIRTIO_STOR_REQUEST_QUEUE_0` (they are both StorPort drivers after all) and the value set with a pre-processor definition at compile time. This would reduce the development effort required to maintain the common code between two StorPort drivers, and increase the reliability and serviceability of the project.

viostor/virtio_stor.h
benyamin-codez: It's a mnemonic that might assist a new contributor to understand the relationship and the difference between `msix_one_vector` and normal operation.

viostor/virtio_stor.c
benyamin-codez: My bad.
I wouldn't presume all commits will compile.
I've just tried to split them as best as I can and as quickly as I can as my time permits.
That being said, I can do some fix-ups if necessary.

viostor/virtio_stor.c
```diff
@@ -838,16 +838,32 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
                 virtio_add_status(&adaptExt->vdev, VIRTIO_CONFIG_S_FAILED);
+                RhelDbgPrint(TRACE_LEVEL_FATAL, " StorPortEnablePassiveInitialization() FAILED..!!!\n");
                 return FALSE;
+            }
+        }
+        else
+        {
+            /* VioStorPoolAlloc() ran out of memory and reported as such */
+            virtio_add_status(&adaptExt->vdev, VIRTIO_CONFIG_S_FAILED);
+            return FALSE;
```
kevmw, line 944: Before this commit, the driver would simply go on without using DPCs. Now it refuses to work at all.

You could convince me that this is a good change if it meant that we can remove the non-DPC paths in the interrupt handlers. Calling the completion code in interrupt context in cases that we probably don't test much has made me feel a bit uneasy. But it seems in `dump_mode` we'd still have a non-DPC path, so I'm less sure.

viostor/virtio_stor.h
benyamin-codez: Both good points. Let me have a think about it.

viostor/virtio_stor.h
benyamin-codez: Thought you meant VIRTIO_BLK_MSIX_CONFIG_VECTOR, but my thinking for   VIRTIO_BLK_REQUEST_QUEUE_0 was as my previous comment.

viostor/virtio_stor.c
```diff
@@ -674,6 +674,95 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+            if (CHECKFLAG(perfData.Flags, STOR_PERF_DPC_REDIRECTION))
+            {
+                adaptExt->perfFlags |= STOR_PERF_DPC_REDIRECTION;
+
+                if (CHECKFLAG(perfData.Flags, STOR_PERF_INTERRUPT_MESSAGE_RANGES))
+                {
+                    adaptExt->perfFlags |= STOR_PERF_INTERRUPT_MESSAGE_RANGES;
+                    perfData.FirstRedirectionMessageNumber = VIRTIO_BLK_REQUEST_QUEUE_0 + VIRTIO_BLK_MSIX_VQ_OFFSET;
+                    perfData.LastRedirectionMessageNumber = perfData.FirstRedirectionMessageNumber +
+                                                            adaptExt->num_queues - VIRTIO_BLK_REQUEST_QUEUE_0 -
```
kevmw: `num_queues` is not final yet here, it may still be reduced further down in the code that you didn't move.

If you have fewer vectors than num_queues + 1, this means that `perfData.LastRedirectionMessageNumber` will be higher than the last available vector and `StorPortInitializePerfOpts` will fail.

viostor/virtio_stor.c
```diff
@@ -681,14 +681,25 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
 
-        RhelDbgPrint(TRACE_LEVEL_VERBOSE,
-                        " Perf Version = 0x%x, Flags = 0x%x, ConcurrentChannels = %d, FirstRedirectionMessageNumber = "
-                        "%d,LastRedirectionMessageNumber = %d\n",
-                        perfData.Version,
-                        perfData.Flags,
-                        perfData.ConcurrentChannels,
-                        perfData.FirstRedirectionMessageNumber,
-                        perfData.LastRedirectionMessageNumber);
+        if (status == STOR_STATUS_SUCCESS)
```
kevmw, line 715: The next thing after this block is already an `if` with the same condition. So they should probably be unified.

viostor/virtio_stor.c
benyamin-codez: It does looks that way doesn't it. I'll have to check my tests, as that's not what happened AFAICT...

viostor/virtio_stor.h
```diff
@@ -95,6 +95,12 @@ typedef struct VirtIOBufferDescriptor VIO_SG, *PVIO_SG;
 
 #define VIOBLK_POOL_TAG                    'BoiV'
 
+/* Pad the following with spaces (ASCII '\20') */
+#define VIRTIO_BLK_VENDOR_ID               "Red Hat "             // UCHAR VendorId[8]
+#define VIRTIO_BLK_PRODUCT_ID              "VirtIO          "     // UCHAR ProductId[16]
+#define VIRTIO_BLK_PROD_REV_LVL            "1001"                 // UCHAR ProductRevisionLevel[4]
+#define VIRTIO_BLK_VEND_SPECIFIC           "1001                " // UCHAR VendorSpecific[20]
```
kevmw, line 110: Each of these are only used once. Hiding them behind constants will only make the code harder to read.

viostor/virtio_stor.c
benyamin-codez: We use this ladder to enable / disable flags.
StorPort reports what it can do and then we set how we want it to behave.

viostor/virtio_stor.h
benyamin-codez: It permits other vendors to set them correctly, conveniently in the header.
They were also previously improperly padded.

viostor/virtio_stor.h
kevmw: I think this part would be good to be rebased on top of #1503 because it does add such helpers (they don't do the full thing yet, so they look a bit simplistic at the moment, but they prepare for the full thing at least).

benyamin-codez:
@YanVugenfirer @kevmw 

Apologies Gents,
I need to get some shut-eye.
I'm a bit sore from the surgery.
Lots of great feedback. Thank you.
I'll pick this up on the other side.

Best regards,
Ben

viostor/virtio_stor.c
kevmw: Compiling the same code twice with different definitions comes with its own problems in maintainability. I think in that case it might be better to do things at runtime. But we can cross that bridge when (well, if) we get there. I don't think adding things to viostor-only code now will help us much then, but it will be a distraction for the time being.

viostor/virtio_stor.c
kevmw: It's not obvious that it failed if you don't look for it in the trace. I hit the same case last week and didn't notice for a while (`msix_one_vector` causes it to fail in master). The way I found it out was because the device was operating in a single queue mode while I had `num_queues > 1` - StorPort would never suggest a non-zero `MessageNumber` when the call failed.

viostor/virtio_stor.c
benyamin-codez: I've disabled DBG in the WIP same as `vioscsi`, so it's either DBG or ETW (default).

Since I worked around the WPP bug for the built-in `%!FUNC!` format string for functions forced inline, there is no longer any discernible overhead for running ETW until you actually monitor the trace. 

These are the macros I set:
```c
// For DBG mode
#if !defined(TRACE_LEVEL_NONE)
#define TRACE_LEVEL_NONE        0
#define TRACE_LEVEL_CRITICAL    1
#define TRACE_LEVEL_FATAL       1
#define TRACE_LEVEL_ERROR       2
#define TRACE_LEVEL_WARNING     3
#define TRACE_LEVEL_INFORMATION 4
#define TRACE_LEVEL_VERBOSE     5
#define TRACE_ENTER_EXIT        6
#define TRACE_VQ                7
#define TRACE_PERF              8
#define TRACE_LOCKS             9
#define TRACE_UNMAP             10
#define TRACE_WRITE_PATH        11
#define TRACE_REGISTRY          12
#define TRACE_INTERRUPT         13
#define TRACE_MSIX_CPU_AFFINITY 14
#define TRACE_DPC               15
#define TRACE_VQ_SRB_ID         16
#define TRACE_VQ_PB_SRB_ID      17
#define TRACE_FUA               18
#define TRACE_FLUSH             19
#define TRACE_GUEST_FEATURES    20
#define TRACE_ALL               31
#endif
....
// For ETW mode
#define WPP_CONTROL_GUIDS \
    WPP_DEFINE_CONTROL_GUID(VioStorTraceGuid,(B17FA150,8C45,482E,9EB8,29611A862BF3), \
        WPP_DEFINE_BIT(TRACE_LEVEL_NONE)               /* bit  0 = 0x00000001 */     \
        WPP_DEFINE_BIT(TRACE_LEVEL_FATAL)              /* bit  1 = 0x00000002 */     \
        WPP_DEFINE_BIT(TRACE_LEVEL_ERROR)              /* bit  2 = 0x00000004 */     \
        WPP_DEFINE_BIT(TRACE_LEVEL_WARNING)            /* bit  3 = 0x00000008 */     \
        WPP_DEFINE_BIT(TRACE_LEVEL_INFORMATION)        /* bit  4 = 0x00000010 */     \
        WPP_DEFINE_BIT(TRACE_LEVEL_VERBOSE)            /* bit  5 = 0x00000020 */     \
        WPP_DEFINE_BIT(TRACE_ENTER_EXIT)               /* bit  6 = 0x00000040 */     \
        WPP_DEFINE_BIT(TRACE_VQ)                       /* bit  7 = 0x00000080 */     \
        WPP_DEFINE_BIT(TRACE_PERF)                     /* bit  8 = 0x00000100 */     \
        WPP_DEFINE_BIT(TRACE_LOCKS)                    /* bit  9 = 0x00000200 */     \
        WPP_DEFINE_BIT(TRACE_UNMAP)                    /* bit 10 = 0x00000400 */     \
        WPP_DEFINE_BIT(TRACE_WRITE_PATH)               /* bit 11 = 0x00000800 */     \
        WPP_DEFINE_BIT(TRACE_REGISTRY)                 /* bit 12 = 0x00001000 */     \
        WPP_DEFINE_BIT(TRACE_INTERRUPT)                /* bit 13 = 0x00002000 */     \
        WPP_DEFINE_BIT(TRACE_MSIX_CPU_AFFINITY)        /* bit 14 = 0x00004000 */     \
        WPP_DEFINE_BIT(TRACE_DPC)                      /* bit 15 = 0x00008000 */     \
        WPP_DEFINE_BIT(TRACE_VQ_SRB_ID)                /* bit 16 = 0x00010000 */     \
        WPP_DEFINE_BIT(TRACE_VQ_PB_SRB_ID)             /* bit 17 = 0x00020000 */     \
        WPP_DEFINE_BIT(TRACE_FUA)                      /* bit 18 = 0x00040000 */     \
        WPP_DEFINE_BIT(TRACE_FLUSH)                    /* bit 20 = 0x00080000 */     \
        WPP_DEFINE_BIT(TRACE_GUEST_FEATURES)           /* bit 21 = 0x00100000 */     \
        WPP_DEFINE_BIT(TRACE_ALL)                      /* bit 22 = 0x00200000 */     \
        )
```
I'm open to suggestions for change here, noting more granularity is preferred.  We can go up to 32 bits...

The `TRACE_VQ_*SRB_ID` bits might disappear if we remove those parts just recently merged, although I see some value to retaining that functionality.

I also set `nViostorDebugLevel = TRACE_ALL` so when DBG is enabled you get everything. With ETW, the default is already all bits, so you have to otherwise specify your desired flags when you run the logger.

I'll raise a separate PR splitting the tracing components from my WIP fairly soon methinks...
...and then another new PR for all_paths to replace the existing cold_path+hot_path PR for `vioscsi` (PR #1228)...
I think I'll rip out that `RUN_UNCHECKED` part from the `vioscsi` PR too, as the macro exclusions are pretty ugly and I'm fairly confident now that we don't have any WPP performance issues remaining.

Then I think the follow up PR is to set all the `TRACE_LEVEL_*` and `TRACE_*` parameters on all `RhelDbgPrint()` and `RhelDbgPrintInline()` calls accordingly. For the ones in this PR, I had mostly `TRACE_PERF`. I think the remainder would likely be new `TRACE_*` definitions: `TRACE_INIT` and `TRACE_INQUIRYDATA`.

I think the introduction of `TRACE_INIT` and the plan above should address your concerns for anything that is otherwise `TRACE_LEVEL_INFORMATION` or `TRACE_LEVEL_VERBOSE` in this PR, yes?

viostor/virtio_stor.c
benyamin-codez: Yep, `vq_idx` it shall be.

Still considering options for `VIRTIO_BLK_REQUEST_QUEUE_0`...

From the long-term perspective of consolidating `VIRTIO_BLK_REQUEST_QUEUE_0` and `VIRTIO_SCSI_REQUEST_QUEUE_0` from `vioscsi` in a common code-base, perhaps as `VIRTIO_STOR_REQUEST_QUEUE_0`, or similar, and set the value with _Preprocessor Definitions_ at compile time.

The distinction with `*_REQUEST_*` here is due to the `*_CONTROL_QUEUE` and `*_EVENTS_QUEUE` in `vioscsi`, but we could also say that `*_REQUEST_*` refers to a StorPort request queue too. So whilst not VirtIO verbiage, there is still some holdover from SCSIPort on StorPort nomenclature.

I think it is perhaps prudent to use `VIRTIO_STOR_REQUEST_QUEUE_0` in `viostor` without the intervening period using `VIRTIO_BLK_REQUEST_QUEUE_0`... This signifies that this is less about `virtio_blk` and more about StorPort.

viostor/virtio_stor.c
benyamin-codez: For completeness, plus we use this ladder as toggles to enable/disable the flags during development.

viostor/virtio_stor.h
benyamin-codez: If the notion of the relationship between the first VQ and it's corresponding MSI-X vector is made clear in PR #1503, especially the distinction around `msix_one_vector` then these particular macro constants are probably of little value.

Having said that, they were somewhat useful during debugging to see what was going on with `msix_one_vector`, which is why it ended up having it's own.

viostor/virtio_stor.c
benyamin-codez: Kevin, We typically wouldn't bother getting the `perfData` for `msix_one_vector` because `adaptExt->num_queues > 1` is FALSE, noting that this perhaps should be under a conditional for `adapExt->msix_vectors > 1` instead. I need to review the ordering as you have raised elsewhere.

Yan, that alternative, i.e. `VIRTIO_BLK_FIRST_MSIX_VQ_IDX`, could work too.
Or even `VIRTIO_STOR_FIRST_MSIX_VQ_IDX`...
Or perhaps `VIRTIO_STOR_VQ_0_MSIX_IDX` is best...

Having said that, I need to have a better look at Kevin's PR #1503 and go from there I think.

viostor/virtio_stor.c
benyamin-codez: A new contributor or auditor may not understand why we use `-1`.
This would give a hint, leading to a table of related constants.

viostor/virtio_stor.h
benyamin-codez: > If we do keep these constants, they need a comment at their definition like there is in the commit message, because how they relate to `msix_one_vector` is really not obvious from the name.

I think this would be prudent.
Later we could also say in such comments if a constant applies  to `viostor`, `vioscsi` or both.

viostor/virtio_stor.c
benyamin-codez: Yes, I think this ladder could do with some consolidation.
This was probably a good interim step...

viostor/virtio_stor.c
benyamin-codez: I disabled MSI-X for dump mode.
iirc, it usually runs in single-thread mode without any locks.
Otherwise, any locks called will be `InterruptLock` type.

I'll do a check with the INITIATED_MANUAL_CRASH method.
Is that sufficient, or do you need a full boot-time crash check too?

viostor/virtio_stor.c
benyamin-codez: I was keeping them separate in case I wrapped this in `#ifdef EVENT_TRACING`...

Haven't decided yet, but probably leaning towards consolidation.

viostor/virtio_stor.c
benyamin-codez: Yeah, I'll do some work on this.

viostor/virtio_stor.c
benyamin-codez: I'll give vector vs queue some more thought. I need to (re)consider the (re)ordering too...

viostor/virtio_stor.c
benyamin-codez: I think I have an OOB issue when `adaptExt->msix_vectors` is 0 here, but it continues regardless.
I will fix it up during consolidation works...

viostor/virtio_stor.c
benyamin-codez: If you have the subject matter knowledge, then `adaptExt->num_queues + 1` makes sense, but to the uninitiated, they are probably left wondering what is going on.

A different name for `VIRTIO_BLK_MSIX_VQ_OFFSET` might help.

For `vioscsi` the `VIRTIO_*_REQUEST_QUEUE_0` is 2...
...so admittedly this really only makes sense when considering this from a consolidation perspective.

The logic here is: are the total number of queues (including any CONTROL and EVENTS queues - as the case may be for `vioscsi`) PLUS the MSI-X config vector MORE THAN than the total number of available MSI-X vectors.

Wrapping the elements in parentheses might help the uninitiated further, e.g. :

```c
        (((adaptExt->num_queues + VIRTIO_BLK_REQUEST_QUEUE_0) + VIRTIO_BLK_MSIX_VQ_OFFSET) > adaptExt->msix_vectors))
```

viostor/virtio_stor.c
benyamin-codez: Once the common code is identified, I think it would only be maintained in `viostor`. Much like we compile the VirtIO sub-project now, as a prerequisite, we would then have a "bare bones" `vioscsi` with the additional functionality. The next step would be to migrate the `vioscsi` elements into a fully consolidated StorPort driver. 

I know I am speaking at some length about consolidation, when there has not been any buy-in from the project despite asking a couple of times. Bugs aside, my team identified lack of effective tracing and the two separate StorPort code-bases amongst the highest risks to RAS (admittedly it provides a low-level of increased availiability). So I'm only getting support to spend time on the project if I can address these concerns in some meaningful way, and whilst that support isn't required, I do prefer it.

It's not all bad news, the first step is already done: the driver has been effectively migrated from SCSIPort to StorPort, albeit as a SAS adapter (not that this is an issue).

viostor/virtio_stor.c
benyamin-codez: That's interesting... I'll check it out...

viostor/virtio_stor.c
benyamin-codez: I had meant to also say I think you are correct regarding the historical cause.

As long as it tests out when:

 * `num_queues >= msix_vectors` (by adjustment to `num_queues`)
 * `msix_vectors >= num_queues` (by adjustment to Message ranges as necessary)
 * `msix_one_vector = TRUE` provided `num_queues > 0` (we only use one VQ)

...all shall be well...

viostor/virtio_stor.c
kevmw: Right. What I'm saying is that the same prerequisites apply to StorPort itself, so it won't report that it can do `STOR_PERF_INTERRUPT_MESSAGE_RANGES` if it doesn't also support `STOR_PERF_DPC_REDIRECTION` etc. So the current code without the nesting is perfectly fine.

If you want to have the dependency mentioned explicitly in the code, you could add assertions that the required flag is set, too.

benyamin-codez:
@YanVugenfirer @kevmw 

Yan and Kevin,

Just a heads up that I haven't yet addressed all of the items you raised in each of your respective reviews. I'll go back through your review comments and address what I can in a couple of hours - just got a few things to take care of.

I've tested most scenarios but still need to do crash dump driver test and multi-HBA too.

Also the outstanding issue of configuration change notifications...

SDV was good.

Please find some pics of traces for various vector counts below.

Best regards,
Ben

cc: @vrozenfe 

Normal MSI-X count (5 vectors on 4 vCPUs):
msix_vectors_full
...
MSI-X vector count limited to 3:
msix_vectors_is_3
...
MSI-X vector count limited to 1:
no_msix_cfg_vector
...
No MSI-X:
no_msix_enable

viostor/virtio_stor.c
benyamin-codez: Ok, fixed this one.

viostor/virtio_stor.c
benyamin-codez: No longer using these conditionals.

viostor/virtio_stor.c
benyamin-codez: If we presume the catch-all that Storport supports all flags on a given hardware platform, in a hypothetical situation where we set `STOR_PERF_ADV_CONFIG_LOCALITY` without setting `STOR_PERF_INTERRUPT_MESSAGE_RANGES`, perhaps Storport will fail `StorPortInitializePerfOpts()` because we overwrite the flags with `perfData.Flags = adaptExt->perfFlags` rather than doing a bitwise AND operation. So we could just "fix" that part...
 
However, I'm inclined to keep it as proposed because it makes it obvious what the dependencies are in the code if any future dev is considering a change to the submitted flags (whether break-fix or experimental).

What's maybe not so obvious is `adaptExt->perfFlags = 0` until we touch it here. I'm minded to set this explicitly to make it clear that we are populating this bit field with these flags before applying it to `perfData.Flags` and submitting it to Storport.

viostor/virtio_stor.c
benyamin-codez: Should be fixed.

viostor/virtio_stor.c
benyamin-codez: I've removed this and will address in future work.

PR #1304, opened 2025-02-17
[viostor] Zero-out adaptExt in VirtIoFindAdapter()

benyamin-codez:
Uses `RtlZeroMemory()` to wipe the `adaptExt` struct in `VirtIoFindAdapter()`.

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -258,6 +258,7 @@ VirtIoFindAdapter(IN PVOID DeviceExtension,
     UNREFERENCED_PARAMETER(Again);
 
     adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
+    RtlZeroMemory(adaptExt, sizeof(ADAPTER_EXTENSION));
 
     adaptExt->system_io_bus_number = ConfigInfo->SystemIoBusNumber;
     adaptExt->slot_number = ConfigInfo->SlotNumber;


benyamin-codez:
Only a `Flush Test` failure...

This one should probably merge ASAP.

YanVugenfirer:

kevmw:
The commit message ought to answer the question why this needs to be done - what is does is obvious enough anyway.

Specifically, StorPort seems to already zero out adaptExt before calling VirtIoFindAdapter() for the first time. Therefore, this should only make a difference on subsequent calls. But on subsequent calls, how don't we explicitly set a new value for everything that was set the first time around?

If this fixes a bug, I'd be curious what the details are. If it is just out of an abundance of caution, fair enough, but you should state so explicitly.

benyamin-codez:
> The commit message ought to answer the question why this needs to be done - what is does is obvious enough anyway.

It probably wasn't the right forum to raise all the reasons... Yes, out of an abundance of caution.

> Specifically, StorPort seems to already zero out adaptExt before calling VirtIoFindAdapter() for the first time. Therefore, this should only make a difference on subsequent calls. But on subsequent calls, how don't we explicitly set a new value for everything that was set the first time around?

There are some disparities for different targets and NTDDI versions as to the behaviour. We also don't make use of the `Again` parameter to reuse extension values. My thoughts were that it was best to zero out the device extension explicitly and create exceptions for certain restart conditions if required. AFAIK, we would init from scratch in most cases anyway...

This was also a duplication of the `vioscsi` behaviour.

kevmw:
Ok. Please update the commit message accordingly, documenting the motivation for the change is as important as the actual code, so that we can later still understand why the change was made. If you remember specific versions where the behaviour is such that it isn't zeroed even the first time (i.e. instances where this would fix an actual bug), it's worth pointing them out, too.

Once this has a good commit message, I'll approve it.

PR #1297, opened 2025-02-17
[viostor] Introduce Registry read capability

benyamin-codez:
Introduces a Registry read capability:

1. Implemented in `VioStorReadRegistryParameter()`
2. Requires neo helper `CopyBufferToAnsiString()`
3. Adds capacity to both detect and read `DWORD` values in:
   (a) `HKLM\SYSTEM\CurrentControlSet\Services\viostor\Paramaters\Device`
   (b) `HKLM\SYSTEM\CurrentControlSet\Services\viostor\Paramaters\Device(d)` (NOT WORKING)
5. Also supports per-HBA values for `\Parameters\Device\Valuename_123` when the `\Parameters\Device(d)` key is unavailable (presently broken)
6. Adds `TRACE_REGISTRY` WPP tracing flag

Issue #907, opened 2023-04-25
viostor: cause qemu virtio_error when an uncompleted request times out


wangyan0507:
**Describe the bug**
I found an virtio_error() in QEMU when handling virtio block request with windows11 guest.

Two uncompleted blk request(req1 and reqN) has the same indirect descriptor table address. The length of indirect descriptor table in req1 is 48, but the lenth in reqN is 160. So, it cause error when handling req1 in QEMU.

The stack trace:
```
(gdb) bt
#0  virtio_error (vdev=0xaaaaad604420, fmt=0xaaaaabb54ad0 "Desc next is %u") at ../hw/virtio/virtio.c:3568
#1  0x0000aaaaab642868 in virtqueue_split_read_next_desc (vdev=0xaaaaad604420, desc=0xffffefff7d00, desc_cache=0xffffefff7d10, max=3, next=0xffffefff7cb4) at ../hw/virtio/virtio.c:1057
#2  0x0000aaaaab643c8c in virtqueue_split_pop (vq=0xaaaaad60f740, sz=240) at ../hw/virtio/virtio.c:1591
#3  0x0000aaaaab644584 in virtqueue_pop (vq=0xaaaaad60f740, sz=240) at ../hw/virtio/virtio.c:1768
#4  0x0000aaaaab5d8d3c in virtio_blk_get_request (s=0xaaaaad604420, vq=0xaaaaad60f740) at ../hw/block/virtio-blk.c:238
#5  0x0000aaaaab5da1fc in virtio_blk_handle_vq (s=0xaaaaad604420, vq=0xaaaaad60f740) at ../hw/block/virtio-blk.c:772
#6  0x0000aaaaab5da304 in virtio_blk_handle_output (vdev=0xaaaaad604420, vq=0xaaaaad60f740) at ../hw/block/virtio-blk.c:806
#7  0x0000aaaaab645b00 in virtio_queue_notify_vq (vq=0xaaaaad60f740) at ../hw/virtio/virtio.c:2249
#8  0x0000aaaaab648fb4 in virtio_queue_host_notifier_read (n=0xaaaaad60f7b4) at ../hw/virtio/virtio.c:3529
#9  0x0000aaaaab951690 in aio_dispatch_handler (ctx=0xaaaaac82f9a0, node=0xfffdb0002ea0) at ../util/aio-posix.c:369
[...]

PR #1305, opened 2025-02-17
Viostor: Introduce action-on-reset feature from vioscsi

MartinDrab:
The action-on-reset determines what the driver should do when a bus reset request arrives. It can be configured to:
- complete pending requests,
- do nothing,
- bug check.

This feature is already present in vioscsi but not in viostor.

Needs to be merged after #1297.

### DIFF
--- a/viostor/viostor.inx
+++ b/viostor/viostor.inx
@@ -76,6 +76,7 @@ ErrorControl   = %SERVICE_ERROR_NORMAL%
 ServiceBinary  = %INX_PLATFORM_DRIVERS_DIR%\viostor.sys
 LoadOrderGroup = SCSI miniport
 AddReg         = pnpsafe_pci_addreg
+AddReg         = bus_reset_addreg
 
 [scsi_inst.HW]
 AddReg         = pnpsafe_pci_addreg_msix
@@ -92,6 +93,9 @@ HKR,,TypesSupported,%REG_DWORD%,7
 HKR, "Parameters\PnpInterface", "5", %REG_DWORD%, 0x00000001
 HKR, "Parameters", "BusType", %REG_DWORD%, 0x00000001
 HKR, "Parameters", DmaRemappingCompatible,0x00010001,0
+
+[bus_reset_addreg]
+HKR, "Parameters\Device", "VirtioActionOnReset", %REG_DWORD%, 0x00000001
 
 [pnpsafe_pci_addreg_msix]
 HKR, "Interrupt Management",, 0x00000010


benyamin-codez:
:+1: 

More registry love.
Nice..!

It looks like clang-format is upset.
You can find usage for the `Tools/clang-format-helper.sh` script [here](https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/contributing.md#use-a-consistent-coding-style).
Probably line 490 wanting to look something like:
```c
VioStorReadRegistryParameter(DeviceExtension,
                             REGISTRY_ACTION_ON_RESET,
                             FIELD_OFFSET(ADAPTER_EXTENSION, action_on_reset));
```
...but other lines look close to the column boundary too, so there might be more.

_P.S.:_
Just checked and it reports lines 490 & 1213 in `virtio_stor.c` and lines 209 and 267 in `virtio_stor.h`.
That mod for line 490 above is correct.

MartinDrab:
> 👍
> 
> More registry love. Nice..!
> 
> It looks like clang-format is upset. You can find usage for the `Tools/clang-format-helper.sh` script [here](https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/contributing.md#use-a-consistent-coding-style). Probably line 490 wanting to look something like:
> 
> ```c
> VioStorReadRegistryParameter(DeviceExtension,
>                              REGISTRY_ACTION_ON_RESET,
>                              FIELD_OFFSET(ADAPTER_EXTENSION, action_on_reset));
> ```
> 
> ...but other lines look close to the column boundary too, so there might be more.
> 
> _P.S.:_ Just checked and it reports lines 490 & 1213 in `virtio_stor.c` and lines 209 and 267 in `virtio_stor.h`. That mod for line 490 above is correct.

Yeah, I expected `clang-format-check` to complain and plan to fix that when https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1297 gets merged (and I rebase above it).

lixianming19951001:
I would like to ask another question: We control the behavior of action_on_reset through the registry, but by default, it still completes all pending requests. Does this not pose a risk of memory access when the host processes requests that have already been completed?

benyamin-codez:
> I would like to ask another question: We control the behavior of action_on_reset through the registry, but by default, it still completes all pending requests. Does this not pose a risk of memory access when the host processes requests that have already been completed?

Thanks Li.

Again, Martin might know best, but `viostor` has the additional function `CompleteRequestWithStatus()` (as pointed out in my comment [here](https://github.com/virtio-win/kvm-guest-drivers-windows/issues/907#issuecomment-2628623199) - not that I'm saying here that it does need to be removed - just that it is different to `vioscsi`), which will complete the requests using the specified status. StorPort is eventually notified in `CompleteSRB()`. I would have thought that if StorPort has been notified, provided the host process makes the request via StorPort, then it should be safe.

MartinDrab:
> I would like to ask another question: We control the behavior of action_on_reset through the registry, but by default, it still completes all pending requests. Does this not pose a risk of memory access when the host processes requests that have already been completed?

Hello,

I apologize for long delay. The risk is there in case the pending requests get completed as a result of bus reset request from the OS (and the host may complete some of them at the same moment -- which means access to memory no longer belonging to the requests).

I will add commit setting the `action_on_reset` on `DoNothing` during the installation.

YanVugenfirer:
ok to test

YanVugenfirer:

YanVugenfirer:
@MartinDrab Can you please use existing registry access function in this PR? 

@vrozenfe @kostyanf14

PR #1315, opened 2025-02-26
[CI-NO-BUILD] [viostor] Increase MAX_PHYS_SEGMENTS limit

benyamin-codez:
Changes the `MAX_PHYS_SEGMENTS` limit to 1,024 equal to QEMU's `VIRTQUEUE_MAX_SIZE`.

This results in:
1. `adaptExt->max_segments` = 1024
2. `ConfigInfo->NumberOfPhysicalBreaks` = 1025
3. `ConfigInfo->MaximumTransferLength` = 4MiB (1,024 * PAGE_SIZE = 4,096KiB)

PR #1313, opened 2025-02-25
[CI-NO-BUILD] [vioscsi] Align memory allocations

benyamin-codez:
Refactors `VioScsiFindAdapter()` to:

1. Pad the uncached extension to ensure it is page-aligned
2. Update memory layout art
3. Improve relevant instrumentation
4. Adds `VIRTIO_SCSI_MSI_CONTROL_Q_OFFSET` and updates the `QUEUE_TO_MESSAGE(QueueId)` and `MESSAGE_TO_QUEUE(MessageId)` macros to use it (`vioscsi.h`)

PR #1309, opened 2025-02-21
[CI-NO-BUILD] [vioscsi] Raise NTDDI version

benyamin-codez:
Refactors `vioscsi.vcxproj` to specify:

1. `NTDDI_WINTHRESHOLD` for `Win10 Release` targets; and
2. `NTDDI_WIN10_CO` for `Win11 Release` targets.

PR #1306, opened 2025-02-18
[CI-NO-BUILD] [vioscsi] Programatically determine max_segments

benyamin-codez:
Refactors `VioScsiFindAdapter()` to:

1. Use new mnemonic `adaptExt->max_segments` to replace `adaptExt->max_physical_breaks` (also used in the `VioScsiBuildIo()` and `VioScsiReadExtendedData()` functions)
2. Enable Registry lookup for `max_segments` in neo value _"MaxPhysicalSegments"_ or the legacy value _"PhysicalBreaks"_. New value defined by `REGISTRY_MAX_PH_SEGMENTS`.
3. Also derive `max_segments` from `adaptExt->scsi_config.seg_max` and from the `adaptExt->scsi_config.max_sectors` value
4. Choose from among best candidates and check within limits
5. Enforce legacy StorPort NOPB off-by-one behaviour and provide explanation
6. Improves relevant instrumentation
7. Use new mnemonic `PHYS_SEGMENTS_LIMIT` to replace `MAX_PHYS_SEGMENTS`
8. Update `VIRTIO_MAX_SG` to `(PHYS_SEGMENTS_LIMIT + 1)`
9. Update `VIRTIO_SCSI_QUEUE_LAST` to `(MAX_CPU - VIRTIO_SCSI_REQUEST_QUEUE_0)`

Note: This commit enables the use of max_segments greater than
      `adaptExt->scsi_config.seg_max` (max = 254) up to
      `PHYS_SEGMENTS_LIMIT` (presently 512).

PR #1316, opened 2025-02-26
[CI-NO-BUILD] [vioscsi] Increase PHYS_SEGMENTS_LIMIT value

benyamin-codez:
Changes the `PHYS_SEGMENTS_LIMIT` to 1,022 equal to QEMU's `VIRTQUEUE_MAX_SIZE` less 2 (for the `CONTROL` and `EVENT` queues).

This results in:
1. `adaptExt->max_segments` = 1022
2. `ConfigInfo->NumberOfPhysicalBreaks` = 1023
3. `ConfigInfo->MaximumTransferLength` = 4,088KiB (1,022 * PAGE_SIZE = 4,088KiB)

### DIFF
--- a/vioscsi/vioscsi.h
+++ b/vioscsi/vioscsi.h
@@ -50,7 +50,7 @@ typedef struct VirtIOBufferDescriptor VIO_SG, *PVIO_SG;
 #endif
 
 #define PHYS_SEGMENTS                        32
-#define MAX_PHYS_SEGMENTS                    512
+#define PHYS_SEGMENTS_LIMIT                  1022
 #define VIOSCSI_POOL_TAG                     'SoiV'
 #define VIRTIO_MAX_SG                        (1 + 1 + MAX_PHYS_SEGMENTS + 1) // cmd + resp + (MAX_PHYS_SEGMENTS + extra_page)
 


benyamin-codez:
I wouldn't much care to attempt merging this until after the following PRs merge:

#1228 
#1293 / #1294
#1214 
#1215 
#1216 
#1306 
#1313 
#1309 

Here is a pic showing the VQ Buffer Length and SRB DataTransferLength at the tail end of a 4MiB sequential read:

vioscsi_4088K_MaxXferLen_tailend

You will observe the transfer is divided into 4,088KiB and 8KiB (4,096KiB = 4MiB) parts.

benyamin-codez:
> You will observe the transfer is divided into 4,088KiB and 8KiB (4,096KiB = 4MiB) parts.

Here we can see how `viostor` will have an advantage with 1,024 segments and aligned MaxXferLen of 4MiB,
per PR #1315.

benyamin-codez:
@vrozenfe 

From PR #1216... I though it best to continue this part here.

>> ...proper support for adaptExt->scsi_config.seg_max which IMHO is a proper way to limit the number of SG elements for DMA.
>
> The fix really needs to come via QEMU. We need it to report the backing's max_segments queue value for breaking cases such as the one in [RHEL-56722](https://issues.redhat.com/browse/RHEL-56722). So this then would be safe if not ideal. It would then be up to sys engineers or admins during implementation / commissioning to tune this if necessary or desired.
>
> There are no doubt many things that QEMU devs would need to consider when coming up with a fix. One consideration needs to be: when in non-pass-through mode with multiple backings with different max_segments queue values. I would suggest the HBA would need to be set to the lowest value.

Vadim, I think it's worthwhile raising this as a bug / feature request at [QEMU](https://gitlab.com/qemu-project/qemu/-/issues).

I could have a poke around and maybe come up with a patch, but I think perhaps the QEMU storage team would likely come up with a more timely, considered and permanent solution.

This, I imagine would be something more elegant than parsing the output of a simple `grep "" /sys/block/*/queue/max_*`, mapping the right backing, making the necessary calculations and then setting the `scsi_config.seg_max` value.

Being aware of the `VIRTQUEUE_MAX_SIZE` and other limits (and maybe prepared to reconsider them), they may also choose to optimise the recommended value based on other considerations unknown to me.

There's also the pesky `- 2`, which interferes with `viostor` (cf. PR #1315), which maybe they can trim for `virtio-blk`...

In any case, *nix hosts (and maybe *bsd hosts too) presently have a clear advantage because they do not suffer from this limitation. This is not a limitation imposed by StorPort either, but rather by this project. It would be good to get this resolved and Windows hosts back on an "even" footing with the other platforms.

JonKohler:
But that assumes that lets say 4MB transfers are even ideal, yea? Depending on the backing media/storage system, they may want to chunk that down anyhow.

In our case, a little inside baseball, our storage system presents a recommended max transfer size of 1MB, and we align segments and such based on that. Furthermore, our vhost-user-scsi aligns preallocations around that, so trying to do something like this //without// the indirect descriptor "advertisements" virtio-spec enhancement that I called out in another review would potentially be a breaking change for non-qemu-managed-datapath-folks (us and I'm sure others).

We get around that by wrapping our max breaks in our installer such that it just does the right thing for our platform (for now), but just a FYI on how others in the community use this code.

benyamin-codez:
Thanks, Jon. I appreciate the insights.

I think it's worth pointing out that this is just the max possible value and it does not need to be set this high. Whatever max_segments is set to will determine the MaxXferLen. To make the point: it could be set to just 16 segments and 64Kib transfers like way back at the beginning of StorPort. Also, this maximum will not be realised if no block request is made for the requisite amount.

My proposed NoPB-coincident changes will - in the absence of the registry override - default to using whatever is in `seg_max` (when in indirect mode). A comparison is made with a max_segments value derived from `max_sectors`, and then whichever is the lesser is chosen. Without any QEMU changes, this is likely to default to 254 segments. If I'm not mistaken, the `vioscsi` default is presently 512 (`0x200`) segments with a MaxXferLen of 2MiB. My understanding is that we start to get problems beyond this. I presume this is mostly - if not entirely - for folks using non-qemu-managed-datapaths.

Usually the MaxXferLen **should** match the stripe size of the backing, but this would depend on where the host sits in the storage stack. If you have pass-through LUNs with large stripes, this would be an advantage (provided there was sufficiently fast data paths and cache). As for stripe size, the 1MiB you appear to be using is a common value, but I have seen values ranging from as low as 64KiB and up to 8MiB. Notably, if using _BusTypeRAID_ (`0x08`) and performing striping itself, having the capacity to use large stripes would be of some utility.

Presently, the only way I'm aware of to set MaxXferLen in a StorPort compliant manner is via max_segments * PAGE_SIZE. However, in our case and imho, this is really `max_segments * split_ring_buffer_size`, which just so happens to be the same size (4KiB). A virtio-scsi PCI BAR exposes 1040 virtqueues of 8KiB each, 1024 of which are usable. We access them as a split ring, so 4KiB per transfer. The `virtio-scsi` impl. uses 2 queues for CONTROL and EVENT leaving 1022 queues for requests, whereas `virtio-blk` can use all 1024.

As for alternate strategies, I'm a little in the dark (along with most I would imagine) until Vadim releases his work on IOMMU and v3 DMA because AFAIK the DDI for these aren't open. I guess, Vadim's work might make a great deal of sense when we get to see it.

Then we have the virtio-spec issue you previously mentioned (https://github.com/oasis-tcs/virtio-spec/issues/122). It was an interesting idea and to me the real value was in the proposed negotiation between device and host as to what the maximum supported values would be. This is much better than just setting it "on paper". I would probably need to thoroughly read the mailing list to get a better understanding of the nuances and fringe cases. Having said that, aren't we effectively already doing some of this without the extra tunables and flags or updates to the standard..?

We already regularly exceed the **Queue Size** whenever max_segments  is greater than 126 or 254, e.g. the `vioscsi` defaults I mentioned above. These values appear to be based on a fallacy that a queue must be tied exclusively to a (v)CPU. So whilst the value was bumped up from 128 to 256 (in SeaBIOS) to reduce block requests for larger transfers in Linux (towards the 1MiB boundary), AFAIK there does not appear to be a rational basis for this limit - at least not anymore. Windows never had the kernel limit that Linux had up until about 7 years ago [`BUG_ON(total_sg > vq->vring.num)`] and the SeaBIOS change was about 5 years back.

Anyway, I'm probably not saying anything you don't already know... 8^d

I'm currently considering wording for the QEMU bug report and collating some info for that.
I'll be sure to copy you so that so you can add anything relevant to the report wrt your kit.

However, again I think it important to stress that having a MaxXferLen in `vioscsi` that you maybe cannot utilise doesn't mean that it will stop your kit from working. You just need to ensure you set max_segments to 256  - and perhaps this is what you do with your installer already..?

I'm caught up with some competing priorities over the next few days, so let me have a think about this along the way...
Having some time away is also usually helpful... 8^D

benyamin-codez:
A `virtio-scsi` PCI BAR exposes 1040 virtqueues, each with a default _Queue Size_ of 8KiB.

Per the standard, this will result in the following allocation:

| Virtqueue Part | Formula | Size |
| :--- | :--- | :---: |
| Descriptor Table | 16 * (Queue Size) | 131,072 B (128 KiB) |
| Available Ring | 6 + 2 * (Queue Size) | 16,390 B |
| Used Ring | 6 + 8 * (Queue Size) | 65,542 B |
| Total Size | | 213,004 B |

Everyone is very kind here - I'm very rarely pulled up for comments made in error.
Better if I am..! 8^D
I can appreciate everyone is busy though.

benyamin-codez:
However, it looks like `virtio_query_queue_allocation()` returns a _Queue Size_ of 256 bytes during allocation queries...
...which results in the following allocation:

| Virtqueue Part | Formula | Size |
| :--- | :--- | :---: |
| Descriptor Table | 16 * (Queue Size) | 4,096 B (4 KiB) |
| Available Ring | 6 + 2 * (Queue Size) | 518 B |
| Used Ring | 6 + 8 * (Queue Size) | 2,054 B |
| Total Size | | 6,668 B |
| Aligned Value | | 8,192 B (8 KiB) |

So I presume my trace is using an incorrect VIRTIO_PCI_COMMON_CFG_CAP struct... Checking...
Anyone feeling dizzy...?

YanVugenfirer:

PR #1327, opened 2025-03-12
[RHEL-82494]netkvm: ensure poll handler scheduling on RSS queue change

ybendito:
Jira link: https://issues.redhat.com/browse/RHEL-82494 Raise IRQL when triggering other poll handler due to RSS queue miss, this effectively prevents synchronous call to second poll handler and prevents deadlock if both poll handlers have the same affinity.


YanVugenfirer (approved):

ybendito:
Server2025: 2c_Mini6RSSSendRecv BSOD from NDProt630:
CNDTSimpleSendCommMgr::WaitForSendsToComplete: Miniport is probably not send completing pending sends."
Checked the netkvm driver - there is one suck packet in SendQueue (lockless queue), i.e. this is https://issues.redhat.com/browse/RHEL-22354

PR #1329, opened 2025-03-15
[RHEL-83609]netkvm: always report registration offload info

ybendito:
Jira link: https://issues.redhat.com/browse/RHEL-83609

Never skip registration offload info even if we do not have some of or even any offload features.


ybendito:
Win10_22H2 RSS failed, asserion of
NDPROT630 "pCurrentNetBufferList" 			[net\kit\legacycontent\ndistest65\commengine\legacy\simplesendcommmanager.cpp @ 896]
Know failure, we assume test driver problem, rerunning

PR #1330, opened 2025-03-19
viogpu: Allocate more buffers than the queue capacity

akihikodaki:
I occasionally saw viogpu crashes the whole system because of AllocCmd()
returning NULL. viogpu allocates buffers according to the queue size and
while it works fine in most cases, the specification says:
> Once the ring is full of driver descriptors, the driver stops sending
> new requests and waits for the device to start processing descriptors
> and to write out some used descriptors before making new driver
> descriptors available.

While waiting for the device in such a scenario, the driver will need to
hold buffers more than the queue size.

Allocate a buffer if we need more than the queue size, but free it once
it becomes unnecessary so that we will not keep unbound amount of memory
indefinitely.

### DIFF
--- a/viogpu/common/viogpu_queue.cpp
+++ b/viogpu/common/viogpu_queue.cpp
@@ -594,6 +594,8 @@ BOOLEAN VioGpuBuf::Init(_In_ UINT cnt)
 
     DbgPrint(TRACE_LEVEL_VERBOSE, ("---> %s\n", __FUNCTION__));
 
+    m_uCountMin = cnt;
+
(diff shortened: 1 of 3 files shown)


YanVugenfirer:

kostyanf14:
test this please

akihikodaki:
@vrozenfe I can always reproduce the issue by browsing the Web for five minutes on a VM on Asahi Linux, but it may not work for you; I suppose reproduction depends on whether the host can keep catching up with frames the guest generates.

PR #1332, opened 2025-03-19
viogpu: Allow combining DBG and WPP

akihikodaki:
Allow combining DBG and WPP by defining DBG and EVENT_TRACING at the same time.

### DIFF
--- a/viogpu/common/bitops.cpp
+++ b/viogpu/common/bitops.cpp
@@ -1,7 +1,7 @@
 #include "bitops.h"
 #include "viogpu.h"
 
-#if !DBG
+#if EVENT_TRACING
 #include "bitops.tmh"
 #endif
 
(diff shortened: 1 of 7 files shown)


YanVugenfirer:

YanVugenfirer (approved):

YanVugenfirer:
test this please

ybendito:
@akihikodaki I have a question (just a question): What is the motivation to define both DBG and WPP?

vrozenfe:
@akihikodaki @ybendito 
Yeah, I also have the same question :)
IIRC, the original intention was to enable WPP in release builds and keep DBG for the developers, if they need it.
Did we change this approach recently?

All the best,
Vadim.

akihikodaki:
@ybendito @vrozenfe My use case is to debug a detached Arm64 system.

Setting `DBG` gives logs via COM or `vDbgPrintEx` and enables debug assertions using `VIOGPU_ASSERT_CHK`. The logging does not work for a detached Arm64 system because:
- IO ports used for COM are not available on Arm64.
- `vDbgPrintEx` works only when attached.

Combining `DBG` and WPP allows having logging and debug assertions at the same time on a detached Arm64 system.

YanVugenfirer:

PR #1336, opened 2025-03-28
[RHEL-22354] potential fix for stuck packet

ybendito:
Jira link: https://issues.redhat.com/browse/RHEL-22354

In the investigation with CNBL history we can see that the decision based on read of m_DpcWaiting was not correct. The value reads as 1 but the m_AuditState.LastTxProcess show the timestamp in past (using perf counter timestamps), 30 micros ago. Making read of ref counter interlocked.

### DIFF
--- a/NetKVM/Common/ParaNdis-Util.h
+++ b/NetKVM/Common/ParaNdis-Util.h
@@ -276,7 +276,8 @@ class CNdisRefCounter
     }
     operator LONG()
     {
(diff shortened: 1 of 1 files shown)


ybendito:
Server 2025: RSSSendRecv - BSOD assertion in ndprot630: "pCurrentNetBufferList", known

YanVugenfirer (approved):

PR #1359, opened 2025-05-17
Tx no map - testing with DMAR with mapping disable for non-IP packets

ybendito:


YanVugenfirer (commented):
But what is the reason we need to copy? If it is because of the bad test, and there is a very low motivation for Win10 changes,  I prefer to have the committees that related to trading and code refactoring, but not to enable DMA remaking for Win10.

ybendito:
@YanVugenfirer The motivation of this PR is first of all to test it and get closer to the root cause. Will we enable DMAR for Win10 or not - this is for later decision. If we don't - this is also OK. Let it run, then we'll discuss it.

ybendito:
@YanVugenfirer Also, there is an open question "where is the break-even" between mapping and copy. The mapping/unmapping operation is heavy and we definitely loose performance on short packets.

PR #1363, opened 2025-05-25
[RHEL-84655]virtiolib-wdf: add check for active IOMMU

ybendito:
Jira link: https://issues.redhat.com/browse/RHEL-84655

Check whether physical addresses received from simple memory mapping and from DMA mapping are different. If they differ and there is no VIRTIO_F_ACCESS_PLATFORM suggested by the device - fail WDF DMA initialization. The driver will appear with yellow bang and with status 'The device is configured incorrectly...'


ybendito:
@vrozenfe @YanVugenfirer Unfortunately this does not work, due to some unclear reason I did not recognize this earlier, but it looks like that even without IOMMU the Windows can have 2 different working physical addresses referring the same memory. I'll think about alternate solution for detecting active IOMMU.

PR #1365, opened 2025-05-30
[Rhel-84655] virtio-wdf: test VIRTIO_F_ACCESS_PLATFORM and active IOMMU 

ybendito:


ybendito:
One failed test on server 2022 for each device is ok (on server 2022 clients desktop version is installed)
viomem 2022 "Embedded signature test" failed - HCK problem, the WTL is ok, passed on rerun
All the rest - passed

YanVugenfirer (approved):

PR #1370, opened 2025-06-03
Vioscsi: Registry tweak not to complete pending requests on bus reset

MartinDrab:
Doing so may lead to a nasty (and rare) memory corruption because when the requests are being completed by the driver, the host may have no idea about the reset yet. Thus, the host may attempt to access physical memory no longer owned by the requests.


YanVugenfirer:

MartinDrab:
Same thing will be added for `viostor` in https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1305 (just need to test it again and move it from its Draft state).

YanVugenfirer:
ok to test

YanVugenfirer:

benyamin-codez:
Martin, did this end up getting fixed with 0686f2c instead...?

PR #1378, opened 2025-06-19
RHEL-95788: Fix- support CLI argument for VirtioFsSvc service

elizashurov:
When creating a VirtioFsSvc service using sc create, command-line arguments appear to be ignored.

To work around this, when argc <= 1, we can manually parse the full command line using GetCommandLineW and CommandLineToArgvW, restoring proper argc/argv behavior.

### DIFF
--- a/viofs/svc/virtiofs.cpp
+++ b/viofs/svc/virtiofs.cpp
@@ -46,6 +46,7 @@
 #include <sys/stat.h>
 #include <wtsapi32.h>
 #include <cfgmgr32.h>
+#include <shellapi.h>
 
 #include <map>
 #include <string>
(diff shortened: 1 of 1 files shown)


YanVugenfirer:

kostyanf14:
ok to test

viofs/svc/virtiofs.cpp
```diff
@@ -2759,18 +2760,41 @@ static NTSTATUS SvcStart(FSP_SERVICE *Service, ULONG argc, PWSTR *argv)
...
+    bool shouldFreeFinalArgv = false;
 
-    if (argc > 1)
+    if (argc <= 1)
     {
-        Status = ParseArgs(argc,
-                           argv,
+        LPWSTR fullCommandLine = GetCommandLineW();
+        finalArgv = CommandLineToArgvW(fullCommandLine, &finalArgc);
+        if (finalArgv)
```
YanVugenfirer, line 2771: Did you test that it will override the registry settings?

elizashurov: Yes, the service now uses the arguments from the image path in the registry.

YanVugenfirer: Please also write in Jira comment for QE regarding how to test and update Wiki that command line will be used in case if registry settings already exists.

YanVugenfirer (approved):

PR #1403, opened 2025-08-05
RHEL-107465: [pciserial driver] Update INF files

elizashurov:
Set defaults to the same values as in drivers released by Red Hat for upstream.


YanVugenfirer:

kostyanf14 (approved):

PR #1415, opened 2025-08-26
Balloon: Small Fixes

MartinDrab:
This pull request addresses some issues that may appear during Balloon power up and power down sequences.
- **BalloonEvtDeviceD0Entry:** do not call `BalloonTerm` when `BalloonInit` fails since there is nothing to finalize in that case,
- **BalloonEvtDeviceD0Entry:** call `BalloonTerm` in case we fail to create the worker thread,
- **BalloonCreateWorkerThread:** check return value of the `ObReferenceObjectByxHandle` call. If it signals failure, wait for the worker thread to terminate,
- **BalloonEvtDeviceD0Exit:** Attempt to terminate the worker thread. This resolves the case when the `BalloonEvtDeviceD0Entry` callback succeeds but the Balloon device fails later but before enabling the interrupt (thus, `BalloonEvtDeviceD0ExitPreInterruptsDisabled` does not get called).

### DIFF
--- a/Balloon/sys/Device.c
+++ b/Balloon/sys/Device.c
@@ -433,6 +433,8 @@ BalloonEvtDeviceD0Exit(IN WDFDEVICE Device, IN WDF_POWER_DEVICE_STATE TargetStat
     }
 #endif // !BALLOON_INFLATE_IGNORE_LOWMEM
 
+    BalloonCloseWorkerThread(Device);
+
 #ifndef USE_BALLOON_SERVICE
     /*
      * interrupts were already disabled (between BalloonEvtDeviceD0ExitPreInterruptsDisabled and this call)


YanVugenfirer:

kostyanf14:
ok to test

kostyanf14:
@MartinDrab please fix the code style issue.

Balloon/sys/Device.c
```diff
@@ -301,28 +301,41 @@ BalloonCreateWorkerThread(IN WDFDEVICE Device)
...
+        goto Exit;
     }
 
+    status = ObReferenceObjectByHandle(hThread, THREAD_ALL_ACCESS, NULL, KernelMode, (PVOID *)&devCtx->Thread, NULL);
+    if (!NT_SUCCESS(status))
+    {
+        TraceEvents(TRACE_LEVEL_ERROR, DBG_PNP, "failed to reference thread: status 0x%08x\n", status);
+        devCtx->bShutDown = TRUE;
+        KeSetEvent(&devCtx->WakeUpThread, EVENT_INCREMENT, FALSE);
+        ZwWaitForSingleObject(hThread, FALSE, NULL);
```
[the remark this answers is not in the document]
MartinDrab: OK, I implemented your suggestion. Although, there is nothing that can be done apart from printing a warning message.

Balloon/sys/Device.c
```diff
@@ -367,20 +367,26 @@ BalloonEvtDeviceD0Entry(IN WDFDEVICE Device, IN WDF_POWER_DEVICE_STATE PreviousS
     if (!NT_SUCCESS(status))
     {
         TraceEvents(TRACE_LEVEL_ERROR, DBG_PNP, "BalloonInit failed with status 0x%08x\n", status);
-        BalloonTerm(Device);
-        return status;
+        goto Exit;
```
YanVugenfirer: For additional clarity maybe it worth renaming BalloonTerm to something like BalloonDeleteQueues?

If the name stays, I would goto Terminate and not Exit on line 370 for future generation clarity, while it functionally is not needed - to verify it is not need is a walk through 3-4 functions :)

MartinDrab: > For additional clarity maybe it worth renaming BalloonTerm to something like BalloonDeleteQueues?

I consider `BalloonInit` and `BalloonTerm` as opposites (the latter undoes what the former does). Thus, I would not rename and would not call `BalloonTerm` if `BalloonInit` failed. Yes, I know that calling `BalloonTerm` multiple times does nothing wrong.

Of course, I can do the changes if you wish.

YanVugenfirer: So let's call BalloonTerm if BalloonInit fails. Thank.

MartinDrab: OK, done.

YanVugenfirer: Thanks!

PR #1419, opened 2025-09-06
vioser: Do not requeue write requests completion of which is imminent

MartinDrab:
When a device stop is imminent (temporary one, not due to device removal), the Virtio Serial driver uses `WdfRequestStopAcknowledge` to requeue pending write requests to its write queue. This operation changes ownership of such requests from the driver to WDF. However, the driver may still attempt to complete the requests during its DPC processing. This is not allowed since a driver can only complete requests it owns. WDF Verifier complains about that.

This commit changes the behavior the following way:
- Write requests still present in the WriteBuffersList list are detached from the list (i.e. the Request pointer in the corresponding list entry is set to NULL) and requeued. The detach operation prevents their completion when owned by WDF.
- Write requests not present in the WriteBuffersList list are ignored (not ackowledged) since their completion is imminent. DPC is in progress and already removed them from the list and is about to complete them. Ignoring such requests in the EvtIoStop callback makes WDF to wait for their completion before continuing with device's power-down sequence.


YanVugenfirer:

vioserial/sys/Port.c
```diff
@@ -1183,7 +1183,44 @@ VOID VIOSerialPortWriteIoStop(IN WDFQUEUE Queue, IN WDFREQUEST Request, IN ULONG
...
     {
-        WdfRequestStopAcknowledge(Request, TRUE);
+        PSINGLE_LIST_ENTRY iter;
+
+        // If the request cannot be found in the WriteBuffersList list,
+        // it was removed from it during DPC processing and gets completed
+        // in a moment, thus, there is no need to acknowledge it here (the framework
+        // will wait for the completion).
+        //
+        // If the request is still in the list, �et's requeue it for next
```
kostyanf14: Strange symbol in comment

kostyanf14:
ok to test

vioserial/sys/Port.c
MartinDrab: Taken care of.

kostyanf14:
rerun tests

vioserial/sys/Port.c
```diff
@@ -1183,7 +1183,44 @@ VOID VIOSerialPortWriteIoStop(IN WDFQUEUE Queue, IN WDFREQUEST Request, IN ULONG
...
+                TraceEvents(TRACE_LEVEL_WARNING,
+                            DBG_WRITE,
+                            "Request 0x%p found in the WriteBuffersList list, requeuing\n",
+                            Request);
+                entry->Request = NULL;
+                break;
+            }
+        }
+
+        WdfSpinLockRelease(pport->OutVqLock);
```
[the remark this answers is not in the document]
MartinDrab: Similar function used.

YanVugenfirer:
@MartinDrab Can you please provide a scenario for the reproduction?

MartinDrab:
> @MartinDrab Can you please provide a scenario for the reproduction?

The idea is simple:
- repeatedly force the guest to write loads of data into the Virtio serial port. I was doing that by executing `type w11.iso` through `virsh` in the guest (`w11.iso` is an installation image for Windows 11, meaning about 6 GB of data to write),
- meanwhile, try to hotplug CPUs and memory to the guest. This causes the Virtio serial (and others) device to be stopped by the PnP Manager. If we are lucky, enough, the issue gets manifested.

For me, the guest got stuck (not reacting to `virsh shutdown`). When I connected WinDbg, the VM was on a breakpoint and `wdfkd.wdflogdump vioser` reported the problem.

I should have scripts for this, I will attach them shortly.

MartinDrab:
Here are the scripts. Take them as an inspiration. You will need to change the VM name and maybe the `x-exec` command (not sure if that is implemented in standard `virsh`).

The `serial-test.sh` script attempts to send file from the guest to the host. THe `test-hotplug.sh` script is responsible for hotplugging CPUs and memory modules. To force the hotplug sequence repeated 100 times, run `./test-hotplug.sh 1 100` (for example).

Both need to be run at once. 

[serial-test.sh](https://github.com/user-attachments/files/22369388/serial-test.sh)
[test-hotplug.sh](https://github.com/user-attachments/files/22369389/test-hotplug.sh)

YanVugenfirer:
Thanks!

PR #1427, opened 2025-09-18
[viostor]: reset virtqueues when driver receive a reset request

Atchess:
Last year, I implemented support for resetting virtio queues in viostor. This change is intended to address [issue #907](https://github.com/virtio-win/kvm-guest-drivers-windows/issues/907).

That said, I’m not deeply familiar with either the virtio specification or the Windows driver model, so I’m not entirely confident that this approach is free of side effects. Additionally, I feel that the current implementation may not be good from a code perspective. Here, I’ve only provided a simple implementation of this functionality, and if the community has a better solution, that would be even preferable.


YanVugenfirer:

YanVugenfirer (changes requested):
Thank you for the patch, I will leave @vrozenfe to comment on the storage related logic. Here are my general comments:
1. Even if virtio lib changes are needed - please keep them in separate commit. I personally don't think they are needed
2. Critical gemini code assist comment is correct
3. Resetting queue and then completing request is problematic as the memory related to descriptors should be released.

YanVugenfirer:
Reset logic was reimplemented.

PR #1446, opened 2025-11-03
netkvm: Add mergeable receive buffers support

zjmletang:
Hello everyone — I’ve prepared a preliminary PR that adds support for the netkvm “mergeable” feature. This can effectively reduce memory usage; I’d appreciate it if you could take a moment to review it when you have time. Below are the details/explanations for this PR.

# VirtIO Mergeable Receive Buffers Implementation Design

## Overview

This document describes the implementation of VirtIO mergeable receive buffers (VIRTIO_NET_F_MRG_RXBUF) support for the Windows NetKVM driver. The implementation optimizes memory usage and reduces buffer allocation overhead for high-throughput network scenarios.

### Problem Statement

The traditional receive path allocates large buffers (up to 64KB) to accommodate maximum-sized packets. This approach:
- Wastes memory when most packets are small (e.g., TCP ACKs, DNS queries)
- Limits the number of available receive buffers due to memory constraints
- Increases buffer pool exhaustion under heavy traffic

### Solution: Mergeable Receive Buffers

VirtIO mergeable buffers allow:
- Small fixed-size buffers (4KB pages) for all packets
- Multi-buffer packet assembly for large packets
- Better memory utilization and larger buffer pools
- Reduced allocation overhead

### 1. **Conditional Activation**
Only enable mergeable buffers when BOTH features are present:
- `VIRTIO_NET_F_MRG_RXBUF` (mergeable buffer support)
- `VIRTIO_F_ANY_LAYOUT` (VirtIO 1.0+ combined header+data layout)

**Rationale**: This is a pragmatic engineering decision. While `VIRTIO_NET_F_MRG_RXBUF` and `VIRTIO_F_ANY_LAYOUT` are theoretically independent features per VirtIO specification, we require both for mergeable buffer support to:
- **Simplify implementation**: Combined layout allows single scatter-gather entry per buffer
- **Reduce code paths**: Eliminates need for separate header/data descriptor logic
- **Minimize testing matrix**: Avoids testing mergeable mode with legacy split-descriptor layout
- **Target modern devices**: VirtIO 1.0+ devices (released 2016+) support both features

Legacy VirtIO 0.95 devices lacking `ANY_LAYOUT` automatically fall back to traditional non-mergeable path.

### 2. **Zero-Allocation Hot Path**
All data structures for packet assembly are pre-allocated:
- **Inline arrays in merge context**: Buffer references and physical page arrays use fixed-size inline storage (`BufferSequence[17]`, `PhysicalPages[18]`) embedded in the context structure, eliminating per-packet heap allocation overhead.

- **Stack-based storage for buffer references**: All temporary tracking uses stack variables or pre-allocated context members, avoiding dynamic memory management in the receive hot path.

- **Bounds enforced at compile time**: Array sizes are compile-time constants (`#define VIRTIO_NET_MAX_MRG_BUFS 17`), not runtime variables. This allows the compiler to enforce bounds checking and eliminates the need for runtime validation checks, improving performance and safety.

### 3. **Backward Compatibility**
Maintain compatibility with existing code paths:
- **Physical page aliasing preserves `ParaNdis_BindRxBufferToPacket` logic**: The existing MDL binding function always starts from `PhysicalPages[1]` (legacy design for traditional mode). By creating an alias where both `PhysicalPages[0]` and `[1]` point to the same physical memory, this function works correctly for mergeable buffers without modification, reducing regression risk.

- **Separate creation paths for mergeable vs. traditional buffers**: Two independent buffer allocation functions (`CreateMergeableRxDescriptor()` for small 4KB buffers, `CreateRxDescriptorOnInit()` for large multi-page buffers) are selected based on feature flags. This separation keeps each mode's logic clean and isolated, avoiding complex conditional branches in shared code.

- **No changes to core packet processing**: After assembly, both modes produce identical MDL chain structures. Upper-layer processing (checksum validation, RSS classification, hardware offload) operates on standard MDL chains regardless of buffer origin. This interface consistency ensures existing packet handling logic works unchanged for both modes.

#### 1. RxNetDescriptor Extensions
```c
struct _tagRxNetDescriptor {
    // Modified fields:
    USHORT NumPages;                           // Logical page count (exposed to offload engines)
    USHORT NumOwnedPages;                      // Physical ownership (vs. logical pages)
    
    // New fields for mergeable support:
    tCompletePhysicalAddress *OriginalPhysicalPages;  // Saved for restoration
    USHORT MergedBufferCount;                  // Additional buffers (excluding this one)
    pRxNetDescriptor MergedBuffers[16];  // Inline storage (no allocation)
};
```

**Design Notes**:
- `NumPages`: **Semantic change** - now represents logical page count for complete packet after assembly
  - Traditional mode: Physical page count = logical page count (same value)
  - Mergeable mode: Logical count > physical count (includes pages from additional buffers)
  - Must remain accurate for checksum/offload engines to process full packet correctly
- `NumOwnedPages`: Prevents double-free during cleanup (only free owned pages)
  - Always equals the actual physical pages this descriptor owns
  - Mergeable mode: Always 2 for single-buffer descriptor
  - After assembly: Still 2 (doesn't change, only first buffer owns its pages)
- `OriginalPhysicalPages`: Enables pointer restoration after merge assembly
- `MergedBuffers` array: Avoids heap allocation, sized for worst case (17 buffers max)

#### 2. Merge Context Structure
```c
struct _MergeBufferContext {
    pRxNetDescriptor BufferSequence[17];           // Buffer collection
    UINT32 BufferActualLengths[17];                // Received lengths
    UINT16 ExpectedBuffers;                        // From virtio header
    UINT16 CollectedBuffers;                       // Current count
    UINT32 TotalPacketLength;                      // Accumulated size
    tCompletePhysicalAddress PhysicalPages[18];  // Pre-allocated array
};
```

**Design Notes**:
- `VIRTIO_NET_MAX_MRG_BUFS = 17`: Calculated from max packet size (65562 bytes / 4096 bytes per buffer)
- `MAX_MERGED_PHYSICAL_PAGES = 18`: First buffer (2 logical pages) + 16 additional buffers
- All arrays are compile-time sized to avoid runtime allocation

### Physical Page Aliasing Design

**Challenge**: `ParaNdis_BindRxBufferToPacket` always starts from `PARANDIS_FIRST_RX_DATA_PAGE` (index 1).

**Solution**: Create an alias where both `PhysicalPages[0]` and `PhysicalPages[1]` point to the same physical memory.

```
Traditional mode:              Mergeable mode (aliasing):
PhysicalPages[0] → Header      PhysicalPages[0] ───┐
PhysicalPages[1] → Data page 1                     ├→ Same 4KB page
PhysicalPages[2] → Data page 2 PhysicalPages[1] ───┘
```

**Safety**: `IsRegionInside()` detects the alias during cleanup and skips freeing `PhysicalPages[1]`.

**Trade-off**: +8 bytes per descriptor (1000 descriptors = 8KB overhead), acceptable for compatibility.

#### CreateMergeableRxDescriptor()
Allocates simplified 4KB buffers for mergeable mode:

1. Allocate single 4KB physical page
2. Create 2-entry `PhysicalPages` array (aliasing design)
3. Set both entries to point to the same physical page
4. Create single scatter-gather entry (ANY_LAYOUT mode)
5. Bind MDL starting from index 1 (compatibility)

**Key Parameters**:
- `NumPages = 2` (logical)
- `NumOwnedPages = 2` (same for single buffer)
- `BufferSGLength = 1` (combined header+data)
- `DataStartOffset = nVirtioHeaderSize`

#### ProcessMergedBuffers()
Main entry point for mergeable packet handling:

1. Read `num_buffers` from virtio header
2. Validate range (1-17 buffers)
3. Handle single-buffer case (fast path, no assembly)
4. Initialize merge context
5. Collect remaining buffers via `CollectRemainingMergeBuffers()`
6. Assemble packet via `AssembleMergedPacket()`

**Error Handling**:
- Invalid `num_buffers`: Drop packet, reuse first buffer
- Collection failure: Drop packet, reuse all collected buffers
- Assembly failure: Drop packet, reuse all collected buffers

#### CollectRemainingMergeBuffers()
Retrieves buffers 2..N from virtqueue:

**VirtIO Protocol Guarantee**: All buffers for a merged packet are atomically available.

**Implementation**:
- Collect remaining buffers based on `num_buffers` from virtio header
- Store actual received lengths for each buffer

#### AssembleMergedPacket()
Combines multiple buffers into single packet:

1. **Save buffer references**: Store additional buffers in `MergedBuffers` array
2. **Switch to inline PhysicalPages**: Use pre-allocated `PhysicalPages` array from merge context
3. **Copy page references**: First buffer (2 pages) + additional buffers (1 page each)
4. **Create MDLs**: New MDLs for additional buffers covering full payload (no header offset)
5. **Update counts**: `NumPages` (logical), `MergedBufferCount` (additional buffers)

**Page Calculation**:
```
totalPages = 2 (first buffer) + (CollectedBuffers - 1) (additional)
           = 1 + CollectedBuffers
```

**MDL Creation**: Additional buffers use `PhysicalPages[PARANDIS_FIRST_RX_DATA_PAGE]` (index 1, aliased to same page as [0] in mergeable mode, for consistency with `ParaNdis_BindRxBufferToPacket`).

#### ReuseReceiveBufferNoLock()
Enhanced to handle merged packets:

1. Check `MergedBufferCount > 0`
2. Recursively reuse all additional buffers
3. Call `DisassembleMergedPacket()` to restore state
4. Standard reuse logic

#### DisassembleMergedPacket()
Inverse operation of `AssembleMergedPacket()`:

1. Free extended MDL chain (keep first buffer's original MDL)
2. Restore `PhysicalPages` pointer from inline array to original
3. Reset `NumPages = 2`, `NumOwnedPages = 2`
4. Clear `MergedBufferCount = 0`

**Result**: Buffer returns to pristine single-buffer state for reuse.

### Per-Descriptor Structure Overhead
Mergeable mode adds new fields to `RxNetDescriptor` structure:
```
Additional fields compared to traditional mode:
   +8 bytes (OriginalPhysicalPages pointer)
   +2 bytes (NumOwnedPages)
   +2 bytes (MergedBufferCount)
   +128 bytes (MergedBuffers array, 16 * 8 bytes)
   = +140 bytes per descriptor
```

### Per-Queue Context Overhead
Mergeable mode adds `_MergeBufferContext` to each RX queue:
```
BufferSequence[17]:         17 * 8  = 136 bytes
BufferActualLengths[17]:    17 * 4  = 68 bytes
ExpectedBuffers:            2 bytes
CollectedBuffers:           2 bytes
TotalPacketLength:          4 bytes
PhysicalPages[18]:          18 * 24 = 432 bytes
                            Total:    644 bytes per queue
```

### Per-Buffer Physical Memory Allocation
Actual shared memory (DMA-capable) allocated per buffer:
```
Traditional mode:  Up to 18 physical pages per buffer (max ~72KB)
                   Layout: 1 header page + up to 17 data pages
                   Note: Must pre-allocate for worst-case packet size
                   
Mergeable mode:    1 physical page per buffer (4KB)
                   Layout: Single 4KB page (header + data combined)
                   Note: 2 logical pages via aliasing, but same physical page
```

### Total Impact (Example: 4096 buffers/queue, 1 queue)
```
Additional metadata overhead:
  Descriptor structure: 4096 * 140 bytes = 574 KB
  Queue context:        1 * 644 bytes   = 0.6 KB
  Total overhead:                         574.6 KB

Shared memory savings:
  Traditional: 4096 * 18 pages * 4KB = ~288 MB (worst-case pre-allocation)
  Mergeable:   4096 * 1 page * 4KB   = ~16 MB
  Net savings:                          ~272 MB (94% reduction)
```

**Conclusion**: Minimal metadata overhead (~575 KB) enables significant shared memory savings (~272 MB).

### Protocol Violations
1. **Invalid `num_buffers`** (0 or >17):
   - Log error, drop packet
   - Reuse first buffer immediately

2. **Missing buffers** (GetBuf returns NULL):
   - Log protocol violation error
   - Reuse all collected buffers

3. **Buffer overflow** (>16 additional buffers):
   - Log critical error, drop packet
   - Reuse all collected buffers
   - Should never happen (pre-validated)

### Resource Exhaustion
1. **MDL allocation failure**:
   - Log error, abort packet assembly, return NULL
   - Caller reuses all collected buffers
   - Partial MDL chain is cleaned up automatically
   - Entire packet is dropped (ensures data integrity)
   - Rare in practice

### Key Functions Added
- `CreateMergeableRxDescriptor()`: Simplified buffer creation
- `ProcessMergedBuffers()`: Main assembly coordinator
- `CollectRemainingMergeBuffers()`: Buffer collection
- `AssembleMergedPacket()`: Multi-buffer packet assembly
- `DisassembleMergedPacket()`: State restoration for reuse
- `ReuseCollectedBuffers()`: Batch buffer return
- `ProcessReceivedPacket()`: Encapsulates packet analysis, filtering, and RSS processing

### Key Functions Modified
- `ReuseReceiveBufferNoLock()`: Added merged packet handling
- `CreateRxDescriptorOnInit()`: Added path routing
- `ProcessRxRing()`: Integrated mergeable path

### DIFF
--- a/NetKVM/Common/ParaNdis_Common.cpp
+++ b/NetKVM/Common/ParaNdis_Common.cpp
@@ -903,7 +903,8 @@ NDIS_STATUS ParaNdis_InitializeContext(PARANDIS_ADAPTER *pContext, PNDIS_RESOURC
         pContext->bUseMergedBuffers = AckFeature(pContext, VIRTIO_NET_F_MRG_RXBUF);
         if (!pContext->bMergeableBuffersConfigured && pContext->bUseMergedBuffers)
         {
-            DPrintf(0, "Mergeable buffers feature ACKed, but disabled by configuration (using traditional large buffers)");
+            DPrintf(0,
+                    "Mergeable buffers feature ACKed, but disabled by configuration (using traditional large buffers)");
         }
         pContext->nVirtioHeaderSize = (pContext->bUseMergedBuffers) ? sizeof(virtio_net_hdr_mrg_rxbuf)
                                                                     : sizeof(virtio_net_hdr);
--- a/NetKVM/Common/ParaNdis_RX.cpp
+++ b/NetKVM/Common/ParaNdis_RX.cpp
@@ -486,9 +486,46 @@ void CParaNdisRX::FreeRxDescriptorsFromList()
     }
 }
 
+// Disassemble a merged packet back to its original single-buffer state.
+// Unlink the MDL chain and restore original page mapping and counters.
+void CParaNdisRX::DisassembleMergedPacket(pRxNetDescriptor pBuffer)
+{
+    PMDL pMDL = pBuffer->Holder;
+    PMDL pNextMDL = NULL;
+
+    while (pMDL)
+    {
+        pNextMDL = NDIS_MDL_LINKAGE(pMDL);
+        NDIS_MDL_LINKAGE(pMDL) = NULL;
+        pMDL = pNextMDL;
+    }
+
+    pBuffer->PhysicalPages = pBuffer->OriginalPhysicalPages;
+    pBuffer->NumPages = 1;
+    pBuffer->NumOwnedPages = 1;
+    pBuffer->MergedBufferCount = 0;
+}
+
 void CParaNdisRX::ReuseReceiveBufferNoLock(pRxNetDescriptor pBuffersDescriptor)
 {
     DEBUG_ENTRY(4);
+
+    // Handle merged packets: recursively reuse all constituent buffers
+    if (pBuffersDescriptor->MergedBufferCount > 0)
+    {
+        DPrintf(4, "Reusing merged packet with %u additional buffers", pBuffersDescriptor->MergedBufferCount);
+
+        // Reuse additional buffers stored in inline array
+        // Note: Additional buffers retain their original MDLs (we used pre-allocated
+        //       FullPageMDLs in AssembleMergedPacket, so their original MDLs are intact)
+        for (USHORT i = 0; i < pBuffersDescriptor->MergedBufferCount; i++)
+        {
+            ReuseReceiveBufferNoLock(pBuffersDescriptor->MergedBuffers[i]);
+        }
+
+        // Disassemble the first buffer back to its original single-buffer state
+        DisassembleMergedPacket(pBuffersDescriptor);
+    }
 
     if (!m_Reinsert)
     {
@@ -726,14 +763,91 @@ static void LogRedirectedPacket(pRxNetDescriptor pBufferDescriptor)
 }
 #endif
 
+// Handles packet analysis, filtering, RSS processing, and queue assignment
+void CParaNdisRX::ProcessReceivedPacket(pRxNetDescriptor pBufferDescriptor, CCHAR nCurrCpuReceiveQueue)
+{
+    // Get data pointer (skip virtio header)
+    PVOID data = pBufferDescriptor->PhysicalPages[pBufferDescriptor->FirstRxDataPage].Virtual;
+    data = RtlOffsetToPointer(data, pBufferDescriptor->DataStartOffset);
+
+    // basic MAC-based analysis + L3 header info
+    BOOLEAN packetAnalysisRC = ParaNdis_AnalyzeReceivedPacket(data,
+                                                              pBufferDescriptor->PacketInfo.dataLength,
+                                                              &pBufferDescriptor->PacketInfo);
+
+    if (!packetAnalysisRC)
+    {
+        ReuseReceiveBufferNoLock(pBufferDescriptor);
+        m_Context->Statistics.ifInErrors++;
+        m_Context->Statistics.ifInDiscards++;
+        return;
+    }
+
+    // filtering based on prev stage analysis
+    if (!ShallPassPacket(m_Context, &pBufferDescriptor->PacketInfo))
+    {
+        pBufferDescriptor->Queue->ReuseReceiveBufferNoLock(pBufferDescriptor);
+        m_Context->Statistics.ifInDiscards++;
+        m_Context->extraStatistics.framesFilteredOut++;
+        return;
+    }
+
+#ifdef PARANDIS_SUPPORT_RSS
+    if (m_Context->RSSParameters.RSSMode != PARANDIS_RSS_MODE::PARANDIS_RSS_DISABLED)
+    {
+        ParaNdis6_RSSAnalyzeReceivedPacket(&m_Context->RSSParameters, data, &pBufferDescriptor->PacketInfo);
+    }
+    CCHAR nTargetReceiveQueueNum;
+    GROUP_AFFINITY TargetAffinity;
+    PROCESSOR_NUMBER TargetProcessor;
+
+    nTargetReceiveQueueNum = ParaNdis_GetScalingDataForPacket(m_Context,
+                                                              &pBufferDescriptor->PacketInfo,
+                                                              &TargetProcessor);
+
+    if (nTargetReceiveQueueNum == PARANDIS_RECEIVE_UNCLASSIFIED_PACKET)
+    {
+        ParaNdis_ReceiveQueueAddBuffer(&m_UnclassifiedPacketsQueue, pBufferDescriptor);
+        m_Context->extraStatistics.framesRSSUnclassified++;
+    }
+    else
+    {
+        ParaNdis_ReceiveQueueAddBuffer(&m_Context->ReceiveQueues[nTargetReceiveQueueNum], pBufferDescriptor);
+
+        if (nTargetReceiveQueueNum != nCurrCpuReceiveQueue)
+        {
+            if (m_Context->bPollModeEnabled)
+            {
+                // ensure the NDIS just schedules the other poll and does not do anything
+                // otherwise if both polls are configured to the same CPU
+                // this may cause a deadlock in return nbl path
+                KIRQL prev = KeRaiseIrqlToSynchLevel();
+                ParaNdisPollNotify(m_Context, nTargetReceiveQueueNum, "RSS");
+                KeLowerIrql(prev);
+            }
+            else
+            {
+                ParaNdis_ProcessorNumberToGroupAffinity(&TargetAffinity, &TargetProcessor);
+                ParaNdis_QueueRSSDpc(m_Context, m_messageIndex, &TargetAffinity);
+            }
+            m_Context->extraStatistics.framesRSSMisses++;
+            LogRedirectedPacket(pBufferDescriptor);
+        }
+        else
+        {
+            m_Context->extraStatistics.framesRSSHits++;
+        }
+    }
+#else
+    UNREFERENCED_PARAMETER(nCurrCpuReceiveQueue);
+    ParaNdis_ReceiveQueueAddBuffer(&m_UnclassifiedPacketsQueue, pBufferDescriptor);
+#endif
+}
+
 VOID CParaNdisRX::ProcessRxRing(CCHAR nCurrCpuReceiveQueue)
 {
     pRxNetDescriptor pBufferDescriptor;
     unsigned int nFullLength;
-
-#ifndef PARANDIS_SUPPORT_RSS
-    UNREFERENCED_PARAMETER(nCurrCpuReceiveQueue);
-#endif
 
     TDPCSpinLocker autoLock(m_Lock);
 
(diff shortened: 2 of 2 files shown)


YanVugenfirer:

kostyanf14:
ok to test

ybendito:
@zjmletang Thank you for the PR, I'll review it as soon as I have a time, but not immediately.

ybendito:
@zjmletang What I do not like in this PR:
- Processing of merged packets in the worst case (64K guest-to-guest or host-to-guest) allocates new MDL ~15 times, on each packet return ~15 MDL free. I think it can't be negligible in terms of CPU consumption. Among other options, there is a possibility to preallocate the MDL describing entire page, which we can use for merged buffers and which we do not need to adjust (last MDL still can describe entire page and the usable area is limited by the length of NET_BUFFER). When the packet is returned we just need to set each MDL->Next to NULL.
- The "use merged buffers" is a default without any fallback possibility. This is especially problematic when the CPU consumption per-packet is definitely higher.
- The process of descriptor preparation at init time is not reflected in the RX layout and do not use RX layout at all. Rx layout is not  meaningless, it helps in Rx memory preparation. For example, in case if the  MTU=1500 and the RSC is disabled in hardware (guest_tso4=off, guest_tso6=off) the layout dictates allocation size of 2048. In such case the 'non-mergeable' code allocates 2048 bytes per package and the 'mergeable' code allocates for each descriptor full 4K page even if it will never be used. 

One thing off-topic: I've had an idea to have one-per-queue endless MDL chain over all the buffers (first-to-last-to-first) and always use it for the indicated NET_BUFFER, each time using as first MDL the actual MDL of the first retrieved packet. Unfortunately, this will not work (( as the packets might be returned out-of-order of the indication and the order in which they will be added to the queue is not constant.

zjmletang:
@ybendito Thank you very much for taking the time to review this PR and for the detailed feedback. Your suggestions are very valuable.
I will address all the points you raised:
1. Pre-allocate MDLs to avoid hot-path allocation/free overhead
2. Add a configuration option to allow fallback to non-mergeable mode
3. Use RxLayout to determine buffer allocation size

I'll prepare a v2 patch series with these changes and submit it soon.

kostyanf14:
@zjmletang in latest build we have MUSTFIX defect `<AssessmentScore ScoreName="netkvm.x64.CodeAnalysis.Defect.6011" ScoreValue="1" ScoreUnit="CA_MUSTFIX_FAILED" />`

zjmletang:
@kostyanf14 , Please help check the causes of these two build failures. I looked through the build logs, and it doesn’t look like the problem is in the code itself.
image

zjmletang:
@ybendito Thank you again for the previous review and suggestions.

I’ve updated this PR with v2 of the changes:
- Pre-allocate MDLs to avoid hot-path allocation/free overhead
- Add a configuration option to allow fallback to non-mergeable mode
- Use RxLayout to determine the buffer allocation size
- Introduce a new `FirstRxDataPage` field to replace `PARANDIS_FIRST_RX_DATA_PAGE`

For the third point, I’m not completely sure whether the approach will be acceptable to the community, so I’d especially appreciate your feedback on it. In v2 I chose a fixed scheme for the mergeable buffer allocation mode:
- For mergeable buffers I use a fixed allocation size of one PAGE_SIZE.
- The virtio header and payload are combined in the same buffer.
- Based on `RxLayout.TotalAllocationsPerBuffer` and other parameters, the driver decides whether to use the mergeable allocation mode; if the conditions are not met, it falls back to the traditional (non-mergeable) allocation mode.

I also evaluated an alternative implementation where the mergeable allocation size would be derived directly from fields in `RxLayout`. However, there is no particularly suitable field for this purpose (unless we introduce a new one), and the current fixed-size approach looks simpler and more straightforward, so I went with the current design.

Please take another look when you have time. Thanks!

NetKVM/netkvm-base.txt
```diff
@@ -224,6 +224,12 @@ HKR, Ndi\params\MinRxBufferPercent,         min,        0,          "0"
 HKR, Ndi\params\MinRxBufferPercent,         max,        0,          "100"
 HKR, Ndi\params\MinRxBufferPercent,         step,       0,          "1"
 
+HKR, Ndi\Params\MergeableBuffers,           ParamDesc,  0,          %MergeableBuffers%
+HKR, Ndi\Params\MergeableBuffers,           Default,    0,          "1"
```
ybendito: Only "0" we can use at this stage, we did not have any throughput comparison tests yet

zjmletang: Got it, I’ll change the default to "0".
I’ll wait for any other review comments and then update the patch all together.

zjmletang:
> @zjmletang in latest build we have MUSTFIX defect `<AssessmentScore ScoreName="netkvm.x64.CodeAnalysis.Defect.6011" ScoreValue="1" ScoreUnit="CA_MUSTFIX_FAILED" />`

hi @kostyanf14 
 I'm using Visual Studio 2022 with SDK 10.0.26100.0 on my end, but I don't seem to see this warning. Could you let me know what build environment or tools you are using? I'd like to try to reproduce it.
image
image

kostyanf14:
> > @zjmletang in latest build we have MUSTFIX defect `<AssessmentScore ScoreName="netkvm.x64.CodeAnalysis.Defect.6011" ScoreValue="1" ScoreUnit="CA_MUSTFIX_FAILED" />`
> 
> hi @kostyanf14 I'm using Visual Studio 2022 with SDK 10.0.26100.0 on my end, but I don't seem to see this warning. Could you let me know what build environment or tools you are using? I'd like to try to reproduce it. image image

Builder uses EDWK for Win11 24H2

NetKVM/netkvm-base.txt
zjmletang: Hi @ybendito,

I've  changed the default value to "0" as we discussed. I'm also currently addressing the CI defect (6011) to ensure the PR is in good shape.

Regarding the performance evaluation, I'm preparing to conduct throughput tests covering both TCP and UDP across various packet sizes, while also monitoring CPU utilization. 

Please let me know if there are any other specific scenarios or workloads you would like to see in the test plan. I want to ensure the data is as comprehensive and informative as possible for the project.

Thanks!

NetKVM/Common/ParaNdis_Common.cpp
```diff
@@ -890,7 +897,16 @@ NDIS_STATUS ParaNdis_InitializeContext(PARANDIS_ADAPTER *pContext, PNDIS_RESOURC
         InitializeMAC(pContext, CurrentMAC);
         InitializeMaxMTUConfig(pContext);
 
-        pContext->bUseMergedBuffers = AckFeature(pContext, VIRTIO_NET_F_MRG_RXBUF);
+        // Honor bMergeableBuffersConfigured to allow fallback to non-mergeable mode
+        if (pContext->bMergeableBuffersConfigured)
+        {
+            pContext->bUseMergedBuffers = AckFeature(pContext, VIRTIO_NET_F_MRG_RXBUF);
```
ybendito: This point I want to discuss. I'll send my notes for this point separately in PR comment.

NetKVM/Common/ndis56common.h
```diff
@@ -444,20 +444,46 @@ struct _tagRxNetDescriptor
...
     USHORT HeaderPage;
-    // data is always pages[1] but might
-    // be after the virtio header
+    // Index of the first page containing payload data
+    // Non-mergeable: 1 (separate header page)
+    // Mergeable: 0 (header and data share same page)
+    USHORT FirstRxDataPage;
+    // data might be after the virtio header
     USHORT DataStartOffset;
 #define PARANDIS_FIRST_RX_DATA_PAGE (1)
```
ybendito: Probably this should be removed to avoid misunderstandings

ybendito:
@zjmletang in the mainstream we always ack the feature if it is suggested by the device. So the device is free to split the packet over several buffers if due to some [misunderstanding] reason the incoming packet is larger than the buffer (case of device RSC is off, i.e. host features VIRTIO_NET_F_GUEST_TSO4 and VIRTIO_NET_F_GUEST_TSO6 are off and there is inconsistency with guest MTU and host MTU of the taps).
In this series the behavior of the driver was changed to avoid ack if the registry setting is 'no merged buffers'. So it becomes possible that the device is not able to copy the complete packet to the guest buffer. Now it depends on the hypervisor what happens. I remember that old QEMU even did the assert in such a case.
The driver is used with many various hypervisors (we are not aware of some of them) so I see that this change may cause the regression in some cases (even if the change seems ok for me).
We have an option to keep backward compatible behavior (or even add a simple code to drop merged packets in the driver) OR to take a risk of such a regression. 
What you think about it?

zjmletang:
> @zjmletang in the mainstream we always ack the feature if it is suggested by the device. So the device is free to split the packet over several buffers if due to some [misunderstanding] reason the incoming packet is larger than the buffer (case of device RSC is off, i.e. host features VIRTIO_NET_F_GUEST_TSO4 and VIRTIO_NET_F_GUEST_TSO6 are off and there is inconsistency with guest MTU and host MTU of the taps). In this series the behavior of the driver was changed to avoid ack if the registry setting is 'no merged buffers'. So it becomes possible that the device is not able to copy the complete packet to the guest buffer. Now it depends on the hypervisor what happens. I remember that old QEMU even did the assert in such a case. The driver is used with many various hypervisors (we are not aware of some of them) so I see that this change may cause the regression in some cases (even if the change seems ok for me). We have an option to keep backward compatible behavior (or even add a simple code to drop merged packets in the driver) OR to take a risk of such a regression. What you think about it?

@ybendito 
Thank you for the detailed feedback. After reconsidering your points, I completely agree that keeping the backward compatible behavior is the much safer and more responsible choice for a driver like NetKVM, which is so widely used in production environments.
I'd rather avoid any risk of regressions with legacy or sensitive Hypervisors, even if it means a small amount of extra logic in the traditional path.
I will update the PR accordingly:
1. Always ACK VIRTIO_NET_F_MRG_RXBUF if offered by the device
2. Handle the 12-byte header in all modes:
In the traditional path (~64KB buffers), I'll add a simple check for num_buffers. If it's ever > 1, the driver will safely drop the packet.

NetKVM/Common/ndis56common.h
zjmletang: ok

zjmletang:
@ybendito I have addressed the two comments you mentioned and pushed the updates. Please take another look when you have a chance to see if any further changes are needed. Thanks!

zjmletang:
> > > @zjmletang in latest build we have MUSTFIX defect `<AssessmentScore ScoreName="netkvm.x64.CodeAnalysis.Defect.6011" ScoreValue="1" ScoreUnit="CA_MUSTFIX_FAILED" />`
> > 
> > 
> > hi @kostyanf14 I'm using Visual Studio 2022 with SDK 10.0.26100.0 on my end, but I don't seem to see this warning. Could you let me know what build environment or tools you are using? I'd like to try to reproduce it. image image
> 
> Builder uses EDWK for Win11 24H2

@kostyanf14 The issue is resolved in the latest update. It turns out I had already inadvertently fixed it in the v2 version  (^_^)

zjmletang:
HLK Server 2022 Failure: NICStrictPropertyValidation. The current .hlkx results do not clearly indicate the root cause of the issue. I have tested it locally on Windows Server 2022 and it passed. I suspect that a simple rerun might resolve the failure.
HLK Server 2025 Failure: we are seeing an excessive number of HLK failures. This might be a known issue on Microsoft's end; however, it also appears that the HLK filters are not taking effect as expected.

zjmletang:
hi @ybendito. Here are some initial throughput results based on the latest commit for our discussion. Feel free to point out if any additional data is required，

Performance Test Report: NIC Mergeable Attribute Evaluation
1. Test Objective
To evaluate the impact of the Mergeable attribute on netkvm performance, focusing on throughput, PPS, CPU overhead, and system memory consumption (Non-Paged Pool) using poolmon.
2. Test Methodology
● Infrastructure: 1 Sender VM and 1 Receiver VM on a single physical host.
● Automation: Python automation script using ntttcp.
● Memory Monitoring: poolmon utility used to measure Non-Paged Pool (NPP) usage.
● Variable: Comparison between mergeable: ON and mergeable: OFF.

3. Test Environment 1 (Low-Spec)
Configuration: 2 vCPU, 2GB RAM, 1 Queue, 256 Ring Size, 1500 MTU.
3.1 Traffic Performance: Mergeable [ON]
image
3.2 Traffic Performance: Not Mergeable [OFF]
image
3.3 Memory Footprint: Non-Paged Pool (Env 1)
image
4. Test Environment 2 (High-Spec)
Configuration: 8 vCPU, 16GB RAM, 8 Queues, 1024 Ring Size, 9000 MTU.
4.1 Traffic Performance: Mergeable [ON]
image
4.2 Traffic Performance: Not Mergeable [OFF]
image
4.3 Memory Footprint: Non-Paged Pool (Env 2)
image

zjmletang:
Additionally, the scripts I used for throughput testing are provided below. To use them, first run ntttcp_send.py on the sender VM, followed by ntttcp_recv.py on the receiver VM. Note that you can adjust the IP address and the test duration for each group within ntttcp_recv.py
[ntttcp_recv.py](https://github.com/user-attachments/files/24741687/ntttcp_recv.py)
[ntttcp_send.py](https://github.com/user-attachments/files/24741688/ntttcp_send.py)

ybendito:
@zjmletang Thank you for the detailed performance data. I'll try/review the final commits soon

zjmletang:
HLK Win10 22H2 x64 Failure Analysis:
There are 2 failures in total, and preliminary analysis suggests that neither of them appear to be rooted in the driver itself:

NDISTest 6.5 - [2 Machine] - E2EPerf: This test passed upon a local re-run, suggesting the initial failure might be transient or environment-related.
NDISTest 6.0 - [2 Machine] - 2c_Mini6RSSSendRecv (Multi-Group Win8+): Based on local testing, this failure seems to be eligible for an HLK filter.
image

ybendito (approved):

PR #1449, opened 2025-11-06
netkvm: Immediately kick device when buffer shortage detected

zjmletang:
When the available buffer count falls below the configured threshold (e.g., MinRxBufferPercent = 25%), the system detects buffer shortage. Previously, the device notification relied solely on batch notification threshold (approximately 25% of buffers), which could cause delayed notification when buffers are already in shortage, leading to packet loss.

This change adds an immediate kick when buffer shortage is detected, ensuring the device is notified promptly rather than waiting for the batch threshold to be reached. The fix checks IsRxBuffersShortage() in addition to the batch counter limit, so that when the system detects buffer shortage (based on user configurable MinRxBufferPercent), it immediately notifies the device.

### DIFF
--- a/NetKVM/Common/ParaNdis_RX.cpp
+++ b/NetKVM/Common/ParaNdis_RX.cpp
@@ -417,7 +417,8 @@ void CParaNdisRX::ReuseReceiveBufferNoLock(pRxNetDescriptor pBuffersDescriptor)
         }
 
         /* TODO - nReusedRXBuffers per queue or per context ?*/
-        if (++m_nReusedRxBuffersCounter >= m_nReusedRxBuffersLimit)
+        m_nReusedRxBuffersCounter++;
+        if (IsRxBuffersShortage() || m_nReusedRxBuffersCounter >= m_nReusedRxBuffersLimit)
         {
             m_nReusedRxBuffersCounter = 0;
             m_VirtQueue.Kick();


YanVugenfirer:

zjmletang:
I’d like to discuss whether this PR is reasonable — the change may to some extent alter the original intent of MinRxBufferPercent (personally, I think this kind of extension is reasonable). 
This PR was inspired by https://github.com/virtio-win/kvm-guest-drivers-windows/issues/1012. I tried the method described in that issue and found that when MinRxBufferPercent was set below 25%, ping packets were still easily lost; using values close to 30% (i.e. slightly above 25%) made packet loss less likely. I suspect this was because m_nReusedRxBuffersLimit happened to be set at roughly 25%. After applying the changes in this PR, setting MinRxBufferPercent to 25% no longer causes ping packet loss.

YanVugenfirer:
@zjmletang Thanks a lot for the PR! Do you have some performance numbers comparing with current code and PR code?

kostyanf14:
ok to test

zjmletang:
> @zjmletang Thanks a lot for the PR! Do you have some performance numbers comparing with current code and PR code?

hi @YanVugenfirer 

# Network Performance Test Report

## Test Background
No particularly suitable testing method was found since this issue can only be reproduced in specific scenarios. Only 3 simple comparative test groups were conducted.

## Test Method
Used the reproduction script mentioned in [virtio-win issue #1012](https://github.com/virtio-win/kvm-guest-drivers-windows/issues/1012) to build the test environment, then used iperf to test packet loss rates on this basis.

### Test Scripts
- **Client**: `iperf3 -c 10.0.0.149 -u -b 1M -t 240`
- **Server**: `iperf3 -s`

## Test Environment
- **RX ring size**: 256
- **Number of queues**: 4
- **Hardware configuration**: 4-core CPU, 16GB memory

## Test Results
| MinRxBufferPercent | Master (Packet Loss Rate) | PR (Packet Loss Rate) |
|:------------------:|:-------------------------:|:---------------------:|
| 25%                | 8%                        | 0.3%                  |
| 50%                | 0%                        | 0%                    |
| 100%               | 0%                        | 0%                    |

## Test Conclusion
For `MinRxBufferPercent` parameters around 25%, the PR version shows significant improvement in packet loss rate compared to the Master version.

80886:
About three years ago, when addressing issue #1012, I also tried kick device Immediately when rxbuffer shortage.  However, after the customer used it in production for some time, they reported high CPU usage in the guest kernel mode, it's random and low-probability, I couldn't reproduce it, so I didn't have the opportunity to conduct further analysis at the time. Theoretically, it might be related to the kick device introducing more vmexits, but it could also be due to core pinning or host task isolation. Therefore, the final solution I provided to the customer and the MinRxBufferPercent PR did not include the kick device part. 

However, this packet loss issue only occurs under specific scenarios, so the additional vmexits introduced might be acceptable. 

Overall, I think zjmletang's commit is meaningful.

zjmletang:
> About three years ago, when addressing issue #1012, I also tried kick device Immediately when rxbuffer shortage. However, after the customer used it in production for some time, they reported high CPU usage in the guest kernel mode, it's random and low-probability, I couldn't reproduce it, so I didn't have the opportunity to conduct further analysis at the time. Theoretically, it might be related to the kick device introducing more vmexits, but it could also be due to core pinning or host task isolation. Therefore, the final solution I provided to the customer and the MinRxBufferPercent PR did not include the kick device part.
> 
> However, this packet loss issue only occurs under specific scenarios, so the additional vmexits introduced might be acceptable.
> 
> Overall, I think zjmletang's commit is meaningful.

hi @80886 ,
Thanks for sharing. The idea behind my patch is that when MinRxBufferPercent is below 25%, there’s little benefit in optimizing for cases where the ring buffer is consumed by the protocol stack, because the kick threshold we send to the device is already slightly above 25%. Another option is simply to require the configurable minimum for MinRxBufferPercent to be above 25% — for example, set the minimum to 30%. If we do that, we may need to make it an enumerated setting (i.e., only allow a few predefined values) rather than a continuous numeric parameter.

PR #1455, opened 2025-11-13
viostor: Remove DmaWidth initialization in HwStorFindAdapter

zjmletang:
## Summary

Remove the `DmaWidth` initialization in `HwStorFindAdapter` to comply with Microsoft Storport driver requirements.

## Problem

According to the [Microsoft documentation](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/ns-storport-_port_configuration_information), the `DmaWidth` field in `PORT_CONFIGURATION_INFORMATION` is initialized by Storport and **miniport drivers must not modify it**. However, the current code explicitly sets `ConfigInfo->DmaWidth = Width32Bits;` in the `HwStorFindAdapter` routine, which violates this requirement.

## Changes

- Removed the line `ConfigInfo->DmaWidth = Width32Bits;` from `virtio_stor.c` 

## Testing

I tested the current behavior (with `DmaWidth` assignment) on **Windows Server 2022** and found that:
- The explicit `DmaWidth` assignment does not appear to affect 64-bit DMA functionality in practice
- The driver works correctly with 64-bit DMA addressing even when `DmaWidth` is set to `Width32Bits`
- No functional issues were observed

However, since:
1. Testing was only performed on Server 2022
2. The MSDN documentation explicitly states that miniport drivers must not modify this field
3. It's better to strictly follow Microsoft's documented requirements to ensure compatibility across all Windows versions

We should remove this assignment to comply with the official documentation.

## Reference

- [PORT_CONFIGURATION_INFORMATION structure](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/ns-storport-_port_configuration_information)

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -265,8 +265,12 @@ VirtIoFindAdapter(IN PVOID DeviceExtension,
 
     ConfigInfo->Master = TRUE;
     ConfigInfo->ScatterGather = TRUE;
+#ifndef _WIN64
+    /* On 32-bit systems, keep the old behavior to avoid unknown side effects.
+     * This will be removed when 32-bit systems are deprecated. */
     ConfigInfo->DmaWidth = Width32Bits;
     ConfigInfo->Dma32BitAddresses = TRUE;
+#endif
     ConfigInfo->Dma64BitAddresses = SCSI_DMA64_MINIPORT_FULL64BIT_SUPPORTED;
     ConfigInfo->WmiDataProvider = FALSE;
     ConfigInfo->AlignmentMask = 0x3;


YanVugenfirer:

YanVugenfirer:
@zjmletang Thank you for the patch.

I have one doubt about the patch - unfortunately we still have to support 32bit version of Windows 10, and I will be more happy to keep the old behavior for 32 bit versions or even for win10 and below in general. 

@vrozenfe what do you think?

zjmletang:
> @zjmletang Thank you for the patch.
> 
> I have one doubt about the patch - unfortunately we still have to support 32bit version of Windows 10, and I will be more happy to keep the old behavior for 32 bit versions or even for win10 and below in general.
> 
> @vrozenfe what do you think?

@YanVugenfirer
Thanks for your concern. I'd like to clarify what was changed:

**What was removed:**
- `ConfigInfo->DmaWidth = Width32Bits;` 

**What was kept:**
- `ConfigInfo->Dma32BitAddresses = TRUE;` (still present)
- `ConfigInfo->Dma64BitAddresses = SCSI_DMA64_MINIPORT_FULL64BIT_SUPPORTED;` (still present)

The change only removes the manual setting of `DmaWidth`, which according to Microsoft documentation should be initialized by Storport, not by miniport drivers. This field is explicitly documented as "Storport initializes this member and miniport drivers must not modify it."

I initially had the same concern about 32-bit system compatibility, but after reviewing the Microsoft documentation more carefully, I realized that:
- The `Dma32BitAddresses` and `Dma64BitAddresses` fields indicate the HBA hardware's DMA addressing capabilities, not the system architecture
- These settings remain unchanged, so the HBA's DMA capabilities are still properly declared
- Storport will automatically initialize `DmaWidth` with the appropriate value based on the system architecture and will handle DMA addressing accordingly

This change follows Microsoft's documented requirements and shouldn't affect functionality on any system architecture. if I misunderstood something, please let me know and I'm happy to discuss further or adjust the approach.

zjmletang:
@YanVugenfirer  Additionally, I should note that according to the MSDN documentation, `Dma32BitAddresses` also should not be set by miniport drivers - the documentation explicitly states "Storport initializes this member to TRUE" and "Miniport drivers must not modify this value". However, I chose to keep this assignment because:

1. The value we're setting (`TRUE`) is always the same as Storport's default initialization value, so there's no practical harm in setting it explicitly
2. Unlike `DmaWidth`, which could be misleading or potentially harmful (though I didn't observe any issues in my testing), `Dma32BitAddresses = TRUE` simply confirms what Storport would set anyway

If you prefer strict compliance with the documentation, we could also remove the `Dma32BitAddresses` assignment, but I believe it's less critical than removing `DmaWidth` since it doesn't introduce any discrepancy with Storport's default behavior.

YanVugenfirer:
@zjmletang I understand compliance with documentation, but I also reminder the history of those flags. And they were specifically designed for HW that might have issues with 64 bit addressing. As verifying the effect of something that worked for ages on 32bit systems is hard to estimate, and more over 32bit systems are lower priority and are going to be deprecated anyway, that's why I suggest to make the change only on 64bit systems. Where it is definitely a bug and a remnant on old historical decision.

But in order not to make unknown side effects  on 32bit systems, I suggest to use compile time decisions that will be removed later on anyway as we depreciate 32 bit systems. 

I also discussed this with @ybendito and @vrozenfe - so this is a unified position from our side.

zjmletang:
@YanVugenfirer  Understood—thanks, I learned a lot! I've updated it with a new version.

YanVugenfirer:
ok to test

YanVugenfirer:
@kostyanf14 In Win11_24H2 disk stress test failed, is this something new?

kostyanf14:
@YanVugenfirer triggered tests

kostyanf14:
@zjmletang This patch cause BSOD on Win10 22H2 on Disk Stress test https://www.dropbox.com/scl/fo/hry3ofa5x328usfhcya4m/AKWuM6rEFF4Ab5aaq4bK1mU?rlkey=h64olhjkr37btpa3jnxlo6y9o&dl=0&lst=
Please investigate.

zjmletang:
> @zjmletang This patch cause BSOD on Win10 22H2 on Disk Stress test https://www.dropbox.com/scl/fo/hry3ofa5x328usfhcya4m/AKWuM6rEFF4Ab5aaq4bK1mU?rlkey=h64olhjkr37btpa3jnxlo6y9o&dl=0&lst= Please investigate.

@kostyanf14 Alright, I’ll review it later. It’s quite strange — in theory, it shouldn’t have any impact

kostyanf14:
@zjmletang looks like we accidentally reproduced one customer issue in our HCK-CI env for Win10 22H2. We have BSODs in this env for other PR. We will investigate too.

benyamin-codez:
This one is a great idea. The compile-time conditional macro is a good compromise.

In my x64 research, I found setting `ConfigInfo->DmaWidth` increases latency and reduces IOPS. StorPort sets this to `Width8Bits (0x0)` but ignores it when `ConfigInfo->Master = true`. Manually setting `ConfigInfo->Dma32BitAddresses` also has a negative impact on performance, increasing latency (+100us latency) and reducing IOPS (-15k IOPS).

kostyanf14:
@benyamin-codez @zjmletang 

What I can say now.

Our customer reported a few very random BSODs with Windows VMs. We don't know is a Windows or a virtio drivers issue.

After some time, we got a BSOD for Win10 HLK for github.com/virtio-win/kvm-guest-drivers-windows/pull/1455 that is not related to the driver. After additional investigation, we found that Win10 and Win11 HLK setups start failing permanently, sometimes with a BSOD on these two setups. BSOD code and stack traces are very different and related to different Windows components, but we already see the same stack traces in the HLK environment and customer crash dumps. Reinstallating of Win10 HLK environment from scratch does not help. So, we are thinking that HLK fails and BSOD is not related to the HLK environment.

Currently, we have no answer to what exactly triggers BSOD, different timeouts, and disk IO errors in HLK for Win10 and Win11 environments.
There were no changes in software (both host and guest) and hardware between the time without and with the issue.

zjmletang:
> @benyamin-codez @zjmletang
> 
> What I can say now.
> 
> Our customer reported a few very random BSODs with Windows VMs. We don't know is a Windows or a virtio drivers issue.
> 
> After some time, we got a BSOD for Win10 HLK for github.com/[/pull/1455](https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1455) that is not related to the driver. After additional investigation, we found that Win10 and Win11 HLK setups start failing permanently, sometimes with a BSOD on these two setups. BSOD code and stack traces are very different and related to different Windows components, but we already see the same stack traces in the HLK environment and customer crash dumps. Reinstallating of Win10 HLK environment from scratch does not help. So, we are thinking that HLK fails and BSOD is not related to the HLK environment.
> 
> Currently, we have no answer to what exactly triggers BSOD, different timeouts, and disk IO errors in HLK for Win10 and Win11 environments. There were no changes in software (both host and guest) and hardware between the time without and with the issue.

hi @kostyanf14 
May I ask if this issue is consistently reproducible on your side? On my side, running this HLK test on Windows 10 passes successfully.
[viostor_pr_win10.log](https://github.com/user-attachments/files/24205803/viostor_pr_win10.log)

kostyanf14:
> May I ask if this issue is consistently reproducible on your side? On my side, running this HLK test on Windows 10 passes successfully.

Yes, we see this issue consistently now.

zjmletang:
> This one is a great idea. The compile-time conditional macro is a good compromise.
> 
> In my x64 research, I found setting `ConfigInfo->DmaWidth` increases latency and reduces IOPS. StorPort sets this to `Width8Bits (0x0)` but ignores it when `ConfigInfo->Master = true`. Manually setting `ConfigInfo->Dma32BitAddresses` also has a negative impact on performance, increasing latency (+100us latency) and reducing IOPS (-15k IOPS).

@benyamin-codez thanks.
May I ask if there is any theoretical explanation for this performance improvement? I’m not very clear about the underlying principle. At the time, my intention was only to check that the settings of DmaWidth and Dma32BitAddresses did not conform to the MSDN specification, and I was worried about introducing unknown risks.

benyamin-codez:
> May I ask if there is any theoretical explanation for this performance improvement? I’m not very clear about the underlying principle. At the time, my intention was only to check that the settings of DmaWidth and Dma32BitAddresses did not conform to the MSDN specification, and I was worried about introducing unknown risks.

I can only speculate, but I assume it is due to additional code paths being employed when these are specified.

The documentation is often revised and what was once recommended does change with time.
This leaves working code with the appearance of being improperly implemented when this is generally not the case.  
There have also been some historical errors, which has negatively affected the reputation of some of the documentation.

zjmletang:
Just rebased

YanVugenfirer:
rerun tests

YanVugenfirer (approved):

PR #1471, opened 2025-12-18
Fix case-sensitive filename issues for builds on EWDK 25H2 [SKIP-HCK-CI]

xz-dev:
Rename files to match their #include references when building on case-sensitive filesystems (e.g., Linux shared folders by virtiofs to windows):
- VirtIO/VirtIO.h -> VirtIO/virtio.h (matches #include "virtio.h")
- build/rhel.ver -> build/RHEL.ver (matches VENDOR_VER=$(_VENDOR_).ver)

Test case:
1. git clone in Linux host then use virtiofs share dir to windows guest
2. Download [EWDK](https://learn.microsoft.com/en-us/legal/windows/hardware/enterprise-wdk-license-2022) in windows guest
3. Setup EWSK env and run `build_all_drivers.bat`


YanVugenfirer:

YanVugenfirer:
Hi @xz-dev ,

Thanks for the PR!

Couple of questions:
1. Did you try to build with Visual Studio to see that the change in the variable name is not breaking VS support?
2. Can you please explain your use case?
3. As we cannot guarantee case sensitivity, isn't it better to have some script that renames the files before the build? this way we will keep git history in place.


Thanks,
Yan.

YanVugenfirer:
ok to test

xz-dev:
> Hi @xz-dev ,
> 
> Thanks for the PR!
> 
> Couple of questions:
> 
>     1. Did you try to build with Visual Studio to see that the change in the variable name is not breaking VS support?
> 
>     2. Can you please explain your use case?
> 
>     3. As we cannot guarantee case sensitivity, isn't it better to have some script that renames the files before the build? this way we will keep git history in place.
> 
> 
> Thanks, Yan.

I think if you use a script to rename at compile time, it might cause a change in git status.

Test case is I not install full vs dev env in Windows, because I want to keep guest env clean.

So I use IDE in Linux host to edit code, and then use share filesystem (virtiofs) to let I can directly build it and test it in Windows guest without always copy from host.

YanVugenfirer:
* You can configure "core.ignoreCase"
* Please check compilation with VS
* Thanks you for exampling your use case

xz-dev:
>As we cannot guarantee case sensitivity, isn't it better to have some script that renames the files before the build? this way we will keep git history in place.

Thanks for review, oh, I got what you meaning.
Regarding keeping git history in place - since git detects this change as a rename (shown as rename in the diff), the line-by-line history is actually preserved. You can verify this with:
  - git log --follow VirtIO/virtio.h - traces full history through the rename
  - git blame VirtIO/virtio.h - correctly attributes each line to the original authors and dates

However, if you believe “core.ignoreCase” is the better approach for this project, I'm willing to try it and report back whether it resolves the issue in my setup.

I will test with Visual Studio and report back.

YanVugenfirer:
@xz-dev BTW: I am looking for team members, are you looking to work on open source projects (not only Windows) professionally ? If yes, please send me your contact details to my Red Hat email: yvugenfi (at) redhat (dot) com

xz-dev:
Thanks for your review this!

**Testing confirmation:**

Tested in viofs, but it's building too slow.
图片

Also simply clone to local C: then build it.
图片

**Context on discovery:**

I actually discovered this issue while working on viogpu auto-scaling functionality. 
The feature is already implemented, but appears to be disabled due to some edge cases 
causing failures (like [this line](https://github.com/virtio-win/kvm-guest-drivers-windows/blob/8f73d3ad38e2aeb38ff3bee136b4048600df8c42/viogpu/common/viogpu_queue.cpp#L338)). 
I'm currently investigating these edge cases and working on fixes.

Looking forward to contributing more improvements to the virtio-win ecosystem.

P.S. I've replied to your email. Thank you for reaching out!

xz-dev:
Closing this PR due to my mistake of creating it from the master branch. Given your core.ignoreCase suggestion, this change may not be needed. I'll create a new PR only if necessary. Thanks!

PR #1474, opened 2025-12-27
RHEL-149886: [viogpu] Reject resolutions exceeding framebuffer segment capacity

xz-dev:
Add resolution validation in IsSupportedVidPn to prevent switching to resolutions that exceed the framebuffer segment size. This fixes an issue where dynamically resizing to a large resolution (e.g., via virt-manager window resize) would fail in CreateFrameBufferObj, leaving the display in an unrecoverable state.

Previously, when the requested resolution required more memory than the pre-allocated framebuffer segment, the driver would:
1. Destroy the existing working framebuffer
2. Fail to create the new larger framebuffer
3. Leave the display inactive with no way to recover

Now the driver checks the required framebuffer size against the segment capacity in IsSupportedVidPn and rejects unsupported resolutions early, allowing Windows to gracefully handle the situation.

### DIFF
--- a/viogpu/viogpudo/viogpudo.cpp
+++ b/viogpu/viogpudo/viogpudo.cpp
@@ -783,7 +783,7 @@ NTSTATUS VioGpuDod::IsSupportedVidPn(_Inout_ DXGKARG_ISSUPPORTEDVIDPN *pIsSuppor
 
         if (bReject)
         {
-            return STATUS_SUCCESS; // IsVidPnSupported remains FALSE
+            return STATUS_NO_MEMORY;
         }
     }
 


YanVugenfirer:

viogpu/viogpudo/viogpudo.cpp
```diff
@@ -721,6 +721,62 @@ NTSTATUS VioGpuDod::IsSupportedVidPn(_Inout_ DXGKARG_ISSUPPORTEDVIDPN *pIsSuppor
...
+                          Width,
+                          Height,
+                          (ULONGLONG)RequiredSize,
+                          (ULONGLONG)SegmentSize));
+                pVidPnInterface->pfnReleaseSourceModeSet(pIsSupportedVidPn->hDesiredVidPn, hVidPnSourceModeSet);
+                return STATUS_SUCCESS;  // IsVidPnSupported remains FALSE
+            }
+        }
+
+        pVidPnInterface->pfnReleaseSourceModeSet(pIsSupportedVidPn->hDesiredVidPn, hVidPnSourceModeSet);
```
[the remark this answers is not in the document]
xz-dev: fixed

YanVugenfirer (changes requested):
@xz-dev Sorry, I missed that the "Signed-off-by: Name <email>" line is missing. Please add in all commits. It can be automated by git when you do "git commit" add "-s" command line parameter. 

Thanks,
Yan

xz-dev:
Code formatted

spurin:
@xz-dev - am keen to make use of this. In the interim, whilst this is progressing, do you happen to have a recent build you can share for testing, thanks.

xz-dev:
> @xz-dev - am keen to make use of this. In the interim, whilst this is progressing, do you happen to have a recent build you can share for testing, thanks.

Of course, but my self-built version cannot use Red Hat signing.

Also, could you tell me what CPU architecture and Windows version you are using? (I can first double-check that the build artifacts are intact, after I have completed the build.)

viogpu/viogpudo/viogpudo.cpp
```diff
@@ -722,6 +722,48 @@ NTSTATUS VioGpuDod::IsSupportedVidPn(_Inout_ DXGKARG_ISSUPPORTEDVIDPN *pIsSuppor
         {
             return STATUS_SUCCESS;
         }
+
+        // Check if resolution exceeds framebuffer segment capacity
+        if (NumPathsFromSource > 0)
```
kostyanf14: I preferred to prevent `if` in `if` in `if` where this is possible. 

- update `if (NumPathsFromSource > 0)` to `if (NumPathsFromSource == 0) then continue` 
- return old logic with `pfnAcquireSourceModeSet` and `if (!NT_SUCCESS(Status))`

viogpu/viogpudo/viogpudo.cpp
```diff
@@ -722,6 +722,48 @@ NTSTATUS VioGpuDod::IsSupportedVidPn(_Inout_ DXGKARG_ISSUPPORTEDVIDPN *pIsSuppor
...
+                                  (ULONGLONG)SegmentSize));
+                        bReject = TRUE;
+                    }
+                    pVidPnSourceModeSetInterface->pfnReleaseModeInfo(hVidPnSourceModeSet, pPinnedVidPnSourceModeInfo);
+                }
+                pVidPnInterface->pfnReleaseSourceModeSet(pIsSupportedVidPn->hDesiredVidPn, hVidPnSourceModeSet);
+
+                if (bReject)
+                {
+                    return STATUS_SUCCESS; // IsVidPnSupported remains FALSE
```
kostyanf14: Should we return `STATUS_NO_MEMORY` (from MSDN: `The function failed because it was unable to allocate memory.`) in this case? From one point of view, we do not have enough memory, but from another, we don't try to allocate more.

viogpu/viogpudo/viogpudo.cpp
xz-dev: https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1474/commits/3462b57f46b2e545e25a5407254722e633dbb1dd

kostyanf14:
ok to test

viogpu/viogpudo/viogpudo.cpp
kostyanf14: @xz-dev What do you think? Does this make sense or not?

kostyanf14:
@spurin, after 8h full build, you can get the binary from CI

kostyanf14:
@xz-dev Can you please write the steps to reproduce the issue simply?

viogpu/viogpudo/viogpudo.cpp
xz-dev: I tested both approaches:
- STATUS_SUCCESS (with IsVidPnSupported = FALSE): Windows treats it as "unsupported VidPn" and falls back to a scaled resolution.
- STATUS_NO_MEMORY: Windows keeps the current resolution unchanged.
Both prevent the unrecoverable black screen. I think STATUS_NO_MEMORY is more appropriate here — we genuinely don't have enough framebuffer segment memory for the requested resolution, and keeping the current resolution unchanged is the correct behavior in this case.

Updated to return STATUS_NO_MEMORY.

xz-dev:
## Steps to Reproduce

**Environment:**
- Host OS: Linux with KDE Plasma on Wayland
- Display: 4K monitor (or any high-resolution display)
- Virtualization: virt-manager/QEMU-KVM
- Guest OS: Windows 11 with ·viogpudo· driver installed
- viogpu configuration: ·vgpusrv· user service enabled

**Steps:**
1. Launch virt-manager and start the Windows VM with ·viogpu· graphics device
2. Wait for the guest to boot and log in to the Windows desktop
3. Switch the virt-manager window to fullscreen mode (or manually resize the window to a large resolution)
4. Continue increasing the resolution by stretching the window

**Expected behavior:**
The display should either maintain the current resolution or gracefully reject unsupported resolutions.

**Actual behavior:**
Once the resolution reaches a certain threshold (exceeding framebuffer capacity), the screen goes black. Reducing the resolution does not restore the display. The only recovery method is to restart the viogpu driver.

PR #1485, opened 2026-01-06
RHEL-137735: [viosock] Fix vsock connectivity and BSOD issue after VM migration

leidwang:
Problem                                                                                                                               
                                                                                                                                        
  The viosock driver experiences connectivity issues and potential BSOD crashes after VM migration due to improper handling of transport
   reset events. The current implementation has several critical issues:                                                                
                                                                                                                                        
  - Event size validation bug causing incorrect packet processing                                                                       
  - Infinite loop in transport reset handler causing system hang                                                                        
  - Race conditions during device state transitions                                                                                     
  - Missing queue cleanup leading to memory corruption                                                                                  
  - Improper socket state management after CID changes
 
Solution                                                                                                                              
                                                                                                                                        
  This PR implements a comprehensive fix through two focused commits:                                                                   
                                                                                                                                        
  1. Transport Reset Infrastructure and Bug Fixes                                                                                       
  - ✅ Fix critical event size validation bug (sizeof(*pEvt) vs sizeof(pEvt))                                                           
  - ✅ Move transport reset handling to workitem for PASSIVE_LEVEL processing                                                           
  - ✅ Add DeviceReady synchronization to prevent null pointer access                                                                   
  - ✅ Add defensive null checks in transmit path during reset                                                                          
  - ✅ Fix missing VOID return type in function declaration                                                                             
                                                                                                                                        
  2. Complete Transport Reset Mechanism                                                                                                 
  - ✅ Implement complete VirtIO queue rebuilding during transport reset                                                                
  - ✅ Add three-layer defense: queue state reset + socket handling + queue rebuild                                                     
  - ✅ Detect guest-CID changes and handle protocol mismatch after migration                                                            
  - ✅ Close connected sockets and update listen sockets appropriately                                                                  
  - ✅ Add queue draining to prevent memory leaks during device shutdown                                                                
  - ✅ Add defensive cleanup for corrupted socket collections                                                                           
                                                                                                                                        
  Key Technical Changes                                                                                                                 
                                                                                                                                        
  - Transport Reset Workitem: Prevents infinite loops by processing resets at PASSIVE_LEVEL                                             
  - Queue Rebuilding: Complete teardown and reconstruction of VirtIO queues after reset                                                 
  - Socket State Management: Proper handling of connected/listening sockets during migration                                            
  - Synchronization: Device ready flag prevents operations during unsafe states                                                         
  - Memory Safety: Queue draining and defensive cleanup prevent leaks and corruption                                                                                                                              
                                                                                                                                        
  Files Modified                                                                                                                        
                                                                                                                                        
  - viosock/sys/Device.c - Infrastructure, workitem, device lifecycle                                                                   
  - viosock/sys/Evt.c - Event processing and transport reset detection                                                                  
  - viosock/sys/Tx.c - Transmit path synchronization and null checks                                                                    
  - viosock/sys/Rx.c - Receive path, queue draining, CID validation                                                                     
  - viosock/sys/Socket.c - Socket state management during reset                                                                         
  - viosock/sys/viosock.h - New declarations and structure members                                                                      
                                                                                                                                        
  Impact                                                                                                                                
                                                                                                                                        
  This fix resolves the VM migration BSOD and connectivity issues, ensuring stable vsock operation across migration events while maintaining backward compatibility.

### DIFF
--- a/viosock/sys/Rx.c
+++ b/viosock/sys/Rx.c
@@ -495,6 +495,46 @@ _Requires_lock_not_held_(pContext->RxLock) VOID VIOSockRxVqCleanup(IN PDEVICE_CO
     TraceEvents(TRACE_LEVEL_VERBOSE, DBG_HW_ACCESS, "<-- %s\n", __FUNCTION__);
 }
 
+VOID VIOSockRxVqDrainAllPackets(IN PDEVICE_CONTEXT pContext)
+{
+    PVIOSOCK_RX_PKT pPkt;
+    PSINGLE_LIST_ENTRY pListEntry;
+    ULONG packetsRemoved = 0;
+
+    TraceEvents(TRACE_LEVEL_WARNING, DBG_HW_ACCESS, "--> %s\n", __FUNCTION__);
+
+    if (!pContext->RxVq)
+    {
+        TraceEvents(TRACE_LEVEL_WARNING, DBG_HW_ACCESS, "<-- %s: No RxVq to drain\n", __FUNCTION__);
+        return;
+    }
+
+    WdfSpinLockAcquire(pContext->RxLock);
+
+    // Drain all unused buffers from virtqueue
+    while (pPkt = (PVIOSOCK_RX_PKT)virtqueue_detach_unused_buf(pContext->RxVq))
+    {
+        VIOSockRxPktCleanup(pContext, pPkt);
+        packetsRemoved++;
+    }
+
+    // Clear all postponed packets
+    while (pListEntry = PopEntryList(&pContext->RxPktList))
+    {
+        pPkt = CONTAINING_RECORD(pListEntry, VIOSOCK_RX_PKT, RxPktListEntry);
+        VIOSockRxPktCleanup(pContext, pPkt);
+        packetsRemoved++;
+    }
+
+    WdfSpinLockRelease(pContext->RxLock);
+
+    TraceEvents(TRACE_LEVEL_WARNING,
+                DBG_HW_ACCESS,
+                "<-- %s: Drained %d packets from receive queue\n",
+                __FUNCTION__,
+                packetsRemoved);
+}
+
 NTSTATUS
 VIOSockRxVqInit(IN PDEVICE_CONTEXT pContext)
 {
@@ -965,7 +1005,19 @@ _Requires_lock_not_held_(pContext->RxLock) VOID VIOSockRxVqProcess(IN PDEVICE_CO
             continue;
         }
 
-        ASSERT(pContext->Config.guest_cid == (ULONG32)pPkt->Header.dst_cid);
+        if (pContext->Config.guest_cid != (ULONG32)pPkt->Header.dst_cid)
+        {
+            TraceEvents(TRACE_LEVEL_CRITICAL,
+                        DBG_SOCKET,
+                        "Packet CID mismatch: expected %d, got %d\n",
+                        (ULONG)pContext->Config.guest_cid,
+                        (ULONG)pPkt->Header.dst_cid);
+
+            ASSERT(FALSE);
+
+            VIOSockRxPktInsertOrPostpone(pContext, pPkt);
+            continue;
+        }
 
         TraceEvents(TRACE_LEVEL_INFORMATION,
                     DBG_SOCKET,
--- a/viosock/sys/Socket.c
+++ b/viosock/sys/Socket.c
@@ -525,7 +525,6 @@ VOID VIOSockPendedTimerFunc(WDFTIMER Timer)
     TraceEvents(TRACE_LEVEL_VERBOSE, DBG_SOCKET, "--> %s\n", __FUNCTION__);
 
     //     ASSERT(VIOSockStateGet(pSocket) == VIOSOCK_STATE_CONNECTING);
-
     if (NT_SUCCESS(VIOSockPendedRequestGetLocked(pSocket, &PendedRequest)) && PendedRequest != WDF_NO_HANDLE)
     {
         WdfRequestComplete(PendedRequest, STATUS_IO_TIMEOUT);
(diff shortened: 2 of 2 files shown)


YanVugenfirer:

leidwang:
Hi @kostyanf14 Could you please review this PR?Thanks!

viosock/sys/Evt.c
```diff
@@ -179,10 +179,41 @@ _IRQL_requires_max_(DISPATCH_LEVEL) VOID VIOSockEvtVqProcess(IN PDEVICE_CONTEXT
...
+                        oldGuestCid,
+                        pContext->Config.guest_cid);
+
+            // During migration, the guest-CID changes which invalidates all existing sockets
+            // because they were bound/connected using the old CID. We must close everything
+            // to allow applications to reconnect with the new CID.
+            TraceEvents(TRACE_LEVEL_WARNING,
+                        DBG_HW_ACCESS,
+                        "Closing all sockets due to guest-CID change - applications must reconnect\n");
+        }
```
kostyanf14: I am confused. You write "CRITICAL:" comment and log "Closing all sockets due to guest-CID change - applications must reconnect" but no action is taken, just logs.

viosock/sys/Rx.c
```diff
@@ -965,7 +965,20 @@ _Requires_lock_not_held_(pContext->RxLock) VOID VIOSockRxVqProcess(IN PDEVICE_CO
             continue;
         }
 
-        ASSERT(pContext->Config.guest_cid == (ULONG32)pPkt->Header.dst_cid);
+        // SAFETY: Handle CID mismatch gracefully (can happen during/after migration)
+        // The original ASSERT could cause BSOD when packets arrive with old CID after migration
+        if (pContext->Config.guest_cid != (ULONG32)pPkt->Header.dst_cid)
```
kostyanf14, line 1008: I think this should not happen
The device/driver should reset the queue and remove all requests when get VIRTIO_VSOCK_EVENT_TRANSPORT_RESET event, and there should not be a request with a CID mismatch.
If it exists, we do something wrong in reset level logic, and this is just a workaround

viosock/sys/Socket.c
```diff
@@ -2563,17 +2563,62 @@ VIOSockGetSocketFromHandle(IN PDEVICE_CONTEXT pContext, IN ULONGLONG uSocket, IN
 VOID VIOSockHandleTransportReset(IN PDEVICE_CONTEXT pContext)
 {
     WDFFILEOBJECT Socket;
-    TraceEvents(TRACE_LEVEL_VERBOSE, DBG_HW_ACCESS, "--> %s\n", __FUNCTION__);
+    WDFOBJECT socketsToClose[256]; // Assuming a reasonable max number of sockets
```
kostyanf14: `ItemCount = WdfCollectionGetCount(pContext->ConnectedList);`

viosock/sys/Socket.c
```diff
@@ -2563,17 +2563,62 @@ VIOSockGetSocketFromHandle(IN PDEVICE_CONTEXT pContext, IN ULONGLONG uSocket, IN
...
+    socketCount = WdfCollectionGetCount(pContext->ConnectedList);
+    if (socketCount > (sizeof(socketsToClose) / sizeof(socketsToClose[0])))
+    {
+        TraceEvents(TRACE_LEVEL_CRITICAL,
+                    DBG_HW_ACCESS,
+                    "Too many sockets to reset (%d), trimming to %d\n",
+                    socketCount,
+                    (ULONG)(sizeof(socketsToClose) / sizeof(socketsToClose[0])));
+        socketCount = (ULONG)(sizeof(socketsToClose) / sizeof(socketsToClose[0]));
+    }
```
kostyanf14: We already have `VIOSockConnectedEnum`, please check the possibility of using this API

viosock/sys/Socket.c
```diff
@@ -2563,17 +2563,62 @@ VIOSockGetSocketFromHandle(IN PDEVICE_CONTEXT pContext, IN ULONGLONG uSocket, IN
...
+        WdfSpinLockAcquire(pCurrentSocket->StateLock);
+
+        if (VIOSockStateGet(pCurrentSocket) == VIOSOCK_STATE_CONNECTED)
+        {
+            // VIOSockStateSet will call VIOSockConnectedRemove, which acquires ConnectedLock.
+            // This is now safe as we are not holding it.
+            VIOSockStateSet(pCurrentSocket, VIOSOCK_STATE_CLOSE);
+            // VIOSockEventSetBit must be called with StateLock held.
+            VIOSockEventSetBit(pCurrentSocket, FD_CLOSE_BIT, STATUS_CONNECTION_RESET);
+        }
```
kostyanf14: You are closing all currently connected sockets, but according to the virtio spec v1.3 section 5.10.6.7  Device Events, there are 2 different actions: `The driver shuts down established connections` and `Existing listen sockets remain but their CID is updated`

```
The VIRTIO_VSOCK_EVENT_TRANSPORT_RESET event indicates that communication has been
interrupted. This usually occurs if the guest has been physically migrated. The driver shuts down
established connections and the guest_cid configuration field is fetched again. Existing listen sockets
remain but their CID is updated to reflect the current guest_cid. 
```

kostyanf14:
BSOD stack 
image

Tx.c#L333: `virtqueue_disable_cb(pContext->TxVq);`
https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/viosock/sys/Tx.c#L333

kostyanf14:
If I got the correct function arguments from the  stack, the broken address 0x50 (very low value) 
image

kostyanf14:
ok to test

kostyanf14:
rerun tests

leidwang:
rerun tests

kostyanf14:
rerun tests

leidwang:
rerun tests

leidwang:
Hi @kostyanf14 Would you please review the code again? It's working perfectly functionally right now, but I'd still like to confirm if there are any other issues. Thanks!

leidwang:
> @leidwang Can you please break the PR into smaller commits. Basically each bullet point in commit message should be a separate commit in my opinion. Smaller commits make it easier for review.

Yes, I completely agree. I've updated the PR, splitting it into two commits, and updated the PR description to provide more details of the code changes.Thanks @YanVugenfirer

YanVugenfirer:
rerun tests

YanVugenfirer:
Hi @leidwang. 

After initial review with Konstantin, we have several questions:
1. Can you make a take a trace of the driver during the migration? 
2. The reason we ask, it appears that for some reason either D0exit is called or for some other reasons virtio queues become NULL. We want to understand what actually happens.
3.  Not for this PR - but the operation of the driver during S3\S4 should be checked.

leidwang:
> Hi @leidwang.
> 
> After initial review with Konstantin, we have several questions:
> 
> 1. Can you make a take a trace of the driver during the migration?
> 2. The reason we ask, it appears that for some reason either D0exit is called or for some other reasons virtio queues become NULL. We want to understand what actually happens.
> 3. Not for this PR - but the operation of the driver during S3\S4 should be checked.

Hi @YanVugenfirer, thanks for the review. 
Do you mean I need to collect the trace during the migration or add some function to take the trace during migration?

YanVugenfirer:
@leidwang Please collect the trace during the migration.

leidwang:
> @leidwang Please collect the trace during the migration.

I tried using the tools in Tools\trace to collect traces, but it seems I didn't collect any useful information. To trigger a blue screen, I need to restart the VM, but restarting the VM will interrupt the trace collection. I will add trace file and dump file together, could you please take a look?

Trace file and dump file: https://drive.google.com/drive/folders/1rZ4H4v7mELI4DJVZI7F04-m5anTI2Ey4

YanVugenfirer:
@kostyanf14 Do you know if we can redirect virtio-sock trace to debug space now, or this feature doesn't exist in virtio-sock?

YanVugenfirer:
@leidwang for vsock checkout utils.c file for debug printing based on InitializeDebugPrints you can force the driver to print to debug space or even serial port.
Serial port you can redirect on QEMU level, DebugSpace you can connect WinDBG to the VM.
But it looks that you need to recompile the driver to enable serial port debugging or debug space 

If you choose to use debug space, use DebugView to extract traces from the crash dump: https://learn.microsoft.com/en-us/sysinternals/downloads/debugview#:~:text=from%20Sysinternals%20Live.-,Introduction,non%2Dstandard%20debug%20output%20APIs.

```
void InitializeDebugPrints(IN PDRIVER_OBJECT  DriverObject, PUNICODE_STRING RegistryPath)
{
    WPP_INIT_TRACING(DriverObject, RegistryPath);
    //TODO - Read nDebugLevel and bDebugPrint from the registry
#if defined(EVENT_TRACING)
    VirtioDebugPrintProc = DebugPrintFuncWPP;
#elif defined(PRINT_DEBUG)
    VirtioDebugPrintProc = DebugPrintFunc;
#elif defined(COM_DEBUG)
    VirtioDebugPrintProc = DebugPrintFuncSerial;
#else
    VirtioDebugPrintProc = NoDebugPrintFunc;
#endif

    UNREFERENCED_PARAMETER(DriverObject);
    UNREFERENCED_PARAMETER(RegistryPath);

    bDebugPrint = 1;
    virtioDebugLevel = 1;

#if !defined(EVENT_TRACING)
    driverDebugFlags = 0xffffffff;
#if DBG
    driverDebugLevel = TRACE_LEVEL_VERBOSE;
#else
    driverDebugLevel = TRACE_LEVEL_INFORMATION;
#endif
#endif
}
```

kostyanf14:
@YanVugenfirer I think WPP (default in Release) will be better. Start traceview, and logs will be persistent during migration, just need to have the exact timestamp when VMswitch occurs to understand logs before and after migration.

kostyanf14:
With debugger, this is an unclear point of the kernel debugger reconnect

YanVugenfirer:
> With debugger, this is an unclear point of the kernel debugger reconnect

Right... we have migration in the middle.

PR #1493, opened 2026-01-20
viostor: Reimplement bus reset operation

kostyanf14:
The VirtIO BLK device does not have special reset events/routines
like the VirtIO SCSI device (controlq command VIRTIO_SCSI_T_TMF_LOGICAL_UNIT_RESET),
but Windows requires to have some mechanism to fix a device when it hangs.
The most appropriate way is perform a full VirtIO reset and initialization
flow based on spec.

The following flow is implemented:

1. Reset VirtIO device (virtio_device_reset)
2. Delete VirtIO queues (virtio_delete_queues)
3. Clean up all device memory (virtio_device_shutdown)
4. Complete all pending SRBs (in guest, CompletePendingRequestsOnReset)
5. Perform VirtIO device initialization (virtio_device_initialize, virtio_find_queues, etc)

Note: We don't pause StorPort because the port driver pauses all device IO queues for the
adapter and then calls the HwStorResetBus routine
https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nc-storport-hw_reset_bus

Note: We cannot just complete all pending SRBs and continue, because this
memory will be still in use by the device and free in Windows. As a result,
we got use-after-free when the device completes the request and sends it back to Windows.
This may cause a memory corruption, as Windows may already have reused the memory.

Before this commit, we only complete all pending SRBs and continue.

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -454,6 +454,7 @@ VirtIoFindAdapter(IN PVOID DeviceExtension,
     {
         adaptExt->num_queues = min(adaptExt->num_queues, (USHORT)num_cpus);
     }
+    adaptExt->reset_in_progress_count = 0;
 
     RhelDbgPrint(TRACE_LEVEL_INFORMATION, " Queues %d CPUs %d\n", adaptExt->num_queues, num_cpus);
 
@@ -839,54 +840,117 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
     return ret;
 }
 
-VOID CompletePendingRequests(IN PVOID DeviceExtension)
+static VOID CompletePendingRequestsOnReset(IN PVOID DeviceExtension)
 {
     PADAPTER_EXTENSION adaptExt;
 
     adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
 #ifdef DBG
-    RhelDbgPrint(TRACE_LEVEL_INFORMATION, "CompletePendingRequests %d %d\n", adaptExt->srb_cnt, adaptExt->inqueue_cnt);
+    RhelDbgPrint(TRACE_LEVEL_INFORMATION,
+                 "CompletePendingRequestsOnReset %d %d\n",
+                 adaptExt->srb_cnt,
+                 adaptExt->inqueue_cnt);
 #endif
-    if (!adaptExt->reset_in_progress)
+    for (ULONG index = 0; index < adaptExt->num_queues; index++)
     {
-        adaptExt->reset_in_progress = TRUE;
-        StorPortPause(DeviceExtension, 10);
-
-        for (ULONG index = 0; index < adaptExt->num_queues; index++)
+        PREQUEST_LIST element;
+        STOR_LOCK_HANDLE LockHandle = {0};
+        ULONG MessageID = index + 1;
+        VioStorVQLock(DeviceExtension, MessageID, &LockHandle, FALSE);
+        element = &adaptExt->processing_srbs[index];
+        while (!IsListEmpty(&element->srb_list))
         {
-            PREQUEST_LIST element = &adaptExt->processing_srbs[index];
-            STOR_LOCK_HANDLE LockHandle = {0};
-            ULONG MessageID = index + 1;
-            VioStorVQLock(DeviceExtension, MessageID, &LockHandle, FALSE);
-            while (!IsListEmpty(&element->srb_list))
+            PLIST_ENTRY entry = RemoveHeadList(&element->srb_list);
+            if (entry)
             {
-                PLIST_ENTRY entry = RemoveHeadList(&element->srb_list);
-                if (entry)
+                pblk_req req = CONTAINING_RECORD(entry, blk_req, list_entry);
+                PSCSI_REQUEST_BLOCK Srb = (PSCSI_REQUEST_BLOCK)req->req;
+                if (Srb)
                 {
-                    pblk_req req = CONTAINING_RECORD(entry, blk_req, list_entry);
-                    PSCSI_REQUEST_BLOCK currSrb = (PSCSI_REQUEST_BLOCK)req->req;
-                    PSRB_EXTENSION currSrbExt = SRB_EXTENSION(currSrb);
-                    if (currSrb)
-                    {
-                        SRB_SET_DATA_TRANSFER_LENGTH(currSrb, 0);
-                        CompleteRequestWithStatus(DeviceExtension, (PSRB_TYPE)currSrb, SRB_STATUS_BUS_RESET);
-                        element->srb_cnt--;
-                    }
+                    SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
+                    CompleteRequestWithStatus(DeviceExtension, (PSRB_TYPE)Srb, SRB_STATUS_BUS_RESET);
+                    element->srb_cnt--;
                 }
             }
-            if (element->srb_cnt)
-            {
-                element->srb_cnt = 0;
-            }
-            VioStorVQUnlock(DeviceExtension, MessageID, &LockHandle, FALSE);
         }
-        StorPortResume(DeviceExtension);
+        element->srb_cnt = 0;
+        VioStorVQUnlock(DeviceExtension, MessageID, &LockHandle, FALSE);
     }
-    else
+}
+
+/*
+ * The VirtIO BLK device does not have special reset events/routines
+ * like the VirtIO SCSI device (controlq command VIRTIO_SCSI_T_TMF_LOGICAL_UNIT_RESET),
+ * but Windows requires to have some mechanism to fix a device when it hangs.
+ * The most appropriate way is perform a full VirtIO reset and initialization
+ * flow based on spec.
+ *
+ * The following flow is implemented:
+ *
+ * 1. Reset VirtIO device (virtio_device_reset)
+ * 2. Delete VirtIO queues (virtio_delete_queues)
+ * 3. Clean up all device memory (virtio_device_shutdown)
+ * 4. Complete all pending SRBs (in guest, CompletePendingRequestsOnReset)
+ * 5. Perform VirtIO device initialization (virtio_device_initialize, virtio_find_queues, etc)
+ *
+ * Note: We don't pause StorPort because the port driver pauses all device IO queues for the
+ * adapter and then calls the HwStorResetBus routine
+ * https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nc-storport-hw_reset_bus
+ *
+ * Note: We cannot just complete all pending SRBs and continue, because this
+ * memory will be still in use by the device and free in Windows. As a result,
+ * we got use-after-free when the device completes the request and sends it back to Windows.
+ * This may cause a memory corruption, as Windows may already have reused the memory.
+ */
+BOOLEAN VioStorResetBus(IN PVOID DeviceExtension)
+{
+    PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
+    if (adaptExt->reset_in_progress_count)
     {
-        RhelDbgPrint(TRACE_LEVEL_ERROR, "RESET IN THE PROGRESS !!!!\n");
+        RhelDbgPrint(TRACE_LEVEL_ERROR, "VioStorResetBus RESET ALREADY IN PROGRESS !!!!\n");
+        return FALSE;
     }
-    adaptExt->reset_in_progress = FALSE;
+
+    // MSFT reports reset as a warning event in event logs, so keep the same level there
+    RhelDbgPrint(TRACE_LEVEL_WARNING, " VioStorResetBus initiated\n");
+
+    /*
+     * Try to lock each Queue/DPC to make sure there is no requests that
+     * are currently in progress in DPC. Each new DPC call will have no effect because
+     * VioStorCompleteRequest will exit for any non-zero reset_in_progress_count value
+     */
+    for (ULONG MessageId = 1; MessageId <= adaptExt->num_queues; MessageId++)
+    {
+        STOR_LOCK_HANDLE LockHandle = {0};
+        VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
+        adaptExt->reset_in_progress_count++;
+        VioStorVQUnlock(DeviceExtension, MessageId, &LockHandle, FALSE);
+    }
+
+    RhelShutDown(DeviceExtension);
+    CompletePendingRequestsOnReset(DeviceExtension);
+
+    if (!VirtIoHwReinitialize(DeviceExtension))
+    {
+        RhelDbgPrint(TRACE_LEVEL_FATAL, " VioStorResetBus cannot reinitialize HW\n");
+        /*
+         * The device is terminally broken in this case. There is no known things to do,
+         * so keep everything as is:
+         *   reset_in_progress_count > 0
+         *   VirtIoStartIo will reject all requests with SRB_STATUS_BUS_RESET.
+         */
+        return FALSE;
+    }
+
+    for (ULONG MessageId = 1; MessageId <= adaptExt->num_queues; MessageId++)
+    {
+        STOR_LOCK_HANDLE LockHandle = {0};
+        VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
+        adaptExt->reset_in_progress_count--;
+        VioStorVQUnlock(DeviceExtension, MessageId, &LockHandle, FALSE);
+    }
+
+    return TRUE;
 }
 
 BOOLEAN
@@ -899,6 +963,22 @@ VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
     adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
 
     SRB_SET_SCSI_STATUS(((PSRB_TYPE)Srb), ScsiStatus);
+
+    /*
+     * https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nc-storport-hw_reset_bus
+     * The port driver pauses all device IO queues for the adapter and then
+     * calls the HwStorResetBus routine at IRQL DISPATCH_LEVEL after acquiring
+     * the StartIo spin lock. A miniport driver is responsible for completing
+     * SRBs received by HwStorStartIo for PathId during this routine and setting
+     * their status to SRB_STATUS_BUS_RESET if necessary.
+     */
+    if (adaptExt->reset_in_progress_count)
+    {
+        RhelDbgPrint(TRACE_LEVEL_ERROR, "VirtIoStartIo RESET ALREADY IN PROGRESS !!!! Srb = 0x%p\n", Srb);
+        SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
+        CompleteRequestWithStatus(DeviceExtension, (PSRB_TYPE)Srb, SRB_STATUS_BUS_RESET);
+        return TRUE;
+    }
 
     RhelDbgPrint(TRACE_LEVEL_VERBOSE, " Srb = 0x%p\n", Srb);
 
@@ -972,15 +1052,18 @@ VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
         case SRB_FUNCTION_RESET_DEVICE:
         case SRB_FUNCTION_RESET_LOGICAL_UNIT:
             {
-                CompletePendingRequests(DeviceExtension);
-                CompleteRequestWithStatus(DeviceExtension, (PSRB_TYPE)Srb, SRB_STATUS_SUCCESS);
+                BOOLEAN status = VioStorResetBus(DeviceExtension);
+                CompleteRequestWithStatus(DeviceExtension,
+                                          (PSRB_TYPE)Srb,
+                                          status ? SRB_STATUS_SUCCESS : SRB_STATUS_ERROR);
 #ifdef DBG
                 RhelDbgPrint(TRACE_LEVEL_INFORMATION,
-                             " RESET (%p) Function %x Cnt %d InQueue %d\n",
+                             " RESET (%p) : Function : %x | Cnt %d | InQueue %d | Status: %s\n",
                              Srb,
                              SRB_FUNCTION(Srb),
                              adaptExt->srb_cnt,
-                             adaptExt->inqueue_cnt);
+                             adaptExt->inqueue_cnt,
+                             status ? "SRB_STATUS_SUCCESS" : "SRB_STATUS_ERROR");
                 for (USHORT i = 0; i < adaptExt->num_queues; i++)
                 {
                     if (adaptExt->vq[i])
@@ -1192,11 +1275,8 @@ BOOLEAN
 VirtIoResetBus(IN PVOID DeviceExtension, IN ULONG PathId)
 {
     UNREFERENCED_PARAMETER(PathId);
-    PADAPTER_EXTENSION adaptExt;
-    adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
 
-    CompletePendingRequests(DeviceExtension);
-    return TRUE;
+    return VioStorResetBus(DeviceExtension);
 }
 
 SCSI_ADAPTER_CONTROL_STATUS
@@ -2107,10 +2187,29 @@ VOID VioStorCompleteRequest(IN PVOID DeviceExtension, IN ULONG MessageID, IN BOO
 
     adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
 
+    VioStorVQLock(DeviceExtension, MessageID, &queueLock, bIsr);
+
     vq = adaptExt->vq[QueueNumber];
     element = &adaptExt->processing_srbs[QueueNumber];
 
-    VioStorVQLock(DeviceExtension, MessageID, &queueLock, bIsr);
+    /*
+     * Device can send ISR that will trigger DPC during SRB_FUNCTION_RESET_*.
+     * Driver can ignore this, because driver must complete all requests independently
+     * of ISR/DPC after reset finishes.
+     * At the same time DPC can be called before device will be initialized again
+     * after the reset.
+     * So, driver will safely ignore DPC during reset and manually complete all requests
+     * after full re-initialization.
+     */
+    if (adaptExt->reset_in_progress_count)
+    {
+        RhelDbgPrint(TRACE_LEVEL_ERROR, "VioStorCompleteRequest RESET ALREADY IN PROGRESS !!!!\n");
+        VioStorVQUnlock(DeviceExtension, MessageID, &queueLock, bIsr);
+        RhelDbgPrint(TRACE_LEVEL_VERBOSE, " <--- MessageID 0x%x\n", MessageID);
+
+        return;
+    }
+
     do
     {
         virtqueue_disable_cb(vq);
@@ -2223,6 +2322,7 @@ VOID VioStorCompleteRequest(IN PVOID DeviceExtension, IN ULONG MessageID, IN BOO
 VOID CompleteDpcRoutine(IN PSTOR_DPC Dpc, IN PVOID Context, IN PVOID SystemArgument1, IN PVOID SystemArgument2)
 {
     ULONG MessageID = PtrToUlong(SystemArgument1);
+
     VioStorCompleteRequest(Context, MessageID, FALSE);
 }
 
--- a/viostor/virtio_stor_hw_helper.c
+++ b/viostor/virtio_stor_hw_helper.c
@@ -113,13 +113,6 @@ RhelDoFlush(PVOID DeviceExtension, PSRB_TYPE Srb, BOOLEAN resend, BOOLEAN bIsr)
         MessageId = QueueNumber + 1;
     }
 
-    if (adaptExt->reset_in_progress)
-    {
-        SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
-        CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_BUS_RESET);
-        return TRUE;
-    }
-
     srbExt->MessageID = MessageId;
     vq = adaptExt->vq[QueueNumber];
 
@@ -139,6 +132,15 @@ RhelDoFlush(PVOID DeviceExtension, PSRB_TYPE Srb, BOOLEAN resend, BOOLEAN bIsr)
     {
         VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
 
+        if (adaptExt->reset_in_progress_count)
+        {
+            VioStorVQUnlock(DeviceExtension, MessageId, &LockHandle, FALSE);
+
+            SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
+            CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_BUS_RESET);
+            return TRUE;
+        }
+
         element = &adaptExt->processing_srbs[QueueNumber];
 
         ULONG_PTR id = element->next_id;
@@ -153,6 +155,13 @@ RhelDoFlush(PVOID DeviceExtension, PSRB_TYPE Srb, BOOLEAN resend, BOOLEAN bIsr)
     }
     else
     {
+        if (adaptExt->reset_in_progress_count)
+        {
+            SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
+            CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_BUS_RESET);
+            return TRUE;
+        }
+
         element = &adaptExt->processing_srbs[QueueNumber];
     }
 
@@ -210,18 +219,20 @@ RhelDoReadWrite(PVOID DeviceExtension, PSRB_TYPE Srb)
     QueueNumber = GetSrbQueueNumber(DeviceExtension, Srb);
     MessageId = QueueNumber + 1;
 
-    if (adaptExt->reset_in_progress)
-    {
-        SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
-        CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_BUS_RESET);
-        return TRUE;
-    }
-
     srbExt->MessageID = MessageId;
     vq = adaptExt->vq[QueueNumber];
     RhelDbgPrint(TRACE_LEVEL_VERBOSE, " QueueNumber 0x%x vq = %p\n", QueueNumber, vq);
 
     VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
+
+    if (adaptExt->reset_in_progress_count)
+    {
+        VioStorVQUnlock(DeviceExtension, MessageId, &LockHandle, FALSE);
+
+        SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
+        CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_BUS_RESET);
+        return TRUE;
+    }
 
     element = &adaptExt->processing_srbs[QueueNumber];
 
@@ -352,13 +363,6 @@ RhelDoUnMap(IN PVOID DeviceExtension, IN PSRB_TYPE Srb)
     QueueNumber = GetSrbQueueNumber(DeviceExtension, Srb);
     MessageId = QueueNumber + 1;
 
-    if (adaptExt->reset_in_progress)
-    {
-        SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
-        CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_BUS_RESET);
-        return TRUE;
-    }
-
     srbExt->MessageID = MessageId;
     vq = adaptExt->vq[QueueNumber];
     RhelDbgPrint(TRACE_LEVEL_INFORMATION,
@@ -368,6 +372,15 @@ RhelDoUnMap(IN PVOID DeviceExtension, IN PSRB_TYPE Srb)
                  srbExt->vbr.out_hdr.type);
 
     VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
+
+    if (adaptExt->reset_in_progress_count)
+    {
+        VioStorVQUnlock(DeviceExtension, MessageId, &LockHandle, FALSE);
+
+        SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
+        CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_BUS_RESET);
+        return TRUE;
+    }
 
     element = &adaptExt->processing_srbs[QueueNumber];
 
@@ -431,13 +444,6 @@ RhelGetSerialNumber(IN PVOID DeviceExtension, IN PSRB_TYPE Srb)
     QueueNumber = GetSrbQueueNumber(DeviceExtension, Srb);
     MessageId = QueueNumber + 1;
 
-    if (adaptExt->reset_in_progress)
-    {
-        SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
-        CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_BUS_RESET);
-        return TRUE;
-    }
-
     srbExt->MessageID = MessageId;
     vq = adaptExt->vq[QueueNumber];
 
@@ -456,6 +462,15 @@ RhelGetSerialNumber(IN PVOID DeviceExtension, IN PSRB_TYPE Srb)
     srbExt->sg[2].length = sizeof(srbExt->vbr.status);
 
     VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
+
+    if (adaptExt->reset_in_progress_count)
+    {
+        VioStorVQUnlock(DeviceExtension, MessageId, &LockHandle, FALSE);
+
+        SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
+        CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_BUS_RESET);
+        return TRUE;
+    }
 
     element = &adaptExt->processing_srbs[QueueNumber];
 
--- a/viostor/virtio_stor.h
+++ b/viostor/virtio_stor.h
@@ -255,7 +255,7 @@ typedef struct _ADAPTER_EXTENSION
     STOR_ADDR_BTL8 device_address;
     blk_discard_write_zeroes blk_discard[16];
     REQUEST_LIST processing_srbs[MAX_CPU];
-    BOOLEAN reset_in_progress;
+    ULONG reset_in_progress_count;
     ULONGLONG fw_ver;
 #ifdef DBG
     LONG srb_cnt;


viostor/virtio_stor.c
```diff
@@ -1201,11 +1249,10 @@ BOOLEAN
...
 {
     UNREFERENCED_PARAMETER(PathId);
-    PADAPTER_EXTENSION adaptExt;
-    adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
 
-    CompletePendingRequests(DeviceExtension);
-    return TRUE;
+    // https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nc-storport-hw_reset_bus
+    // No info about cleaning SRBs, so do nothing with them
+    return VioStorResetBus(DeviceExtension, FALSE);
```
XanClic: Resetting the device will discard all requests, so it makes sense to me to also always discard (cancel) all associated SRBs.

I suppose not doing that will lead to the SRBs hanging, and if Windows doesn’t like that, it will at some point still send an `SRB_FUNCTION_RESET_*` to cancel them. So not completing the SRBs here should not pose a catastrophic problem, but it still would make sense to me to complete them, given their underlying virtio-blk requests are gone.

viostor/virtio_stor.c
```diff
@@ -840,54 +840,98 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
     {
-        RhelDbgPrint(TRACE_LEVEL_ERROR, "RESET IN THE PROGRESS !!!!\n");
+        RhelDbgPrint(TRACE_LEVEL_ERROR, "VioStorResetBus RESET IN THE PROGRESS !!!!\n");
+        return FALSE;
     }
+
+    RhelDbgPrint(TRACE_LEVEL_WARNING, " VioStorResetBus\n");
+
+    adaptExt->reset_in_progress = TRUE;
+    StorPortPause(DeviceExtension, 60); // The same as SCSI Reset pause time
```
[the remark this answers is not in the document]
XanClic: The better question is, where does this value come from. I assume it’s just the default Windows I/O timeout? Makes sense to me, but can we maybe get the actual runtime timeout in the driver?

(Edit: FWIW, I’m fine with it as-is. Just wondering whether it could be a bit cleaner.)

viostor/virtio_stor.c
```diff
@@ -981,8 +1025,12 @@ VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
         case SRB_FUNCTION_RESET_DEVICE:
         case SRB_FUNCTION_RESET_LOGICAL_UNIT:
             {
-                CompletePendingRequests(DeviceExtension);
-                CompleteRequestWithStatus(DeviceExtension, (PSRB_TYPE)Srb, SRB_STATUS_SUCCESS);
+                // https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/srb/ns-srb-_storage_request_block
+                // The miniport driver should complete any active requests for the target controller.
+                BOOLEAN status = VioStorResetBus(DeviceExtension, TRUE);
```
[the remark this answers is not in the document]
XanClic: It is declared on line 941 and set on line 945.

viostor/virtio_stor.c
```diff
@@ -840,54 +840,98 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+    adaptExt->reset_in_progress = TRUE;
+    StorPortPause(DeviceExtension, 60); // The same as SCSI Reset pause time
+    RhelShutDown(DeviceExtension);
+
+    if (CompleteSrb)
+    {
+        CompletePendingRequestsOnReset(DeviceExtension);
+    }
+
+    if (!VirtIoHwReinitialize(DeviceExtension))
```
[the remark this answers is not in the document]
XanClic: Those functions won’t encounter unrecoverable errors.

viostor/virtio_stor.c
```diff
@@ -840,54 +840,98 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
-            {
-                element->srb_cnt = 0;
-            }
-            VioStorVQUnlock(DeviceExtension, MessageID, &LockHandle, FALSE);
         }
-        StorPortResume(DeviceExtension);
+        if (element->srb_cnt)
+        {
+            element->srb_cnt = 0;
+        }
```
kevmw: Preexisting, but this 'if' is unnecessary, we can just unconditionally set element->srb_cnt = 0.

viostor/virtio_stor.c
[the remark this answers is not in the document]
kevmw: I'd suggest rewording the comment as "same as virtio-scsi" rather than just "SCSI", which is a lot more generic.

I find the whole idea that the StorPort automatically resumes the device after a timeout a bit concerning, this means that we could get new requests before reset has completed. Maybe VirtIoStartIo() should check adaptExt->reset_in_progress and return a bus reset error if it's true?

viostor/virtio_stor.c
```diff
@@ -840,54 +840,98 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+    RhelShutDown(DeviceExtension);
+
+    if (CompleteSrb)
+    {
+        CompletePendingRequestsOnReset(DeviceExtension);
+    }
+
+    if (!VirtIoHwReinitialize(DeviceExtension))
+    {
+        RhelDbgPrint(TRACE_LEVEL_FATAL, " VioStorResetBus cannot reinitialize HW\n");
```
kevmw, line 935: Is it intentional that this code path leaves adaptExt->reset_in_progress == TRUE? If so, I think a comment would be useful.

I'm not completely sure myself what I should expect, as if we ever end up here, the device seems terminally broken.

viostor/virtio_stor.c
```diff
@@ -2231,6 +2278,22 @@ VOID VioStorCompleteRequest(IN PVOID DeviceExtension, IN ULONG MessageID, IN BOO
...
+     * Driver can ignore this, because driver must complite all requests independantly
+     * of ISR/DPC after reset finishes.
+     * At the same time DPC can be called before device will be initialized again
+     * after the reset.
+     * So, driver will safely ignore DPC during reset and manualy comple all requests
+     * after full re-initialization.
+     */
+    if (adaptExt->reset_in_progress)
+    {
+        RhelDbgPrint(TRACE_LEVEL_ERROR, "CompleteDpcRoutine RESET IN THE PROGRESS !!!!\n");
```
kevmw: Minor nit: Should be "RESET IN PROGRESS" without "THE" (also in the other place that has the same message)

viostor/virtio_stor.c
kostyanf14: I also don't know what to do in this case. Device broken, storport paused, virtio queues destroyed. Throw a BSOD with a custom error code?

viostor/virtio_stor.c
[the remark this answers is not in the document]
kostyanf14: >Maybe VirtIoStartIo() should check adaptExt->reset_in_progress and return a bus reset error if it's true?

Sounds interesting

kostyanf14:
@YanVugenfirer Please review

viostor/virtio_stor.c
```diff
@@ -840,54 +840,98 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
-    if (!adaptExt->reset_in_progress)
+    for (ULONG index = 0; index < adaptExt->num_queues; index++)
     {
-        adaptExt->reset_in_progress = TRUE;
-        StorPortPause(DeviceExtension, 10);
-
-        for (ULONG index = 0; index < adaptExt->num_queues; index++)
+        PREQUEST_LIST element = &adaptExt->processing_srbs[index];
+        STOR_LOCK_HANDLE LockHandle = {0};
+        ULONG MessageID = index + 1;
```
YanVugenfirer, line 858: Corner case: One MSI vector only (adaptExt->msix_one_vector) - I think the message ID should be 0.

YanVugenfirer: We don't see in the rest of the code back VioStorVQLock handles it, and MESSAGENUMBER_TO_QUEUE()     had this line
`        if (param.MessageNumber != 0)                                                                                  \`

viostor/virtio_stor.c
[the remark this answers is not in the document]
YanVugenfirer: StorPortPause(DeviceExtension, 60);  - 60 is our internal driver value. Maybe we should somehow make it related to the disk reset value? But in any case, I think those experiments can be done later.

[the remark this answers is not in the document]
YanVugenfirer: After looking at the [documentation](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nf-storport-storportpause) of StorPortPause, I don't think setting it to 60 seconds is a good idea. Because we know this is a disk service time out value for latest Widows, meaning that pausing our adapter for 60 seconds will trigger upper layer timeout. Do we want it?

viostor/virtio_stor.c
benyamin-codez: > Corner case: One MSI vector only (adaptExt->msix_one_vector) - I think the message ID should be 0.

It should be 0, but we see 1 still being used. I have fixes for this.

viostor/virtio_stor.c
```diff
@@ -840,54 +840,98 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
     return ret;
 }
 
-VOID CompletePendingRequests(IN PVOID DeviceExtension)
+VOID CompletePendingRequestsOnReset(IN PVOID DeviceExtension)
```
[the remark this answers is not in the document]
benyamin-codez: per `vioscsi`...

viostor/virtio_stor.c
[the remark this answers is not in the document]
benyamin-codez: > After looking at the [documentation](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nf-storport-storportpause) of StorPortPause, I don't think setting it to 60 seconds is a good idea. Because we know this is a disk service time out value for latest Widows, meaning that pausing our adapter for 60 seconds will trigger upper layer timeout. Do we want it?

Yan, this might be the intent, and could potentially fix the issues mentioned in #907. Your hint to use the a driver based value instead has some merit. Obtaining the `disk.sys` driver `IoTimeoutValue` and `TimeOutValue` value would break SDV. We can, however, lookup the `IoTimeoutValue` for `viostor`, but would need the capability per PR #1297 (which I will update shortly).

viostor/virtio_stor.c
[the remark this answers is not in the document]
benyamin-codez: The bot's a little blind, but I have been preferring:

`PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;`

However, in most instances, if the function uses `PADAPTER_EXTENSION` rather than `PVOID` for `DeviceExtension`, you could just use `DeviceExtension` instead...

viostor/virtio_stor.c
benyamin-codez: I used:

```c
RhelDbgPrint(TRACE_LEVEL_ERROR, "RESET ALREADY IN PROGRESS !!!!\n");
```

viostor/virtio_stor.c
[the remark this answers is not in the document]
kostyanf14: @benyamin-codez We are not sure about #1297. VioSCSI implementation of work with the registry looks simpler.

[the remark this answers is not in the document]
benyamin-codez: It adds per-HBA settings.
Having, e.g. common `max_segments` values, can break some multi-disk/HBA implementations.
This was the original purpose.
I have dropped it into it's own source and include files, so we don't have to scroll past... 8^d
Will force-push the update soon.

It would be great to get the builtin StorPort `Device(d)` type working...
...but AFAICT, this can only work after init is done, i.e. too late.

benyamin-codez:
I've got some PRs that will touch much of this one, esp. wrt locking, _LU control_, `QueueNumber<->MessageId`, MSI-X, etc.
It's possible the combination might fix the problem too, so I'll also drop my WIP as a draft PR.
I'll try to drop my related PRs in the next 12 hours or so, so you can at least consider them in part or as an alternative.

viostor/virtio_stor.c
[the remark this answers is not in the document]
kevmw: It seems clear to me that we can't take any new requests while reset is still in progress. So the choice that we have is letting StorPort queue them until the reset is done or to accept them and complete them with an error. I'm not sure what StorPort does with SRB_STATUS_BUS_RESET from a new request - but could it be any better than queuing the request, even if it potentially ran into a timeout eventually?

If I understand correctly, the disk service time out is configurable anyway, so I don't think any number lower than 60 is strictly better in this regard either. I'm actually tempted to suggest ULONG_MAX instead - or not pausing at all and relying only on adaptExt->reset_in_progress everywhere.

I also see that fo HwResetBus it doesn't matter because that one stops new requests from coming in anyway. So this is only about SRB_FUNCTION_RESET_*.

viostor/virtio_stor.c
kevmw: I'm not sure what the philosophy in Windows is, but I wouldn't trigger a kernel panic in Linux because this might just be one disk of many and the rest of the system might still do useful things when this disk doesn't work any more.

Setting a flag that makes the driver return an error on every future operation until another reset succeeds sounds like the best option to me.

kostyanf14:
@benyamin-codez, can you please try to revert https://github.com/virtio-win/kvm-guest-drivers-windows/commit/c09af90 , apply this PR and check if https://github.com/virtio-win/kvm-guest-drivers-windows/issues/1453 will be fixed too?

We think that the original problem was with the reset, and with proper implementation, we can remove SRB Id at all

benyamin-codez:
> @benyamin-codez, can you please try to revert [c09af90](https://github.com/virtio-win/kvm-guest-drivers-windows/commit/c09af90) , apply this PR and check if #1453 will be fixed too?
> 
> We think that the original problem was with the reset, and with proper implementation, we can remove SRB Id at all

Sure @kostyanf14 . Give me an hour so to check `viostor`.

@kevmw can you advise if the routine would be similar for `vioscsi`...? Any obvious differences in the strategy?

kostyanf14:
@benyamin-codez vioscsi is more complicated. As I mentioned, `The VirtIO BLK device does not have special reset events/routines like the VirtIO SCSI device (controlq command VIRTIO_SCSI_T_TMF_LOGICAL_UNIT_RESET)`.  VirtIO SCSI devices have a reset mechanism via the control queue, and the driver sends it. So, something goes wrong, and it's not clear where.

benyamin-codez:
@kostyanf14 

Yep, via `DeviceReset()` in `helper.c`.

I was asking if the following should still apply regardless:

> 1. Pauses the VirtIO BLK adapter (in guest, StorPortPause)
> 1. Reset VirtIO device (virtio_device_reset) 
> 1. Delete VirtIO queues (virtio_delete_queues)
> 1. Cleaup all device memory (virtio_device_shutdown)
> 1. Complete all pendings SRBs (if needed, in guest, CompletePendingRequestsOnReset)
> 1. Perform VirtIO device initialization (virtio_device_initialize, virtio_find_queues, etc)
> 1. Resume the VirtIO BLK adapter (in guest, StorPortResume)

Perhaps I can rework `DeviceReset()`...

cc: @kevmw

kostyanf14:
@benyamin-codez 

Yes, but we thought that DeviceReset() via VIRTIO_SCSI_T_TMF_LOGICAL_UNIT_RESET is enough and we don't need a full virtio_device_reset flow. 

So this is an open question: why does VIRTIO_SCSI_T_TMF_LOGICAL_UNIT_RESET still require SRB ID
This is under discussion with QEMU team

kevmw:
> @kevmw can you advise if the routine would be similar for vioscsi...? Any obvious differences in the strategy?

The biggest difference between virtio-blk and virtio-scsi with respect to reset is probably that virtio-blk is always a single disk whereas virtio-scsi is a SCSI controller that can serve many disks. So in virtio-scsi, you'll want to try to reset a single LUN first (this is VIRTIO_SCSI_T_TMF_LOGICAL_UNIT_RESET) instead of resetting the whole controller.

But apart from this, the fundamental principle regarding in-flight SRBs is the same: You first wait for the reset to complete, and then you know that the requests don't exist any more on the device side and you can complete them on the software side with SRB_STATUS_BUS_RESET.

benyamin-codez:
> @benyamin-codez, can you please try to revert [c09af90](https://github.com/virtio-win/kvm-guest-drivers-windows/commit/c09af90) , apply this PR and check if #1453 will be fixed too?
> 
> We think that the original problem was with the reset, and with proper implementation, we can remove SRB Id at all

@kostyanf14 

Ok, so I merged this PR onto a tree with c09af90 reverted.

So far I am unable to reproduce the problem with `fio` using the verify reproducer.
I did a 5 minute and a 20 minute run without issue. Previously I could produce errors within 1 minute.
Performance is reasonable. I did NOT perform a comparative performance test with any other PRs or `master.`

I did NOT do a SDV build (running now).

I did NOT force a reset or specifically test reset functionality.
Would you like me to? I can run up my GDB harness and hang some IOs with `blkdebug`...

I did include some very minor but relevant changes with unused parameters, constants, declarations and tracing plus a minor locking fix I found (grab the `element` under spinlock). I did NOT include any other functional fixes, e.g. proper MSI-X locking mechanisms, etc. I also did NOT include any other improvements in c09af90, e.g trace when `!bFound`.
The diff against your PR follows:

```diff
diff -ru4 ./a/viostor/virtio_stor.c ./b/viostor/virtio_stor.c
--- ./a/viostor/virtio_stor.c
+++ ./b/viostor/virtio_stor.c
@@ -258,9 +258,8 @@
     UNREFERENCED_PARAMETER(Again);
 
     adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
 
-    adaptExt->last_srb_id = 1;
     adaptExt->system_io_bus_number = ConfigInfo->SystemIoBusNumber;
     adaptExt->slot_number = ConfigInfo->SlotNumber;
     adaptExt->dump_mode = IsCrashDumpMode;
 
@@ -841,35 +840,37 @@
 }
 
 VOID CompletePendingRequestsOnReset(IN PVOID DeviceExtension)
 {
-    PADAPTER_EXTENSION adaptExt;
+    PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
 
-    adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
-#ifdef DBG
-    RhelDbgPrint(TRACE_LEVEL_INFORMATION,
-                 "CompletePendingRequestsOnReset %d %d\n",
+    RhelDbgPrint(TRACE_LEVEL_WARNING,
+                 " CompletePendingRequests - srb_cnt : %d | inqueue_cnt : %d\n",
                  adaptExt->srb_cnt,
                  adaptExt->inqueue_cnt);
-#endif
-    for (ULONG index = 0; index < adaptExt->num_queues; index++)
+
+    PREQUEST_LIST element = NULL;
+    STOR_LOCK_HANDLE LockHandle = {0};
+    ULONG MessageID = 0;
+
+    for (ULONG vq_req_idx = VIRTIO_BLK_REQUEST_QUEUE_0; vq_req_idx < adaptExt->num_queues; vq_req_idx++)
     {
-        PREQUEST_LIST element = &adaptExt->processing_srbs[index];
-        STOR_LOCK_HANDLE LockHandle = {0};
-        ULONG MessageID = index + 1;
+        MessageID = vq_req_idx + VIRTIO_BLK_MSIX_VQ_OFFSET;
         VioStorVQLock(DeviceExtension, MessageID, &LockHandle, FALSE);
+
+        element = &adaptExt->processing_srbs[vq_req_idx];
+
         while (!IsListEmpty(&element->srb_list))
         {
             PLIST_ENTRY entry = RemoveHeadList(&element->srb_list);
             if (entry)
             {
                 pblk_req req = CONTAINING_RECORD(entry, blk_req, list_entry);
-                PSCSI_REQUEST_BLOCK currSrb = (PSCSI_REQUEST_BLOCK)req->req;
-                PSRB_EXTENSION currSrbExt = SRB_EXTENSION(currSrb);
-                if (currSrb)
+                PSCSI_REQUEST_BLOCK Srb = (PSCSI_REQUEST_BLOCK)req->req;
+                if (Srb)
                 {
-                    SRB_SET_DATA_TRANSFER_LENGTH(currSrb, 0);
-                    CompleteRequestWithStatus(DeviceExtension, (PSRB_TYPE)currSrb, SRB_STATUS_BUS_RESET);
+                    SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
+                    CompleteRequestWithStatus(DeviceExtension, (PSRB_TYPE)Srb, SRB_STATUS_BUS_RESET);
                     element->srb_cnt--;
                 }
             }
         }
@@ -938,20 +939,11 @@
 VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
 {
     PCDB cdb = SRB_CDB(Srb);
     PADAPTER_EXTENSION adaptExt;
-    PSRB_EXTENSION srbExt;
     UCHAR ScsiStatus = SCSISTAT_GOOD;
 
     adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
-    srbExt = SRB_EXTENSION(Srb);
-    srbExt->id = adaptExt->last_srb_id;
-    adaptExt->last_srb_id++;
-    if (adaptExt->last_srb_id == 0)
-    {
-        adaptExt->last_srb_id++;
-    }
-
     SRB_SET_SCSI_STATUS(((PSRB_TYPE)Srb), ScsiStatus);
 
     RhelDbgPrint(TRACE_LEVEL_VERBOSE, " Srb = 0x%p\n", Srb);
 
@@ -1032,13 +1024,14 @@
                                           (PSRB_TYPE)Srb,
                                           status ? SRB_STATUS_SUCCESS : SRB_STATUS_ERROR);
 #ifdef DBG
                 RhelDbgPrint(TRACE_LEVEL_INFORMATION,
-                             " RESET (%p) Function %x Cnt %d InQueue %d\n",
+                             " RESET (%p) : Function : %x | Count : %d | InQueue :%d | Status: %s\n",
                              Srb,
                              SRB_FUNCTION(Srb),
                              adaptExt->srb_cnt,
-                             adaptExt->inqueue_cnt);
+                             adaptExt->inqueue_cnt,
+                             status ? "SRB_STATUS_SUCCESS" : "SRB_STATUS_ERROR");
                 for (USHORT i = 0; i < adaptExt->num_queues; i++)
                 {
                     if (adaptExt->vq[i])
                     {
@@ -2152,9 +2145,9 @@
     PADAPTER_EXTENSION adaptExt = NULL;
     ULONG QueueNumber = MessageID - 1;
     STOR_LOCK_HANDLE queueLock = {0};
     struct virtqueue *vq = NULL;
-    ULONG_PTR srbId = 0;
+    pblk_req vbr = NULL;
     PSRB_TYPE Srb = NULL;
     PSRB_EXTENSION srbExt = NULL;
     UCHAR srbStatus = SRB_STATUS_SUCCESS;
     PREQUEST_LIST element = NULL;
@@ -2169,41 +2162,36 @@
     VioStorVQLock(DeviceExtension, MessageID, &queueLock, bIsr);
     do
     {
         virtqueue_disable_cb(vq);
-        while ((srbId = (ULONG_PTR)virtqueue_get_buf(vq, &len)) != 0)
+        while ((vbr = (pblk_req)virtqueue_get_buf(vq, &len)) != NULL)
         {
             PLIST_ENTRY le = NULL;
             BOOLEAN bFound = FALSE;
+            Srb = (PSTORAGE_REQUEST_BLOCK)vbr->req;
 #ifdef DBG
             InterlockedDecrement((LONG volatile *)&adaptExt->inqueue_cnt);
 #endif
             for (le = element->srb_list.Flink; le != &element->srb_list && !bFound; le = le->Flink)
             {
                 pblk_req req = CONTAINING_RECORD(le, blk_req, list_entry);
 
-                Srb = (PSRB_TYPE)req->req;
-                srbExt = SRB_EXTENSION(Srb);
-                // Only SRBs with existing (i.e. non-NULL) extension
-                // are inserted into our queues, thus, we may help
-                // the Code Analysis and provide it with this information
-                // in order to avoid false-positive warnings.
-                _Analysis_assume_(srbExt != NULL);
-                if (srbExt->id == srbId)
+                if (Srb)
                 {
-                    RemoveEntryList(le);
-                    element->srb_cnt--;
-                    bFound = TRUE;
-                    break;
+                    srbExt = SRB_EXTENSION(Srb);
+                    PSTORAGE_REQUEST_BLOCK currSrb = (PSTORAGE_REQUEST_BLOCK)req->req;
+                    PSRB_EXTENSION currSrbExt = SRB_EXTENSION(currSrb);
+                    if (currSrbExt == srbExt && (PSRB_TYPE)currSrb == Srb)
+                    {
+                        RemoveEntryList(le);
+                        element->srb_cnt--;
+                        bFound = TRUE;
+                        break;
+                    }
                 }
             }
 
-            if (!bFound)
-            {
-                RhelDbgPrint(TRACE_LEVEL_WARNING, " No Srb to complete for ID 0x%p\n", (void *)srbId);
-            }
-
-            if (bFound && srbExt->vbr.out_hdr.type == VIRTIO_BLK_T_GET_ID)
+            if (bFound && vbr->out_hdr.type == VIRTIO_BLK_T_GET_ID)
             {
                 adaptExt->sn_ok = TRUE;
                 if (Srb)
                 {
@@ -2245,9 +2233,10 @@
                 continue;
             }
             if (bFound && Srb)
             {
-                srbStatus = DeviceToSrbStatus(srbExt->vbr.status);
+                srbExt = SRB_EXTENSION(Srb);
+                srbStatus = DeviceToSrbStatus(vbr->status);
                 RhelDbgPrint(TRACE_LEVEL_INFORMATION,
                              " srb %p, QueueNumber %lu, MessageId %lu.\n",
                              Srb,
                              QueueNumber,
diff -ru4 ./a/viostor/virtio_stor.h ./b/viostor/virtio_stor.h
--- ./a/viostor/virtio_stor.h
+++ ./b/viostor/virtio_stor.h
@@ -80,8 +80,10 @@
 #define IO_PORT_LENGTH                     0x40
 #define MAX_CPU                            256u
 #define MAX_DISCARD_SEGMENTS               256u
 
+#define VIRTIO_BLK_MSIX_VQ_OFFSET          1
+#define VIRTIO_BLK_REQUEST_QUEUE_0         0
 #define VIRTIO_BLK_QUEUE_LAST              MAX_CPU
 
 #define VIRTIO_BLK_MSIX_CONFIG_VECTOR      0
 #define MIN_DISCARD_SECTOR_ALIGNMENT       8
@@ -255,13 +257,10 @@
     blk_discard_write_zeroes blk_discard[16];
     REQUEST_LIST processing_srbs[MAX_CPU];
     BOOLEAN reset_in_progress;
     ULONGLONG fw_ver;
-    ULONG_PTR last_srb_id;
-#ifdef DBG
     LONG srb_cnt;
     LONG inqueue_cnt;
-#endif
 } ADAPTER_EXTENSION, *PADAPTER_EXTENSION;
 
 typedef struct _VRING_DESC_ALIAS
 {
@@ -277,9 +276,8 @@
     ULONG out;
     ULONG in;
     ULONG MessageID;
     BOOLEAN fua;
-    ULONG_PTR id;
     VIO_SG sg[VIRTIO_MAX_SG];
     VRING_DESC_ALIAS desc[VIRTIO_MAX_SG];
 } SRB_EXTENSION, *PSRB_EXTENSION;
 
diff -ru4 ./a/viostor/virtio_stor_hw_helper.c ./b/viostor/virtio_stor_hw_helper.c
--- ./a/viostor/virtio_stor_hw_helper.c
+++ ./b/viostor/virtio_stor_hw_helper.c
@@ -131,10 +131,9 @@
     if (!resend)
     {
         VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
     }
-    if (virtqueue_add_buf(vq, &srbExt->sg[0], srbExt->out, srbExt->in, (void *)srbExt->id, va, pa) ==
-        VQ_ADD_BUFFER_SUCCESS)
+    if (virtqueue_add_buf(vq, &srbExt->sg[0], srbExt->out, srbExt->in, &srbExt->vbr, va, pa) == VQ_ADD_BUFFER_SUCCESS)
     {
         notify = virtqueue_kick_prepare(vq);
         InsertTailList(&element->srb_list, &srbExt->vbr.list_entry);
         element->srb_cnt++;
@@ -217,10 +216,9 @@
     RhelDbgPrint(TRACE_LEVEL_VERBOSE, " QueueNumber 0x%x vq = %p\n", QueueNumber, vq);
 
     element = &adaptExt->processing_srbs[QueueNumber];
     VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
-    if (virtqueue_add_buf(vq, &srbExt->sg[0], srbExt->out, srbExt->in, (void *)srbExt->id, va, pa) ==
-        VQ_ADD_BUFFER_SUCCESS)
+    if (virtqueue_add_buf(vq, &srbExt->sg[0], srbExt->out, srbExt->in, &srbExt->vbr, va, pa) == VQ_ADD_BUFFER_SUCCESS)
     {
         notify = virtqueue_kick_prepare(vq);
         InsertTailList(&element->srb_list, &srbExt->vbr.list_entry);
         element->srb_cnt++;
@@ -370,10 +368,9 @@
                  srbExt->vbr.out_hdr.type);
 
     element = &adaptExt->processing_srbs[QueueNumber];
     VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
-    if (virtqueue_add_buf(vq, &srbExt->sg[0], srbExt->out, srbExt->in, (void *)srbExt->id, va, pa) ==
-        VQ_ADD_BUFFER_SUCCESS)
+    if (virtqueue_add_buf(vq, &srbExt->sg[0], srbExt->out, srbExt->in, &srbExt->vbr, va, pa) == VQ_ADD_BUFFER_SUCCESS)
     {
         notify = virtqueue_kick_prepare(vq);
         InsertTailList(&element->srb_list, &srbExt->vbr.list_entry);
         element->srb_cnt++;
@@ -464,10 +461,9 @@
     srbExt->sg[2].length = sizeof(srbExt->vbr.status);
 
     element = &adaptExt->processing_srbs[QueueNumber];
     VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
-    if (virtqueue_add_buf(vq, &srbExt->sg[0], srbExt->out, srbExt->in, (void *)srbExt->id, va, pa) ==
-        VQ_ADD_BUFFER_SUCCESS)
+    if (virtqueue_add_buf(vq, &srbExt->sg[0], srbExt->out, srbExt->in, &srbExt->vbr, va, pa) == VQ_ADD_BUFFER_SUCCESS)
     {
         notify = virtqueue_kick_prepare(vq);
         InsertTailList(&element->srb_list, &srbExt->vbr.list_entry);
         element->srb_cnt++;
```

benyamin-codez:
@kevmw @kostyanf14 

wrt to `vioscsi` I suspect we need to implement specific routines for `SRB_FUNCTION_RESET_BUS` / `SRB_FUNCTION_RESET_DEVICE` and not rely on the `SRB_FUNCTION_RESET_LOGICAL_UNIT` pathway ...

...or at least discern between these when calling:
 `CompletePendingRequestsOnReset()`  ==> `DeviceReset()` ==> `VIRTIO_SCSI_T_TMF_LOGICAL_UNIT_RESET`

I suspect if we implement something similar to this PR (with the full shutdown and restart) for `SRB_FUNCTION_RESET_BUS` / `SRB_FUNCTION_RESET_DEVICE`, then that might solve the problem for `vioscsi` too.

If it's not a duplication of effort, I can give that a try if you like. Let me know...

kostyanf14:
@benyamin-codez, I will work on the cleanup for this PR first, so you can try to reimplement bus reset for vioscsi.

benyamin-codez:
> I did NOT do a SDV build (running now).

@kostyanf14 FYI, the SDV on that patch ran ok for Win10 and win11 on x86 and amd64. I didn't check ARM.

viostor/virtio_stor.c
benyamin-codez: @kostyanf14 

Just to be clear - the doco referenced states:

_"The port driver pauses all device IO queues for the adapter and then calls the HwStorResetBus routine at IRQL DISPATCH_LEVEL after acquiring the StartIo spin lock. **A miniport driver is responsible for completing SRBs received by [HwStorStartIo](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nc-storport-hw_startio) for PathId during this routine and setting their status to SRB_STATUS_BUS_RESET if necessary.**" [Emphasis mine]_

So I'm not sure `BOOLEAN CompleteSrb` is a required parameter of `VirtIoResetBus()`, but perhaps instead we should use the `PathId` information StorPort supplies to limit the scope of `CompletePendingRequestsOnReset()`...

kostyanf14: @benyamin-codez As we have only one disk in case of viostor, can we just complete all SRBs received by HwStorStartIo with SRB_STATUS_BUS_RESET?

kostyanf14:
@benyamin-codez @kevmw @YanVugenfirer @XanClic 

I hope I addressed all the issues from the review. 

Please check the PR again. I have not tested the new version with HLK yet.

viostor/virtio_stor.c
benyamin-codez: > ...can we just complete all SRBs...

I think that is advisable @kostyanf14 . We can glean anything learnt from `vioscsi` experimentation later.

I implemented a HW_UNIT_CONTROL callback for `viostor`, which I did see get used by StorPort, at least in traces. So, as one would expect, it does appear StorPort will act on unit controls before calling device and bus functions. It is not yet determined whether there is any advantage for VirtIO block devices with this behaviour.

viostor/virtio_stor.c
benyamin-codez: Failure should result in execution of:
```c
virtio_add_status(&adaptExt->vdev, VIRTIO_CONFIG_S_FAILED);
```
...inside HW_INITIALIZE...

Should we build a retry loop?

viostor/virtio_stor.c
```diff
@@ -839,67 +840,135 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+                 adaptExt->inqueue_cnt);
 #endif
-    if (!adaptExt->reset_in_progress)
+    for (ULONG index = 0; index < adaptExt->num_queues; index++)
     {
-        adaptExt->reset_in_progress = TRUE;
-        StorPortPause(DeviceExtension, 10);
-
-        for (ULONG index = 0; index < adaptExt->num_queues; index++)
+        PREQUEST_LIST element = &adaptExt->processing_srbs[index];
```
benyamin-codez: Ok to declare here, but the definition should probably be under spinlock.

viostor/virtio_stor.c
```diff
@@ -839,67 +840,135 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+        STOR_LOCK_HANDLE LockHandle = {0};
+        ULONG MessageID = adaptExt->msix_one_vector ? 0 : index + 1;
```
benyamin-codez: Not sure if this will always work when `adaptExt->msix_one_vector == TRUE`...
I saw traces with `MessageId = 1` when debugging.
This is the same as everywhere else in master at present anyway.
We can address it in one of my PRs.

viostor/virtio_stor.c
```diff
@@ -839,67 +840,135 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
-            ULONG MessageID = index + 1;
-            VioStorVQLock(DeviceExtension, MessageID, &LockHandle, FALSE);
-            while (!IsListEmpty(&element->srb_list))
+            PLIST_ENTRY entry = RemoveHeadList(&element->srb_list);
+            if (entry)
             {
-                PLIST_ENTRY entry = RemoveHeadList(&element->srb_list);
-                if (entry)
+                pblk_req req = CONTAINING_RECORD(entry, blk_req, list_entry);
+                PSCSI_REQUEST_BLOCK currSrb = (PSCSI_REQUEST_BLOCK)req->req;
```
benyamin-codez: References to `currSrb` can just be `Srb`.
This part was lifted from elsewhere and retains the verbiage.

viostor/virtio_stor.c
```diff
@@ -839,67 +840,135 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+                PSRB_EXTENSION currSrbExt = SRB_EXTENSION(currSrb);
```
benyamin-codez: `currSrbExt` is unused and can be removed.
This part was lifted from elsewhere and retains the verbiage.

viostor/virtio_stor.c
benyamin-codez: Can we report `status` here too? Per:

```c
RhelDbgPrint(TRACE_LEVEL_INFORMATION,
             " RESET (%p) : Function : %x | Count : %d | InQueue :%d | Status: %s\n",
             Srb,
             SRB_FUNCTION(Srb),
             adaptExt->srb_cnt,
             adaptExt->inqueue_cnt,
             status ? "SRB_STATUS_SUCCESS" : "SRB_STATUS_ERROR");
```

viostor/virtio_stor.c
YanVugenfirer: I think you saw MessageId = 1 because when there is one queue, we still should have two MSIx interrupts in the default case. One for the queue and one configuration\notification interrupt. 

But, if we have only one MSIx interrupt - then it will be used for both. I think all this logic should be resolved in MESSAGE_TO_QUEUE otherwise it is easy for  someone that changes the code to miss those corner cases. But it can be done latter.

viostor/virtio_stor.c
YanVugenfirer: For later: I am wondering if in order to reduce the chance of failure here, we shouldn't reallocated the memory for the queues, but rather keep the old one and only reinitialize the device with reused queues.

viostor/virtio_stor.c
```diff
@@ -2231,6 +2290,22 @@ VOID VioStorCompleteRequest(IN PVOID DeviceExtension, IN ULONG MessageID, IN BOO
...
+    /*
+     * Device can send ISR that will trigger DPC during SRB_FUNCTION_RESET_*.
+     * Driver can ignore this, because driver must complite all requests independantly
+     * of ISR/DPC after reset finishes.
+     * At the same time DPC can be called before device will be initialized again
+     * after the reset.
+     * So, driver will safely ignore DPC during reset and manualy comple all requests
+     * after full re-initialization.
+     */
+    if (adaptExt->reset_in_progress)
```
YanVugenfirer: @kostyanf14 Why there is no lock by MessageID?

viostor/virtio_stor.c
benyamin-codez: Currently, we essentially have `adaptExt->msix_one_vector == FALSE` when :

```c
adaptExt->msix_vectors >= adaptExt->num_queues + 1
```

...which means that `adaptExt->msix_one_vector == TRUE` ONLY when `adaptExt->msix_vectors = 1`...

I agree it's best to fix it more comprehensively elsewhere.
Hope to have a working proposal PR for this out for consideration in the next couple of hours.
Just consolidating changes between my WIP and drafts... 8^d

viostor/virtio_stor.c
benyamin-codez: Correct me if I'm wrong, but this is just an early exit if a reset is already underway.
The lock occurs in `VioStorCompleteRequest()`.

YanVugenfirer: If the assumption that DPC might be scheduled when the reset is on progress  - then the following in    VioStorCompleteRequest before the lock is acquired might be uninitialized:
 vq = adaptExt->vq[QueueNumber]; 

Also, I don't think we take lock when initializing device.

viostor/virtio_stor.c
```diff
@@ -2223,6 +2290,22 @@ VOID VioStorCompleteRequest(IN PVOID DeviceExtension, IN ULONG MessageID, IN BOO
...
 {
     ULONG MessageID = PtrToUlong(SystemArgument1);
+    PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)Context;
+    /*
+     * Device can send ISR that will trigger DPC during SRB_FUNCTION_RESET_*.
+     * Driver can ignore this, because driver must complite all requests independantly
+     * of ISR/DPC after reset finishes.
+     * At the same time DPC can be called before device will be initialized again
+     * after the reset.
+     * So, driver will safely ignore DPC during reset and manualy comple all requests
```
benyamin-codez: Small nit: typo: comple -> complete

viostor/virtio_stor.c
benyamin-codez: Normally StorPort manages the locks during init, but if we are forcing (and managing) the re-init then we might need to do some locking too.

Thorough reset testing should reveal if this is an issue that needs to be addressed.

Given that we are returning early and not hitting `VioStorCompleteRequest()`, then this shouldn't be an issue, but there are other calls to `VioStorCompleteRequest()` that might need to be checked still, but I think the prudent ones are all covered. In my testing I had already moved most of the `if (adaptExt->reset_in_progress)...` checks to the head of each calling routine, so we may need to check this.

viostor/virtio_stor.c
XanClic: Summary of the below is: I think the `num_queues` assignment in `VirtIoHwInitialize()` needs to be changed, because I believe as-is, it’s possible that we get `msix_one_vector == TRUE` even though `num_queues > 1`, which seems bad.

And if that isn’t done, we must not set `MessageID` to 0 here.

---

I was wondering why you said `msix_one_vector` can only be `TRUE` if `msix_vectors == 1`.

So there is this context before the `msix_one_vector` assignment:
```c
if (adaptExt->num_queues > 1 && ((adaptExt->num_queues + 1U) > adaptExt->msix_vectors)) {
    adaptExt->num_queues = (USHORT)adaptExt->msix_vectors;
}

/* abbreviated and condition transformed to match the condition above */
adaptExt->msix_one_vector = (adaptExt->num_queues + 1U) > adaptExt->msix_vectors;
```

As far as I can tell, that means we need `num_queues + 1` MSI vectors[0], and if we don’t have that, the driver limits the number of queues it uses. Strangely, though, it sets `num_queues = msix_vectors` – but with that assignment, we still get `num_queues + 1 > msix_vectors`. Don’t we need to set `num_queues = MAX(msix_vectors, 2) - 1`? (Then we could also drop the `num_queues > 1` part of the condition.)

As it stands, it looks like we’ll always set `msix_one_vector` if the original `num_queues + 1` exceeds `msix_vectors`, e.g. also if `num_queues` was 8 and `msix_vectors` is 4.

Only if we do set `num_queues = MAX(msix_vectors, 2) - 1`, then `msix_one_vector` will only be `true` if `msix_vectors == 1` (which will also imply `num_queues == 1`). (Maybe the condition should then also be more obvious with `msix_one_vector = msix_vectors == 1`.)

---

Now, back to `MessageID` here, which is passed only to `VioStorVQLock()` and `VioStorVQUnlock()`.

Say you have 8 vCPUs, the device has 16 queues, but only 4 MSI vectors[1]. I think `VirtIoFindAdapter()` then limits `num_queues` to 8 already (number of CPUs) before we get to `VirtIoHwInitialize()`. Then we find `num_queues > 1 && num_queues + 1 > msix_vectors` (`8 > 1 && 8 + 1 > 4`), so we set `num_queues = msix_vectors`, i.e. 4.

We then find `msix_vectors >= num_queues + 1` to be false (i.e. `4 >= 4 + 1` is false), so `msix_one_vector` is set to `TRUE`.

With this configuration, `CompletePendingRequestsOnReset()` will set `MessageID` to 0, and `VioStorVQLock()` will find `!isr`, `msix_enabled`, and `num_queues > 1`. Then, maybe the `NT_ASSERT(MessageID > 0)` fails, or we get an underflow and thus out of bounds access in `adaptExt->dpc[MessageID - 1]`.

---

One way to directly address the out-of-bounds issue would be not to set `MessageID` to 0 in case of `msix_one_vector == TRUE` here. It’s ignored by `VioStorVQLock()` anyway if `num_queues == 1`, and if `num_queues > 1`, it probably actually still makes more sense to set it to `index + 1`.

But as far as I understand, we always want to use the `StorPortAcquireMSISpinLock()` path in `VioStorVQLock()` if `msix_one_vector == TRUE`, i.e. we expect that `msix_one_vector` is true only if `num_queues == 1`. To achieve this, I think the `num_queues` assignment in `VirtIoHwInitialize()` needs to be fixed.

---

[0] Not sure why `+ 1`, but qemu has `num_queues + 1` as the default for the number of MSI vectors, too, so I assume we have one vector for global device stuff, and then one per queue.

[1] I think you get that configuration in qemu via `-smp 8 -device virtio-blk,...,num-queues=16,vectors=4`.

benyamin-codez: Yeah, so the NT_ASSERT would fail and we get a range of out-of-bounds access errors.

The reason it still works some of the time is QEMU presents the extra VQ.

In practice, if we have > 2 vectors we will map a VQ to each vector, except for vector 0 (`MessageId` 0).

However, it does **_appear_** to work quite happily with just a single vector, per the mods in PR #1499. The tracing in that WIP makes it pretty obvious what's going on. Typically the number of VQs will be limited to number of vectors - 1, except when `msi_one_vector == TRUE`, when they will share one (vector 0) and in the case that `num_queues = msi_vectors` (because `num_queues` > `msi_vectors`) the first VQ will share vector 0 and then additional queues will have their own vector. This last condition could be modded to be:

```c
adaptExt->num_queues = adaptExt->msix_vectors + VIRTIO_BLK_MSIX_VQ_OFFSET;
```
...which would give a more consistent init and free up vector 0 for device change notifications via MSI-X.

I am presently splitting a series of commits relevant to this particular issue, so that PR might make it clearer. It could be the wrong approach or need more testing, but it presently works quite well AFAICT.

In any case, I added the necessary constants for the `msi_one_vector` scenario so changes can be made wholesale, e.g. we could just bump everything by 1.

YanVugenfirer: Looks like refactoring of the decision of how many queues are used is needed and the calculations of messageID and queueID should not be split between several functions.

I wrote about `VirtIoHwInitialize` in https://github.com/mstsirkin/kvm-guest-drivers-windows/commit/69483523c73bfb1a5b033cf554f9e5344a3464cc#commitcomment-174843043

> [0] Not sure why + 1, but qemu has num_queues + 1 as the default for the number of MSI vectors, too, so I assume we have one vector for global device stuff, and then one per queue.

There is configuration interrupt for virtio-blk. It is handled in `VirtIoMSInterruptRoutine`, and there is code to check that MessageID is 0.  So let's say for some reason we have only one vector, this case is not different from using legacy PCI interrupt. We should handle it. Although we can decide that this is invalid configuration. In such case let's do it in explicit manner.

viostor/virtio_stor.c
kevmw: The msix_one_vector code paths are messy at the moment, but they kind of work. It is effectively single queue, but can have num_queues > 1, and the vector that it actually uses is 0, but the MessageId used throughout the code is 1.

I believe it needs to be 1 here, too, because it's used like this in VioStorVQLock: `&adaptExt->dpc[MessageID - 1]`. You don't want a negative index there. (In fact, I see that there is even an `NT_ASSERT(MessageID > 0)` there.

I definitely intend to clean this up, but I'm waiting for @benyamin-codez's PRs to avoid duplication.

viostor/virtio_stor.c
benyamin-codez: > This last condition could be modded to be:
> ```c
> adaptExt->num_queues = adaptExt->msix_vectors + VIRTIO_BLK_MSIX_VQ_OFFSET;
> ```
> ...which would give a more consistent init and free up vector 0 for device change notifications via MSI-X.

It will need a little more than that. I'm having a quick look...

kevmw: Let's not go into cleaning up that mess in the context of this PR.

We just need to keep MessageId = index + 1 here for msix_one_vector, too, to keep it consistent with the expectations of the called code, i.e. VioStorVQLock() and VioStorVQUnlock(). Most of the loop won't do anything because effectively only the first queue is used in this mode (by virtue of failing StorPortInitializePerfOpts() because perfData.LastRedirectionMessageNumber is out of bounds), but trying to special case that here would only make it more messy.

benyamin-codez: Its - not + ... I need a break 8^d

```c
/* edge case when queues exceed MSI vectors */
    ((adaptExt->num_queues + VIRTIO_BLK_REQUEST_QUEUE_0 + VIRTIO_BLK_MSIX_VQ_OFFSET) > adaptExt->msix_vectors))
{
    adaptExt->num_queues = adaptExt->msix_vectors - VIRTIO_BLK_MSIX_VQ_OFFSET;
}
```

benyamin-codez: Nope that BSODs for `msi_one_vector`. I'll leave it be for now methinks.

viostor/virtio_stor.c
kostyanf14: @YanVugenfirer Why do we need a lock there? If DPC is called before we start the reset, the queue exists, and StorPort waits for the StartIO lock before send reset. 
If we are already in reset, we don't remove the flag until the re-init pass.

viostor/virtio_stor.c
kostyanf14: We know about msix_one_vector logic issues. Let's fix them one by one. In current, bad implementation of msix_one_vector, we can set MessageID = 0 if msix_one_vector = TRUE, and the code will work. 

Let's commit this and fix the msix_one_vector logic in the next PR in all places, or let's fix the msix_one_vector logic first and then rebase PR on top of fix.

XanClic: > In current, bad implementation of msix_one_vector, we can set MessageID = 0 if msix_one_vector = TRUE, and the code will work.

That’s not true, I put `if (MessageID == 0) { *(volatile int *)0x42 = 23; }` into the `num_queues > 1` path of `VioStorVQLock()` (i.e. to make the assertion a hard one that isn’t compiled out), then ran with `-smp 8 -device virtio-blk,...,vectors=4`, and got a BSOD. I don’t get that BSOD if I drop the `vectors=4` (i.e. with the implicit `vectors=9`).

viostor/virtio_stor.c
YanVugenfirer: The comment says: "So, driver will safely ignore DPC during reset and manualy comple all requests
     * after full re-initialization." 
If DPC is called during reset for some reason then we have race condition before adaptExt->reset_in_progress is set to true.

viostor/virtio_stor.c
benyamin-codez: I think I've fixed it, but let's handle it in another PR. I should raise one soon.

viostor/virtio_stor.c
```diff
@@ -840,54 +840,99 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
     return ret;
 }
 
-VOID CompletePendingRequests(IN PVOID DeviceExtension)
+VOID CompletePendingRequestsOnReset(IN PVOID DeviceExtension)
```
kevmw: This function can be static.

viostor/virtio_stor.c
XanClic: I agree that the `num_queues` setting fix can be done elsewhere, but without that fix, setting `MessageID` to 0 here is wrong. That’s what causes the problem in `VioStorVQLock()` (out of bounds access / `NT_ASSERT`, or BSOD if I force it).

Which is why I believe the `ULONG MessageID = adaptExt->msix_one_vector ? 0 : index + 1;` line should be reverted to how it was in the initial version of this PR (`ULONG MessageID = index + 1;`).

viostor/virtio_stor.c
kevmw: I think I agree that we need to take the lock before checking `adaptExt->reset_in_progress`, but of course that only helps if we also take the DpcLock of all queues in VioStorResetBus() before setting `adaptExt->reset_in_progress` to make sure that all DPCs have either completed or will do nothing when the reset delete the queues.

viostor/virtio_stor.c
benyamin-codez: FWIW, the fix in `VirtIoHwInitialize()` was:

```c
/* edge case when queues exceed MSI vectors */
    ((adaptExt->num_queues + VIRTIO_BLK_REQUEST_QUEUE_0 + VIRTIO_BLK_MSIX_VQ_OFFSET) > adaptExt->msix_vectors))
{
    if (adaptExt->msix_vectors == 1)
    {
        adaptExt->num_queues = 1;
    }
    else
    {
        adaptExt->num_queues = (adaptExt->msix_vectors - VIRTIO_BLK_REQUEST_QUEUE_0 - VIRTIO_BLK_MSIX_VQ_OFFSET);
    }
}
```
More is required, and I will drop the PR ASAP. In the meantime you can glean parts from PR #1499.

viostor/virtio_stor.c
```diff
@@ -972,15 +1051,18 @@ VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
...
 #ifdef DBG
                 RhelDbgPrint(TRACE_LEVEL_INFORMATION,
-                             " RESET (%p) Function %x Cnt %d InQueue %d\n",
+                             " RESET (%p) : Function : %x | Cnt %d | InQueue %d | Status: %s\n",
                              Srb,
                              SRB_FUNCTION(Srb),
                              adaptExt->srb_cnt,
-                             adaptExt->inqueue_cnt);
+                             adaptExt->inqueue_cnt,
+                             status ? SRB_STATUS_SUCCESS : SRB_STATUS_ERROR);
```
XanClic: I think these are integers, i.e. don’t match `%s`; I think this was supposed to be `status ? "SRB_STATUS_SUCCESS" : "SRB_STATUS_ERROR"`.

viostor/virtio_stor.c
```diff
@@ -2223,6 +2318,8 @@ VOID VioStorCompleteRequest(IN PVOID DeviceExtension, IN ULONG MessageID, IN BOO
 VOID CompleteDpcRoutine(IN PSTOR_DPC Dpc, IN PVOID Context, IN PVOID SystemArgument1, IN PVOID SystemArgument2)
 {
     ULONG MessageID = PtrToUlong(SystemArgument1);
+    PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)Context;
```
benyamin-codez: This is unused.

viostor/virtio_stor.c
```diff
@@ -2111,6 +2190,21 @@ VOID VioStorCompleteRequest(IN PVOID DeviceExtension, IN ULONG MessageID, IN BOO
...
+     * At the same time DPC can be called before device will be initialized again
+     * after the reset.
+     * So, driver will safely ignore DPC during reset and manualy complete all requests
+     * after full re-initialization.
+     */
+    if (adaptExt->reset_in_progress_count)
+    {
+        RhelDbgPrint(TRACE_LEVEL_ERROR, "VioStorCompleteRequest RESET ALREADY IN PROGRESS !!!!\n");
+        goto out;
+    }
```
benyamin-codez, line 2211: I would have though putting this at the top of the function (immediately after the `adaptExt` definition) outside of the lock would have been sufficient. I think that would be functionally the same as where it lived in your last commit.

We haven't got a single `goto` anywhere else in the storage drivers. It would be awful to start using them. If this is necessary to have this under lock - which I continue to doubt - it would be better to have the `do` in an else leaf.

I note this mechanism will effectively stop processing if a reset occurs when in crash-dump mode.

viostor/virtio_stor.c
benyamin-codez: They need to be wrapped in quotes.

viostor/virtio_stor.c
```diff
@@ -839,54 +840,117 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+     * are currently in progress in DPC. Each new DPC call will have no affect because
+     * VioStorCompleteRequest will exits for any non-zero reset_in_progress_count value
+     */
+    for (ULONG MessageId = 1; MessageId <= adaptExt->num_queues; MessageId++)
     {
-        RhelDbgPrint(TRACE_LEVEL_ERROR, "RESET IN THE PROGRESS !!!!\n");
+        STOR_LOCK_HANDLE LockHandle = {0};
+        VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
+        adaptExt->reset_in_progress_count++;
+        VioStorVQUnlock(DeviceExtension, MessageId, &LockHandle, FALSE);
```
benyamin-codez, line 927: I don't think this is achieving what you want. I think you need to:

1. Remove the `VioStorVQUnlock()` here
2. Remove the `VioStorVQLock()` in the complimentary for loop below

This probably satisfies the need for a lock around the "skip if in reset" in `VioStorCompleteRequest()`, which is probably better if reverted to `CompleteDpcRoutine()`.

viostor/virtio_stor.c
```diff
@@ -839,54 +840,117 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+        return FALSE;
+    }
+
+    for (ULONG MessageId = 1; MessageId <= adaptExt->num_queues; MessageId++)
+    {
+        STOR_LOCK_HANDLE LockHandle = {0};
+        VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
+        adaptExt->reset_in_progress_count--;
+        VioStorVQUnlock(DeviceExtension, MessageId, &LockHandle, FALSE);
+    }
```
benyamin-codez, line 951: Per notes above for complimentary for loop.

viostor/virtio_stor.c
benyamin-codez: Noting you will clash with spinlocks in `CompletePendingRequestOnReset()`...

viostor/virtio_stor.c
kevmw: Yes, it is necessary to have it under the lock. Otherwise you get a TOCTOU race condition where `adaptExt->reset_in_progress_count` can be increased between the check and taking the lock.

I don't think there is a reason to be afraid of a `goto out`, this is standard practice in C.

benyamin-codez:
> The `reset_in_progress_count++`/`--` loops in `VioStorResetBus()` look a bit strange, but I suppose they will do the job. (I think the originally supposed idea was to lock all queues before setting `reset_in_progress`, then unlocking all of them? But this should effectively do the same thing.)

@kostyanf14 @XanClic 

Unless i'm missing additional mechanics, these spinlocks aren't wrapping anything but the counter...

viostor/virtio_stor_hw_helper.c
```diff
@@ -113,7 +113,7 @@ RhelDoFlush(PVOID DeviceExtension, PSRB_TYPE Srb, BOOLEAN resend, BOOLEAN bIsr)
         MessageId = QueueNumber + 1;
     }
 
-    if (adaptExt->reset_in_progress)
+    if (adaptExt->reset_in_progress_count)
```
kevmw: Should these checks be made while we hold the lock? I think we might access a stale `vq` after taking the lock if a reset took place between the check and taking the lock.

viostor/virtio_stor.c
kevmw: As @XanClic already mentioned in her comment above, this is indeed not what I had in mind (I thought of keeping the lock across reset like you suggest here), but the mechanism should actually work fine. Taking the lock before incrementing the counter means that no `VioStorCompleteRequest()` for the same queue is still in progress, and afterwards, it's harmless because the counter is increased.

Keeping it like this also means that `CompletePendingRequestOnReset()` doesn't need changing.

viostor/virtio_stor_hw_helper.c
kostyanf14: No. In general, this check can be removed, but I don't want to refactor these functions in this PR. These functions are called only from StartIO, so we never will never reach this check during reset.

viostor/virtio_stor.c
```diff
@@ -2111,6 +2190,21 @@ VOID VioStorCompleteRequest(IN PVOID DeviceExtension, IN ULONG MessageID, IN BOO
     element = &adaptExt->processing_srbs[QueueNumber];
 
     VioStorVQLock(DeviceExtension, MessageID, &queueLock, bIsr);
```
kevmw: I don't think it's a race that you'll ever lose in practice, but in theory if the reset fully completes while we're trying to take the lock here, and therefore the `reset_in_progress_count` check doesn't error out any more below, `vq` and maybe also `element` may be stale. It's probably better to assign them only when we have the lock.

viostor/virtio_stor.c
kostyanf14: @XanClic 

> (I think the originally supposed idea was to lock all queues before setting reset_in_progress, then unlocking all of them? But this should effectively do the same thing.)

Getting several locks for all queues is a more complicated task. I reuse the idea from NetKVM.

viostor/virtio_stor.c
benyamin-codez: I generally avoid jumping if I can - it's very bad on the knees... 8^D

Making the code into spaghetti is almost universally frowned upon in most projects I've been involved in. iinm, it's definitely not best practice - for at least two decades.

Did this pass SDV @kostyanf14 ..?

Another alternative to using `goto` would be to just call `VioStorUnlock()` and return early.

wrt to TOC-TOU race:

Ok, I was thinking from the perspective the counter had already incremented from 0, but I suppose in the instance the counter is about to increment (from 0) when the check occurs this might be a concern. Will we also need locks to wrap the same in `RhelGetSerialNumber()` and `RhelDo*()` too...?

viostor/virtio_stor.c
benyamin-codez: Kevin, I'm clashing with this one and PR #1495 a bit, but I'll try to drop relevant PRs in the next few hours (even if it's at risk of extensive rebasing)...

viostor/virtio_stor.c
benyamin-codez: Right. I will make a note for my PR(s) to allow for `msix_one_vector` in the loop.

kostyanf14: >I don't think this is achieving what you want. I think you need to:
    Remove the VioStorVQUnlock() here
    Remove the VioStorVQLock() in the complimentary for loop below

Totally NO. In this case, you need to save all STOR_LOCK_HANDLE, and as this is a SpinLoc,k you need to release it in reverse order. The same idea is used in NetKVM. You can work only when  reset_in_progress_count = 0; if reset_in_progress_count > 0, the reset is initiated and starts preparing the driver for this.

viostor/virtio_stor_hw_helper.c
benyamin-codez: I see you are already onto this...

We get a fair way down the path if we leave it to the last post to check for the reset. It might be worthwhile to consider if this is the best place to do the check. There might be a better alternative between the ISR and ~here~ `VioStorCompleteRequest()`.

viostor/virtio_stor.c
benyamin-codez: I've been putting `element` under lock. The `vq` has been safe to date... Maybe not after this work...

viostor/virtio_stor.c
benyamin-codez: Yeah, I'm caught up. Apologies for the cross talk...!

viostor/virtio_stor.c
benyamin-codez: > For later: I am wondering if in order to reduce the chance of failure here, we shouldn't reallocated the memory for the queues, but rather keep the old one and only reinitialize the device with reused queues.

wrt to `vioscsi`: I'll make a note to consider splitting this idea depending on BUS, DEVICE or LU.

viostor/virtio_stor.c
```diff
@@ -840,54 +840,98 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+{
+    PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
+    if (adaptExt->reset_in_progress)
     {
-        RhelDbgPrint(TRACE_LEVEL_ERROR, "RESET IN THE PROGRESS !!!!\n");
+        RhelDbgPrint(TRACE_LEVEL_ERROR, "VioStorResetBus RESET IN THE PROGRESS !!!!\n");
+        return FALSE;
     }
+
+    RhelDbgPrint(TRACE_LEVEL_WARNING, " VioStorResetBus\n");
```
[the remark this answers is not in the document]
benyamin-codez: Maybe consider amending the string " initiated..."

viostor/virtio_stor_hw_helper.c
benyamin-codez: If you do consider checking earlier than `VioStorCompleteRequest()` - apologies if my meaning wasn't clear - aside from `VirtIoInterrupt()` and `VirtIoMSInterruptRoutine()`, the other caller is `CompleteDpcRoutine()`... We don't need to consider `RhelDoReadWrite()` completing early because we already check for reset beforehand.

benyamin-codez: For the inerrupts, the check could be done in `CompleteDPC()`...

benyamin-codez: > These functions are called only from StartIO, so we never will never reach this check during reset.

Kostiantyn, whilst this is probably true, what if `VirtIoStartIo()` is called contemporaneously with a reset. If `VirtIoStartIo()` is called, and then immediately the reset, I think there would be a race, yes?

> We don't need to consider `RhelDoReadWrite()` completing early because we already check for reset beforehand.

If the check was earlier than `VioStorCompleteRequest()`, we would likely need to keep the check in `RhelDoReadWrite()`.

benyamin-codez:
Looking really good @kostyanf14 ..!

Just my pet dislike of that `goto`...
Any chance you could drop it in favour of a `VioStorVQUnlock()` + `return` and put me out of my misery..?

kostyanf14:
> Just my pet dislike of that `goto`...
> Any chance you could drop it in favour of a `VioStorVQUnlock()` + `return` and put me out of my misery..?

Looks like viostor/vioscsi are only 2 drivers without goto haha
Anyway, let's wait for the other review and test results

viostor/virtio_stor.c
kevmw: Reset reallocates the vqs after this PR.

viostor/virtio_stor_hw_helper.c
kevmw: > For the inerrupts, the check could be done in CompleteDPC()...

Only if CompleteDPC() were already taking the lock, otherwise you'd get a race.

viostor/virtio_stor.c
kevmw: Nobody is suggesting turning code into spaghetti. These are clean forward jumps within the same block to the cleanup sequence. They are essentially what you do for exceptions in C, except more explicit and therefore easier to keep track of than the try/catch blocks other languages have. (Though yes, duplicating one line as an alternative might be tolerable.)

Yes, we do need to have it under the lock in these functions, too, and the meanwhile updated version does it like this.

viostor/virtio_stor.c
```diff
@@ -989,7 +1071,7 @@ VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
                     }
                 }
 #endif
-                return TRUE;
+                return status;
```
kevmw: For the record, as discussed in chat, this should remain TRUE. Even in case of errors, we accept the request, but complete it with an error status code (a request that was never initiated couldn't be completed). This matches the pattern used for other request types.

viostor/virtio_stor.c
benyamin-codez: Noting we cannot call `StorPortEnablePassiveInitialization()` again...
...as it must only be called from the `HW_INITIALIZE` callback `VirtIoHwInitialize()`...

For reference here is the relevant part I used in PR #1300:

https://github.com/virtio-win/kvm-guest-drivers-windows/blob/d5c5a718839fca5a118eeb0d5acb3d81d6c0df25/viostor/virtio_stor.c#L921-L950

viostor/virtio_stor.c
benyamin-codez: > Nobody is suggesting turning code into spaghetti.

No perhaps not today. Apologies for any misunderstanding - it wasn't my intended meaning.

Spaghetti could potentially be introduced by future contributors and maintainers if someone misses it.
If there isn't anyone around to say "no to `goto`", in my experience, they will inevitably multiply like rabbits... Per CVE-2014-0160 (Heartbleed), CVE-2014-1266 (Apple goto fail), etc.

The arguments for and against are well established and don't need to be aired further here, but I will say that one person's ease of tracking is not necessarily the case for everyone. Due to past experience, some of us get a knot in the stomach at the sight of a `goto` - quickly reaching for a find-all on the label, both with and without the statement...

In this case, I think it's perfectly fine to tolerate two or three lines:
```c
VioStorVQUnlock(...);
RhelDbgPrint(...); // or EXIT_FN(); or the previous trace  with "Returning..." appended
return;
```
Mischief managed...!
...but to each his own.

viostor/virtio_stor.c
benyamin-codez: This is the RESET SRB, yes?
Did anyone consider setting SRB_STATUS_BUSY rather than SRB_STATUS_ERROR and returning TRUE...?

viostor/virtio_stor.c
```diff
@@ -839,54 +840,117 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+ * memory will be still in use by the device and free in Windows. As a result,
+ * we got use-after-free when the device completes the request and sends it back to Windows.
+ * This may cause a memory corruption, as Windows may already have reused the memory.
+ */
+BOOLEAN VioStorResetBus(IN PVOID DeviceExtension)
+{
+    PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
+    if (adaptExt->reset_in_progress_count)
+    {
+        RhelDbgPrint(TRACE_LEVEL_ERROR, "VioStorResetBus RESET ALREADY IN PROGRESS !!!!\n");
```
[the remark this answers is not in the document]
benyamin-codez: 8^( - It's very fussy isn't it...

benyamin-codez:
@kostyanf14 - I was too slow... Did the HCK runs fair well..? I was looking forward to seeing all greens..!

kostyanf14:
@benyamin-codez 

Win11 - all pass
Win2025 - all pass

Win2022 - was not fully finished, but Disk Stress passed
Win10 - HCK-CI crashed, need to extend wait time, manual test of Disk Stress pass

kostyanf14:
@benyamin-codez The previous results and binary on Dropbox https://people.redhat.com/~kkostiuk/auto_hck/virtio-win/kvm-guest-drivers-windows/viostor-Win2025x64_host/2026_01_26_19_46_02/results.html. I am asking you to test this version in any tests that you want to give some feedback about stability.

kostyanf14:
@iops-hunter @JonKohler @sb-ntnx I would ask you to test this PR as well.

benyamin-codez:
> The [previous results](https://people.redhat.com/~kkostiuk/auto_hck/virtio-win/kvm-guest-drivers-windows/viostor-Win2025x64_host/2026_01_26_19_46_02/results.html) and binary on Dropbox. I am asking you to test this version in any tests that you want to give some feedback about stability.

Nice.
I'll run up a performance test, comparing against master.

I'd post an invite on the slack too maybe.
Also @wangyan0507 might be interested wrt what effect this PR has on issue #907.

benyamin-codez:
@kostyanf14 - Here are some pics of IOPS count variance from the performance test:

win10_random_4k_o32_t16_r_iops_variance_latency

win10_random_64k_o32_t16_r_iops_variance_latency

win10_sequential_1m_o8_t1_s_iops_variance_latency

win10_sequential_1m_o32_t8_T4k_s8192k_iops_variance_latency
...

As you can see they are pretty close. Win11 had similar outcomes, i.e. the plots were similar between master and this PR. There was one minor oddity with the 8MiB-stride sequential read where both master and this PR had a dip:

win11_sequential_1m_o32_t8_T4k_s8192k_iops_variance_latency
...

I tested twice in reverse order, so it was reproducible at least to that extent.
More testing would be needed to see if it is a phantom unrelated to the driver.
I note the dip is also seen in master for the 4MiB-stride sequential read too.

Anyway, if there is an issue, it is likely not in this PR...

viostor/virtio_stor.c
kevmw: I think it's ok because `adaptExt->dpc_ok` is still TRUE, so we won't try to call `StorPortEnablePassiveInitialization()` again. We also keep the old `adaptExt->dpc`.

viostor/virtio_stor.c
kevmw: Wouldn't that cause the reset to be retried and fail again later? (Because `adaptExt->reset_in_progress_count` is still > 0) Why would you want that?

benyamin-codez:
... and as expected, no `fio` verify errors.

benyamin-codez:
Kostiantyn, it would be prudent to extensively test reset operations in a variety of scenarios.

Perhaps Kevin, or someone in his team, has a purpose-built `blkdebug` harness...?

No point re-inventing the wheel if there's something already available. Mine is defintely not prime-time.

viostor/virtio_stor.c
benyamin-codez: Right - We could do a retry loop and OR the existing conditional with a retry counter.
In qualifying scenarios, what are the chances another go around would succeed...?
Is the strategy to leave it and get hit with new reset requests _ad infinitum_..?

benyamin-codez:
@kostyanf14 - Now I'm really itching to fix the Flush Test cleanup failures...! HCK's looking much better now, hey..!

kostyanf14:
@benyamin-codez If you want, I can export the HLK Flush test as a standalone app, and feel free to find the  missing DLL

benyamin-codez:
> If you want, I can export the HLK Flush test as a standalone app, and feel free to find the missing DLL

Yeah, ok. Let's give it a go. It's probably a good start.
I do have the HLK, but it might pay for me to try with what you have.

To reproduce the fault I might need the environment details too.
iirc, this was happening in a containerised Server Core install. Is that right?

kostyanf14:
Server Core
Win10 Pro

PR #1494, opened 2026-01-20
Add VS Code tasks for building virtio-win drivers CI-NO-BUILD

elizashurov:
Adds tasks.json file that provides VS Code task configurations for building the drivers.

Tasks include:
- Build all drivers for Win10/Win11 (all architectures or specific) 
- Individual driver builds with interactive OS and architecture selection 
- Clean and sign operations

### DIFF
--- a/vioserial/buildAll.bat
+++ b/vioserial/buildAll.bat
@@ -1,6 +1,6 @@
 @echo off
 if "%VIRTIO_WIN_NO_ARM%"=="" call ..\build\build.bat vioser.sln "Win10 Win11" ARM64
 if errorlevel 1 goto :eof
-call ..\build\build.bat vioser.sln "Win11" %*
+call ..\build\build.bat vioser.sln "Win10 Win11" %*
 if errorlevel 1 goto :eof
-call ..\build\build.bat sys\vioser.vcxproj "Win11_SDV" %*
+call ..\build\build.bat sys\vioser.vcxproj "Win10_SDV Win11_SDV" %*


.vscode/tasks.json
```diff
@@ -0,0 +1,386 @@
+{
+  // For build prerequisites and setup instructions, see:
```
kostyanf14: Based on the spec, JSON does not support comments. 

To prevent issues when the VSCode parser will be updated, use the the following structure: `_comment": "comment text goes here..."`. "_comment" is not a mandatory style, but some unused key "based_on", "details", etc.

.vscode/tasks.json
```diff
@@ -0,0 +1,386 @@
...
+        "panel": "shared",
+        "showReuseMessage": true,
+        "clear": false
+      }
+    },
+    {
+      "label": "Build: fwcfg64",
+      "type": "shell",
+      "command": "cmd",
+      "args": ["/c", "cd fwcfg64 && build_AllNoSdv.bat ${input:windowsVersion} ${input:architecture}"],
```
kostyanf14: about "cd fwcfg64". Previously, I saw that you can specify a workdir to prevent "cd". Is this valid now?

.vscode/tasks.json
elizashurov: changed to _comment

.vscode/tasks.json
elizashurov: I changed to use options.cwd instead of cd as Gemini suggested. it works great

YanVugenfirer:
I just started to play with it. Looks nice (screen shot with deliberate compilation error)!

I think some wiki is needed for first timers:
1. How to execute VS tasks
2. Clicking on error to get to exact line (holding Ctrl)

Also terminal window has limited size in lines. Definitely not enough for our projects compilation. Check how to increase it VS config.

Screenshot 2026-02-03 at 22 49 44

YanVugenfirer:
VirtIO library build task is missing. Sometimes we want to build only library.

YanVugenfirer:
Another think - I see in some cases (pvpanic and virtio-serial for example, maybe others) when I choose specific configuration (x64), still all configurations for the OS are built. We have the ability to limit the build for specific configuration only.

elizashurov:
@YanVugenfirer I updated based on the comments.
Fixed ARM64 auto-building (only selected arch builds now), added TASKS_GUIDE.md readme and increased terminal scrollback to 10,000 lines.

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+  // Increase terminal scrollback buffer to handle large build outputs
```
kostyanf14: The same comment about comment

vioserial/buildAll.bat
```diff
@@ -1,6 +1,6 @@
 @echo off
 if "%VIRTIO_WIN_NO_ARM%"=="" call ..\build\build.bat vioser.sln "Win10 Win11" ARM64
 if errorlevel 1 goto :eof
-call ..\build\build.bat vioser.sln "Win11" %*
+call ..\build\build.bat vioser.sln "Win10 Win11" %*
 if errorlevel 1 goto :eof
-call ..\build\build.bat sys\vioser.vcxproj "Win11_SDV" %*
+call ..\build\build.bat sys\vioser.vcxproj "Win10_SDV Win11_SDV" %*
```
kostyanf14, line 6: Please fix as a separate commit

.vscode/settings.json
elizashurov: done

vioserial/buildAll.bat
elizashurov: done

PR #1299, opened 2025-02-17
[viostor] Align memory allocations

benyamin-codez:
Refactors `VirtIoFindAdapter()` to:

1. Pad the uncached extension to ensure it is page-aligned
2. Include memory layout art
3. Improves relevant instrumentation

PR #1495, opened 2026-01-20
RHEL-149878: viostor: Fix MessageId update in MESSAGENUMBER_TO_QUEUE() and related cleanup

kevmw:
This is a series of three commits:

The first commit fixes the MessageId calculation in MESSAGENUMBER_TO_QUEUE() for the case where StorPortGetStartIoPerfParams() returns an optimal MessageNumber that is higher than the MessageNumber of the existing queues.

The other two commits are cleanups that deduplicate and simplify the code that is using MESSAGENUMBER_TO_QUEUE() today.

### DIFF
--- a/viostor/virtio_stor_hw_helper.c
+++ b/viostor/virtio_stor_hw_helper.c
@@ -44,48 +44,40 @@
         pa = va ? StorPortGetPhysicalAddress(DeviceExtension, NULL, va, &len).QuadPart : 0;                            \
     }
 
-#define MESSAGENUMBER_TO_QUEUE()                                                                                       \
-    {                                                                                                                  \
-        if (param.MessageNumber != 0)                                                                                  \
-        {                                                                                                              \
-            MessageId = param.MessageNumber;                                                                           \
-            QueueNumber = MessageId - 1;                                                                               \
-            if (QueueNumber >= adaptExt->num_queues)                                                                   \
-            {                                                                                                          \
-                QueueNumber %= adaptExt->num_queues;                                                                   \
-                MessageId = QueueNumber + 1;                                                                           \
-            }                                                                                                          \
-        }                                                                                                              \
-    }
-
 static ULONG GetSrbQueueNumber(IN PVOID DeviceExtension, IN PSRB_TYPE Srb)
 {
     PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
-    ULONG QueueNumber = 0;
-    ULONG MessageId = 1;
+    STARTIO_PERFORMANCE_PARAMETERS param;
+    ULONG status;
+    ULONG QueueNumber;
 
-    if (adaptExt->num_queues > 1)
+    if (adaptExt->num_queues <= 1)
     {
-        STARTIO_PERFORMANCE_PARAMETERS param;
-        ULONG status = STOR_STATUS_SUCCESS;
-
-        param.Size = sizeof(STARTIO_PERFORMANCE_PARAMETERS);
-        status = StorPortGetStartIoPerfParams(DeviceExtension, (PSCSI_REQUEST_BLOCK)Srb, &param);
-        if (status == STOR_STATUS_SUCCESS)
-        {
-            RhelDbgPrint(TRACE_LEVEL_INFORMATION,
-                         " srb %p, QueueNumber %lu, MessageNumber %lu, ChannelNumber %lu.\n",
-                         Srb,
-                         QueueNumber,
-                         param.MessageNumber,
-                         param.ChannelNumber);
-            MESSAGENUMBER_TO_QUEUE();
-        }
-        else
-        {
-            RhelDbgPrint(TRACE_LEVEL_ERROR, " StorPortGetStartIoPerfParams failed. srb %p status 0x%x.\n", Srb, status);
-        }
+        return 0;
     }
+
+    param.Size = sizeof(STARTIO_PERFORMANCE_PARAMETERS);
+    status = StorPortGetStartIoPerfParams(DeviceExtension, (PSCSI_REQUEST_BLOCK)Srb, &param);
+    if (status != STOR_STATUS_SUCCESS)
+    {
+        RhelDbgPrint(TRACE_LEVEL_ERROR, " StorPortGetStartIoPerfParams failed. srb %p status 0x%x.\n", Srb, status);
+        return 0;
+    }
+
+    if (param.MessageNumber == 0)
+    {
+        QueueNumber = 0;
+    }
+    else
+    {
+        QueueNumber = (param.MessageNumber - 1) % adaptExt->num_queues;
+    }
+    RhelDbgPrint(TRACE_LEVEL_INFORMATION,
+                 " srb %p, MessageNumber %lu, ChannelNumber %lu -> QueueNumber %lu\n",
+                 Srb,
+                 param.MessageNumber,
+                 param.ChannelNumber,
+                 QueueNumber);
 
     return QueueNumber;
 }


YanVugenfirer:

kostyanf14:
ok to test

viostor/virtio_stor_hw_helper.c
```diff
@@ -58,6 +58,38 @@
         }                                                                                                              \
     }
 
+static ULONG GetSrbQueueNumber(IN PVOID DeviceExtension, IN PSRB_TYPE Srb)
+{
+    PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
+    ULONG QueueNumber = 0;
+    ULONG MessageId = 1;
```
YanVugenfirer: What about the corner case when there is only one vector for the queue and the notification interrupt

viostor/virtio_stor_hw_helper.c
```diff
@@ -58,6 +58,38 @@
         }                                                                                                              \
     }
 
+static ULONG GetSrbQueueNumber(IN PVOID DeviceExtension, IN PSRB_TYPE Srb)
```
YanVugenfirer, line 47: Just a suggestion, not to recalculate MessageId in the caller, why not make it more "windows" style:
`static ULONG GetSrbQueueAndMessageNumber(IN PVOID DeviceExtension, IN PSRB_TYPE Srb, OUT PULONG QueueNumber, OUT PULONG MessageId);`

The return value is status.

This way we can return the status of the `StorPortGetStartIoPerfParams` (although it is not used) and not to recalculate MessageId again in the caller.

viostor/virtio_stor_hw_helper.c
```diff
@@ -419,26 +396,9 @@ RhelGetSerialNumber(IN PVOID DeviceExtension, IN PSRB_TYPE Srb)
...
-            MESSAGENUMBER_TO_QUEUE();
-        }
-        else
-        {
-            RhelDbgPrint(TRACE_LEVEL_ERROR, " StorPortGetStartIoPerfParams failed srb %p status 0x%x.\n", Srb, status);
-        }
-    }
+    QueueNumber = GetSrbQueueNumber(DeviceExtension, Srb);
+    MessageId = QueueNumber + 1;
+
```
YanVugenfirer, line 393: And then the call will be

status = GetSrbQueueAndMessageNumber(DeviceExtension, Srb, &QueueNumber, &MessageId);

viostor/virtio_stor_hw_helper.c
YanVugenfirer: There is later convoluted logic in VioStorVQLock to account for this case

VOID VioStorVQLock(IN PVOID DeviceExtension, IN ULONG MessageID, IN OUT PSTOR_LOCK_HANDLE LockHandle, IN BOOLEAN isr)
{
    PADAPTER_EXTENSION adaptExt;
    RhelDbgPrint(TRACE_LEVEL_VERBOSE, " ---> MessageID = %d isr = %d\n", MessageID, isr);

    if (!isr)
    {
        if (adaptExt->msix_enabled)
        {
            if (adaptExt->num_queues > 1)
            {

                NT_ASSERT(MessageID > 0);
                NT_ASSERT(MessageID <= adaptExt->num_queues);
                StorPortAcquireSpinLock(DeviceExtension, DpcLock, &adaptExt->dpc[MessageID - 1], LockHandle);
            }
            else
            {
                ULONG oldIrql = 0;
                StorPortAcquireMSISpinLock(DeviceExtension, **(adaptExt->msix_one_vector ? 0 : MessageID),** &oldIrql);
                LockHandle->Context.OldIrql = (KIRQL)oldIrql;
            }
        }
        else
        {
            StorPortAcquireSpinLock(DeviceExtension, InterruptLock, NULL, LockHandle);
        }
    }
    RhelDbgPrint(TRACE_LEVEL_VERBOSE, " <--- MessageID = %d\n", MessageID);
}

YanVugenfirer: adaptExt->msix_one_vector ? 0 : MessageID)

kevmw: I'm not entirely sure how this mode really works. We still store some non-zero number in srbExt->MessageID and that doesn't feel right, but maybe that isn't really used anywhere? At least I can't find the place right now without asking the compiler.

Anyway, my patches shouldn't change anything about it. The first commit is a one-liner and a real change (but the old line didn't take msix_one_vector into account either), and the rest is pure refactoring that should be functionally equivalent to what was there before.

viostor/virtio_stor_hw_helper.c
kevmw: I actually had it like that at first (except returning VOID), and found it too ugly to keep it. I think the main thing I didn't like was that I had to dereference the pointers everywhere throughout GetSrbQueueNumber(), and having a function that does only one thing and returns it by value felt better anyway, so you can use it in arbitrary expressions and don't have to remember if it's important to initialise the variable to specific values. Ultimately a matter of taste, though.

After the third commit, MessageId is gone from GetSrbQueueNumber(), so it's not a recalculation any more in the callers, but the only time it's calculated.

benyamin-codez:
I implemented this using the two macros per `vioscsi`, i.e. `QUEUE_TO_MESSAGE()` and `MESSAGE_TO_QUEUE()` but added two more for when `adaptExt->msix_one_vector == TRUE` plus related constants.

I also fixed the locking stuff so that we always use `StorPortAcquireMSISpinLock` when using MSI-X... It currently is broken.

The changes comprehensively address all the edge cases.

I will try to raise the PRs ASAP.

viostor/virtio_stor_hw_helper.c
benyamin-codez: > adaptExt->msix_one_vector ? 0 : MessageID)

It is broken. I have the fix.

kevmw: I spent some time yesterday to understand better how msix_one_vector works. My conclusion is that while some cleanup definitely won't hurt, this code in the PR is not only functionally unchanged, but also QueueNumber = 0 and MessageId = 1 are consistent with how it has to work today for msix_one_vector.

viostor/virtio_stor_hw_helper.c
```diff
@@ -44,49 +44,37 @@
...
-                MessageId = QueueNumber + 1;                                                                           \
-            }                                                                                                          \
-        }                                                                                                              \
-    }
-
 static ULONG GetSrbQueueNumber(IN PVOID DeviceExtension, IN PSRB_TYPE Srb)
 {
     PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
+    STARTIO_PERFORMANCE_PARAMETERS param;
+    ULONG status = STOR_STATUS_SUCCESS;
```
XanClic: This initialization does nothing; it makes sense to keep in commit 2 because all the call sites have it, but we could drop it in this commit.

(Just a personal opinion, because when I see a status/return variable be initialized, I look for the path where it isn’t set. There isn’t such a path here, so I found it strange.  Obviously this is absolute bike-shedding territory where it’s 100 % possible I have an objectively incorrect stylistic opinion.)

XanClic: To add to this bike shedding: It would look more clean to me to also drop the `QueueNumber` initialization and instead pull it down into:
```c
if (param.MessageNumber == 0)
{
    QueueNumber = 0;
}
else
{
    QueueNumber = (param.MessageNumber - 1) % adaptExt->num_queues;
}
```

viostor/virtio_stor_hw_helper.c
benyamin-codez: @kevmw 

Kevin, I'm quite curious about this. If we must always have `msix_vectors = 2` then that is normal operations and not `msix_one_vector`. I do currently have it working (in my WIP) with `MessageId = 0` and `QueueNumber = 0` when `msix_one_vector == TRUE`, which is when `msix_vectors = 1` and `num_queues >= 1`...

AFAICT, it is working like this, but if we need to always have `msix_vectors = 2` then perhaps we should consider decommissioning `msix_one_vector`...?

viostor/virtio_stor_hw_helper.c
```diff
@@ -44,19 +44,39 @@
...
-            {                                                                                                          \
-                QueueNumber %= adaptExt->num_queues;                                                                   \
-                MessageId += 1;                                                                                        \
-            }                                                                                                          \
-        }                                                                                                              \
+static ULONG GetSrbQueueNumber(IN PVOID DeviceExtension, IN PSRB_TYPE Srb)
+{
+    PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
+    STARTIO_PERFORMANCE_PARAMETERS param;
+    ULONG status = STOR_STATUS_SUCCESS;
```
benyamin-codez: Would it be better to use:
```c
ULONG status = STOR_STATUS_UNSUCCESSFUL;
```
...and let `StorPortGetStartIoPerfParams()` set it otherwise...?

viostor/virtio_stor_hw_helper.c
kevmw: Fine with me on both accounts, I pushed a new version.

viostor/virtio_stor_hw_helper.c
kevmw: No, as @XanClic suggested, the value is unused, so if we want to improve this line, then not initialising the variable here at all is what should be done (and what I just did).

viostor/virtio_stor_hw_helper.c
YanVugenfirer: @benyamin-codez I remember on older Windows OSes cases when Windows decided to use legacy interrupts instead of MSI, or some other strange behaviors. I frankly didn’t see it for years.
Are you using QEMU or some other hypervisor?

As I mentioned, we can decide that we always need at least two MSIx vectors to function, but then please add explicit check that we have at least two vectors.

viostor/virtio_stor_hw_helper.c
benyamin-codez: Right. I only mentioned it in case `StorPortGetStartIoPerfParams()` bombed and we were left with `STOR_STATUS_SUCCESS`. As `STOR_STATUS_SUCCESS = 0x0`, the same as the uninitialised value, this is why I suggested `STOR_STATUS_UNSUCCESSFUL`, which is 0xC1000001. I think there was a previous disposition to use the success value when defining at declaration, which I think is somewhat ill-advised.

cc: @XanClic

viostor/virtio_stor_hw_helper.c
benyamin-codez: @YanVugenfirer 

Mostly QEMU. I'm using MSI-X fairly exclusively.

We still use legacy ISRs at `InterruptLock` in `viostor` as the fallback. I still test for that for PRs.
Vadim mentioned some RH customer kit in the field, FPGA kit iirc, has `msix_one_vector` or ancient or no MSI.
The legacy ISRs still work ok for a single queue.

I am aware that some implementations will use MSI-X but then a legacy ISR for device change notifications.
I haven't seen that for `viostor` on QEMU, so I've presumed it was via DPC and StorPort managed. 

I was just trying to understand the rationale behind the statement @kevmw made:

> ...also QueueNumber = 0 and MessageId = 1 are consistent with how it has to work today for msix_one_vector.

...as I have MSI-X working with `MessageId = 0` and `QueueNumber = 0` in my WIP (PR #1299), even with MSI-X restricted to 1 in the registry.

Kevin, did you mean as master stands today, we need it to be `QueueNumber = 0` and `MessageId = 1`...?
Or is there a functional reason, e.g. an underlying QEMU / KVM reason, or Device Change Notifications, or some other reason. I would like to research this further to understand the limitations of the solution I have proposed.

If we do decom `msix_one_vector` we can just change the conditional in `VirtIoHwInitialize()`:

```c
if (!adaptExt->dump_mode && adaptExt->msix_vectors > 1) // change from 0 to 1 
{
...
}
else
{
    /* initialize queues with no MSI interrupts */
    adaptExt->msix_enabled = FALSE;
}
```
...and scrap the rest.

... but if it's working, why would we...? Especially if this is used by a customer somewhere...

My theory is that it currently only works because QEMU supplies the extra vector and it's just there and used anyway... It does appear to work in one vector too though. ¯\\\_(ツ)_/¯

viostor/virtio_stor_hw_helper.c
XanClic: I don’t understand what “bombed” is supposed to mean. Unless this turns out not to be actual C code, it is my understanding that a statement like
```c
status = StorPortGetStartIoPerfParams(DeviceExtension, (PSCSI_REQUEST_BLOCK)Srb, &param);
```
will always overwrite any value that `status` may have had before.

benyamin-codez: Just a colloquialism like "crashed out", "meeting an unhandled fate", "fell off a cliff", etc.

If the routine - here a third-party one - say, does not return a value, the result is _undefined behaviour_. We rely on the third-party routine to not fall off the end of the function body without returning a value and the third-party compiler to warn us if this is possible, but neither is guaranteed. It can be that the value, in this case `status`, may not be updated as a result of this undefined behaviour. In the code above, the uninitialised value might be 0, and `STOR_STATUS_SUCCESS` certainly is, so our test of the value might end up being invalid.

XanClic: I still don’t understand. If there is undefined behavior in `StorPortGetStartIoPerfParams()`, we can’t work around it, and because it’s undefined behavior, anything can happen anyway, and the driver’s behavior as a whole will be unpredictable.

Regarding what would happen in practice if `StorPortGetStartIoPerfParams()` returned before having set up a return value: AFAIU the MS x64 calling convention returns integer values in RAX (EAX in this case, as `StorPortGetStartIoPerfParams()` returns a `ULONG`, which is 32 bit). Even if the function “does not return anything”, i.e. does not deliberately put a valid status code into EAX, there will be some value in it, and (1) there is no reason to assume it would be what we initialized `status` with, and (2) the caller will take this EAX value as the status code, regardless of whether it’s valid or not.

As for (1), the compiler might conceivably have set up EAX/RAX with `status`’s initial value before the call, but I can’t come up with a reason why that would ever happen. A non-optimizing compiler will put the initial value on the stack, not into RAX; and an optimizing compiler will know that in this calling convention, RAX is a volatile register that is not used to pass arguments, so writing something to it right before a call to a third-party function is effectively a no-op.

As for (2), an optimizing compiler would never even set up an actual local variable, but just compare the EAX value after the function call; and a non-optimizing compiler would unconditionally overwrite the local variable’s initial value on the stack with the EAX value after the function call, before making the comparison.

In fact, for even a slightly optimizing compiler, I expect absolutely no difference in the generated machine code of `GetSrbQueueNumber()` whether `status` is initialized or not.

benyamin-codez: I'm pretty sure in C the undefined behaviour occurs at the comparison, and only if the comparison happens. It's been years since I've seen it myself, but in the past I have seen both garbage data and the initial value in situations like the one I described. I guess it depends on the compiler and the config, and moreover the reliability of the called routine, which in this case is arguably very reliable.

I guess this is all somewhat academic, with different schools of thought having different styles and practices. From a pragmatic perspective, starting with a value different to what you expect it to change to can make it easier to see the change when monitoring a values table whilst debugging. In similar vein, when you see a value defined at declaration, you look for the unused path, which is quite prudent as this is certainly where a pre-defined return variable would typically be used.

Anyway, I don't mind either way, just thought it worth mentioning - especially given I caught it after your comment and the push. 8^d

YanVugenfirer: @benyamin-codez The [documentation](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nf-storport-storportgetstartioperfparams) of `StorPortGetStartIoPerfParams` is clear about failure cases.

Did you ever see it fail for other reasons?


```
STOR_STATUS_INVALID_PARAMETER | 
Either the Srb parameter or the StartIoPerfParams parameter is NULL.
-or-

The HwDeviceExtension parameter is NULL or not valid.

-or-

The Srb parameter is not pointing to an SRB sent by Storport

-or-

The structure pointed to by StartIoPerfParams is not valid because the value of its Size member indicates that structure is too small to contain the returned data.

```

Personal comment, only the below is something that might happen, and it will happen for every SRB. The theoretical reason for it might be that we are using outdated driver on a new OS that changed the size of structure. Again. Theoretical case, and will fail on the first usage:
> The structure pointed to by StartIoPerfParams is not valid because the value of its Size member indicates that structure is too small to contain the returned data.

benyamin-codez: I seem to remember _NOT_IMPLEMENTED way back in the early days, probably when dragging Win2k clients of various service packs up-level. Probably all of those _INVALID_PARAMETER causes at one stage or another too, but yes version upgrades can produce this one as well, but usually only with a NTDDI mismatch when cross-compiling or with the wrong target.

My comment was really only because I viewed the PR due to a push notification for the commit (before seeing Kevin's comment) and noticed the _SUCCESS definition after Hanna's comment was wiped by the push (I was looking in the _Files changed_ tab). I also didn't take the time to check for an unused path because these variables can be predefined as a matter of style and practice. I was coming at it from a general and stylistic perspective without regard to `StorPortGetStartIoPerfParams()` proper.

I've generally used STOR_STATUS_UNSUCCESSFUL for any predefined StorPort routine status variables (when used - and only for LONG or ULONG declarations). This routine doesn't return it, but as it is a non-zero value and clearly a fail-state, it has worked well for me in the past.

My apologies for any inconvenience caused by the academic conversation.

YanVugenfirer: @benyamin-codez No problem.

viostor/virtio_stor_hw_helper.c
```diff
@@ -44,19 +44,43 @@
...
+    if (status != STOR_STATUS_SUCCESS)
+    {
+        RhelDbgPrint(TRACE_LEVEL_ERROR, " StorPortGetStartIoPerfParams failed. srb %p status 0x%x.\n", Srb, status);
+        return 0;
+    }
+
+    if (param.MessageNumber == 0)
+    {
+        QueueNumber = 0;
     }
```
benyamin-codez, line 70: Wouldn't this bump `MessageId` on return to the caller, i.e. `param.MessageNumber` is 0 but after returning it will be incremented and another (unprepared by StorPort) vector will be used (presuming it exists).

benyamin-codez:
Kevin, I've been splitting relevant commits from my WIP and thought it prudent to mention a couple of things having looked at your PR in more depth too.

I note Yan's previous review comment about returning `MessageId` too (via a pointer) and your reply about dereferencing and ugliness, which I certainly can appreciate.

I didn't come across these problems because I consolidated the four functions (`RhelDoFlush()`, `RhelDoReadWrite()`, `RhelDoUnMap()` & `RhelGetSerialNumber()`) into a new `SendSRB()` function based on the same in `vioscsi`, which imho is a better implementation of this part of the common StorPort functional code. In that function I was able to use:

```c
if (resend == SEND_SRB_ALREADY_UNDER_SPINLOCK)
{
    MessageId = srbExt->MessageID;
}
else if (adaptExt->num_queues > 1)
{
    param.Size = sizeof(STARTIO_PERFORMANCE_PARAMETERS);
    status = StorPortGetStartIoPerfParams(DeviceExtension, (PSCSI_REQUEST_BLOCK)Srb, &param);
    if (status == STOR_STATUS_SUCCESS && param.MessageNumber != 0)
    {
        MessageId = param.MessageNumber;
    }
}
else if (adaptExt->msix_enabled && adaptExt->msix_one_vector)
{
    MessageId = VIRTIO_BLK_MSIX_1_VECTOR_MSG_ID;
}
else
{
    MessageId = 1;
}

if (adaptExt->msix_enabled && adaptExt->msix_one_vector)
{
    QueueNumber = MESSAGE_TO_QUEUE_1_VECTOR(MessageId);
}
else
{
    QueueNumber = MESSAGE_TO_QUEUE(MessageId);
    if (QueueNumber >= adaptExt->num_queues + VIRTIO_BLK_REQUEST_QUEUE_0)
    {
        RhelDbgPrint(TRACE_VQ,
                        " Modulo assignment required for QueueNumber as it"
                        " exceeds the number of virtqueues available.\n");
        QueueNumber %= adaptExt->num_queues;
    }
}

if (resend == SEND_SRB_NO_EXISTING_SPINLOCK && adaptExt->num_queues > 1)
{
    if (status == STOR_STATUS_SUCCESS && param.MessageNumber != 0)
    {
        RhelDbgPrint(TRACE_PERF,
                        " StorPortGetStartIoPerfParams for SRB %p :"
                        " QueueNumber %lu | MessageNumber %lu | ChannelNumber %lu \n",
                        Srb,
                        QueueNumber,
                        MessageId,
                        param.ChannelNumber);
    }
    else
    {
        RhelDbgPrint(TRACE_LEVEL_ERROR,
                        " StorPortGetStartIoPerfParams failed. SRB : %p | Status : 0x%x \n",
                        Srb,
                        status);
    }
}
```

...together with:

```c
#define VIRTIO_BLK_MSIX_VQ_OFFSET            1
#define VIRTIO_BLK_MSIX_VQ_1_VCTR_OFFSET     0
#define VIRTIO_BLK_MSIX_1_VECTOR_MSG_ID      0
#define VIRTIO_BLK_MSIX_CONFIG_VECTOR        0
#define VIRTIO_BLK_REQUEST_QUEUE_0           0
#define VIRTIO_BLK_QUEUE_LAST                MAX_CPU
...
#define QUEUE_TO_MESSAGE(QueueId)            ((QueueId) + VIRTIO_BLK_MSIX_VQ_OFFSET)
#define MESSAGE_TO_QUEUE(MessageId)          ((MessageId)-VIRTIO_BLK_MSIX_VQ_OFFSET)
#define QUEUE_TO_MESSAGE_1_VECTOR(QueueId)   ((QueueId) + VIRTIO_BLK_MSIX_VQ_1_VCTR_OFFSET)
#define MESSAGE_TO_QUEUE_1_VECTOR(MessageId) ((MessageId)-VIRTIO_BLK_MSIX_VQ_1_VCTR_OFFSET)
```

To properly address `msix_one_vector` with your patch as-is, I think this would still need to be handled by the caller.

On style, I can envisage that newcomers with a Windows focus may look at your use of `return int;` rather than `return QueueNumber;` and misunderstand, but refactoring for this would require a conditional ladder before a universal return.

I also note my review comment on lines R67-70 wrt the caller incrementing `MessageId` away from the StorPort prepared vector.

Hoping to have a minimalist PR for consideration out in the next 24 hours...

viostor/virtio_stor_hw_helper.c
kevmw: What I'm saying is only about how the code is in current master. It definitely ought to be cleaned up, but for now, in `msi_one_vector` mode, you need `MessageId = 1`. One of the reasons is the `VioStorVQLock()` code Yan posted above. If you had `MessageId = 0` and `num_queues > 1` (which is easily possible for `msi_one_vector`), you would both fail the assertion and access `adaptExt->dpc[-1]`, which is obviously out of bounds.

This doesn't mean that you'll actually get the interrupt with message 1, it's still message 0. But in the code, you still have `MessageId = 1`. See also `VirtIoMSInterruptRoutine()`, which sets the same value again as the first thing in `msi_one_vector` mode, ignoring the actual `MessageID` it got passed (which will be 0 in practice).

This also means that I'm not suggesting that we should require at least two vectors. While the code is a bit confusing, it works with just one.

And I can only repeat what I already said: This PR really changes absolutely nothing about this case. It's working exactly the same as it is currently in master. I'm planning to send a separate PR to clean up the confusing parts, though.

viostor/virtio_stor_hw_helper.c
kevmw: Yes, `MessageId` will be 1 in the caller. The only thing it is used for is `VioStorVQLock/Unlock()`, which expects the value like this. The only other use is storing it in `srbExt->MessageID`, so that `RhelDoFlush()` can reuse it, for the same `VioStorVQLock/Unlock()` calls that require it like this.

The request will be queued in queue 0 (according to the returned `QueueNumber`) and vector 0 will be raised on completion if `msix_one_vector == TRUE`, or vector 1 if `msix_one_vector == FALSE`. It doesn't use any extra vector that isn't prepared.

kevmw:
We can certainly try to refactor things to remove some more duplication, but I see this as a separate change from this one. I think having a function to figure out the right queue number for a request is still the right thing to do then.

Not completely sure what you mean with `return int`?

viostor/virtio_stor_hw_helper.c
benyamin-codez: We normally ignore `param.MessagNumber == 0` because it is what `StorPortGetStartIoPerfParams()` returns when there is no MSI-X support. Thus we normally use this pathway:
```c
if (status == STOR_STATUS_SUCCESS && param.MessageNumber != 0)
```
So I presume this must be for the msix_one_vector / legacy ISR scenario, and the off-by-one in the caller can be fixed in the caller or elsewhere in another PR...

In any case, I'm guessing this only works because vector 1 is unnecessarily provided (in the case of msix_one_vector) and the legacy ISR doesn't care...?

benyamin-codez:
> Let's merger as is now.
> 
> It is clear that the refactoring of the MessageID/QueueID calculation is needed and it shouldn't be spread between several location in the code to prevent confusion or possible bugs.

imho, for the same reasons it would have been nice to fix `MessagId` here too, as this is also in every caller. 

In my research, the ideal order was `MessageId` first (as this covers the various cases), then `QueueNumber` (including modulo), then report for tracing.

Apologies for not producing an alternate PR yet. I had an unexpected medical issue to deal with. I've got several weeks invested in researching and extensively testing this and related issues and not for the need to split my commits more, I would have had a PR raised already. I'm somewhat concerned I'm wasting my time now things are moving quickly...

benyamin-codez:
> We can certainly try to refactor things to remove some more duplication, but I see this as a separate change from this one. I think having a function to figure out the right queue number for a request is still the right thing to do then.

Happy to provide some or all of this... As I said, I've also already got fixes for the locking bugs that overlap with this.

> Not completely sure what you mean with `return int`?

Just the `return 0;`. It's not as intuitive and semantically mnemonic as `return QueueNumber;`, but to use the latter would require refactoring your patch to use a conditional ladder to set `QueueNumber` before returning with it.

PR #1497, opened 2026-01-21
viostor: Resolve Coverity defects

elizashurov:
Add fallthrough comment for WRITE/READ switch cases

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -1154,6 +1154,7 @@ VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
                     return TRUE;
                 }
             }
+            /* fallthrough */
         case SCSIOP_READ:
         case SCSIOP_READ16:
             {


viostor/virtio_stor.c
```diff
@@ -866,9 +866,9 @@ VOID CompletePendingRequests(IN PVOID DeviceExtension)
                 {
                     pblk_req req = CONTAINING_RECORD(entry, blk_req, list_entry);
                     PSCSI_REQUEST_BLOCK currSrb = (PSCSI_REQUEST_BLOCK)req->req;
-                    PSRB_EXTENSION currSrbExt = SRB_EXTENSION(currSrb);
                     if (currSrb)
                     {
+                        PSRB_EXTENSION currSrbExt = SRB_EXTENSION(currSrb);
```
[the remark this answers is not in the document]
benyamin-codez: `currSrb` can also just be `Srb`...

Best see PR #1493 too...

viostor/virtio_stor.c
```diff
@@ -1058,6 +1067,12 @@ VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
                     CompleteRequestWithStatus(DeviceExtension, (PSRB_TYPE)Srb, SrbStatus);
                     return TRUE;
                 }
+                SRB_SET_SRB_STATUS(Srb, SRB_STATUS_PENDING);
+                if (!RhelDoReadWrite(DeviceExtension, (PSRB_TYPE)Srb))
+                {
+                    CompleteRequestWithStatus(DeviceExtension, (PSRB_TYPE)Srb, SRB_STATUS_BUSY);
+                }
+                return TRUE;
```
benyamin-codez: This is already handled by the SCSIOP_READ16 case below (there is no break here so the write operation flows down into the SCSIOP_READ* case).

YanVugenfirer:
I suggest splitting commits for each fix.

elizashurov:
> I suggest splitting commits for each fix.
Done

viostor/virtio_stor.c
elizashurov: Thanks. You are right. I replaced the redundant code with a /* Fallthrough */ comment to silence static analysis warnings.

kevmw:
The currSrb patch doesn't apply any more, the problem is already gone in master.

The (UCHAR) casts don't feel right to me because they imply that we intentionally want to truncate the value there. We don't, we know that the value is already safe to convert. I'd check how Coverity thinks an integer overflow can be triggerred, but it seems to be down at the moment. If we can't make it understand the logic (or there is a real bug somewhere), it's probably better to mark it as a false positive.

I agree with the fall-through comment. It actually confused me for a second when first reading the code.

benyamin-codez:
> I agree with the fall-through comment. It actually confused me for a second when first reading the code.

Yeah, when reading switch-case ladders, I look for the breaks...
It's a less intuitive style but admittedly more efficient and less repetitive.

YanVugenfirer:
@elizashurov Please rebase

YanVugenfirer:
@elizashurov please rebase

elizashurov:
> @elizashurov please rebase

Done, I removed the unnecessary changes. Only the fallthrough comment remains

PR #1498, opened 2026-01-21
viosock: Resolve Coverity defects

elizashurov:
Fix Coverity CID 475471 by adding VOID return type to VIOSockReadProcessDequeueCb. 
Fix Coverity CID 475412 by preventing DWORD-to-int overflow in Send() function and fixing infinite loop caused by checking pointer instead of value.


kostyanf14:
rerun tests

PR #1302, opened 2025-02-17
[CI-NO-BUILD] [viostor] Refactor RhelSetGuestFeatures()

benyamin-codez:
Refactors `RhelSetGuestFeatures()` to improve instrumentation.
Also added `TRACE_GUEST_FEATURES` WPP tracing flag.

PR #1301, opened 2025-02-17
[viostor] Refactor RhelGetDiskGeometry()

benyamin-codez:
Refactors `RhelGetDiskGeometry()` to improve instrumentation.

PR #1499, opened 2026-01-23
[viostor] WIP-phase1 - Various fixes and enhancements

benyamin-codez:
Here we have the first phase of my current Work In Progress on `viostor` for your consideration.

I will update this post with child PRs as they are raised or updated as the case may be.

Please consider keeping review and discussion in these child PRs where possible.

Please also give me a couple of days to raise and update same.

Best regards,
Ben

Child PRs:

* PR #1289 (Ready)
* PR #1297 (To Be Updated)
* PR #1298 (To Be Updated)
* PR #1299 (To Be Updated)
* PR #1300 (To Be Updated)
* PR #1301 (To Be Updated)
* PR #1302 (To Be Updated)
* PR #1304 (Ready)
* PR #1315 (Ready)

* `adaptExt->num_queues` as ULONG
* struct member alignment
* Locking
* Interrupts
* Tracing

More to follow... 8^{d>~

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -30,6 +30,9 @@
  * SUCH DAMAGE.
  */
 #include "virtio_stor.h"
+#include "ntddi_ver.h"
+#include "virtio_stor_reg_helper.h"
+
 #if defined(EVENT_TRACING)
 #include "virtio_stor.tmh"
 #endif
@@ -37,42 +40,29 @@
 BOOLEAN IsCrashDumpMode;
 
 #ifdef EVENT_TRACING
-PVOID TraceContext;
+PVOID TraceContext = NULL;
+VOID WppCleanupRoutine(PVOID arg1)
+{
+    RhelDbgPrint(TRACE_LEVEL_INFORMATION, " WppCleanupRoutine\n");
+    WPP_CLEANUP(NULL, TraceContext);
+}
 #endif
 
 sp_DRIVER_INITIALIZE DriverEntry;
 HW_INITIALIZE VirtIoHwInitialize;
+HW_BUILDIO VirtIoBuildIo;
 HW_STARTIO VirtIoStartIo;
 HW_FIND_ADAPTER VirtIoFindAdapter;
 HW_RESET_BUS VirtIoResetBus;
 HW_ADAPTER_CONTROL VirtIoAdapterControl;
+HW_UNIT_CONTROL VirtIoUnitControl;
 HW_INTERRUPT VirtIoInterrupt;
-HW_BUILDIO VirtIoBuildIo;
-HW_DPC_ROUTINE CompleteDpcRoutine;
 HW_MESSAGE_SIGNALED_INTERRUPT_ROUTINE VirtIoMSInterruptRoutine;
+HW_DPC_ROUTINE CompleteDpcRoutine;
 HW_PASSIVE_INITIALIZE_ROUTINE VirtIoPassiveInitializeRoutine;
 
-extern int vring_add_buf_stor(IN struct virtqueue *_vq,
-                              IN struct VirtIOBufferDescriptor sg[],
-                              IN unsigned int out,
-                              IN unsigned int in,
-                              IN PVOID data);
-
-BOOLEAN
-VirtIoHwInitialize(IN PVOID DeviceExtension);
-
-BOOLEAN
-VirtIoHwReinitialize(IN PVOID DeviceExtension);
-
-BOOLEAN
-VirtIoBuildIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb);
-
-VOID CompleteDpcRoutine(IN PSTOR_DPC Dpc, IN PVOID Context, IN PVOID SystemArgument1, IN PVOID SystemArgument2);
-BOOLEAN
-VirtIoMSInterruptRoutine(IN PVOID DeviceExtension, IN ULONG MessageID);
-
-BOOLEAN
-VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb);
+ULONG
+DriverEntry(IN PVOID DriverObject, IN PVOID RegistryPath);
 
 ULONG
 VirtIoFindAdapter(IN PVOID DeviceExtension,
@@ -82,11 +72,47 @@ VirtIoFindAdapter(IN PVOID DeviceExtension,
                   IN OUT PPORT_CONFIGURATION_INFORMATION ConfigInfo,
                   OUT PBOOLEAN Again);
 
+static ULONG InitVirtIODevice(PVOID DeviceExtension);
+
+BOOLEAN
+VirtIoHwInitialize(IN PVOID DeviceExtension);
+
+BOOLEAN
+VirtIoHwReinitialize(IN PVOID DeviceExtension);
+
+BOOLEAN
+VirtIoPassiveInitializeRoutine(IN PVOID DeviceExtension);
+
+static BOOLEAN InitializeVirtualQueues(PADAPTER_EXTENSION adaptExt, ULONG numQueues);
+
+BOOLEAN
+VirtIoBuildIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb);
+
+BOOLEAN
+VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb);
+
+BOOLEAN
+VirtIoInterrupt(IN PVOID DeviceExtension);
+
+BOOLEAN
+VirtIoMSInterruptRoutine(IN PVOID DeviceExtension, IN ULONG MessageId);
+
+static BOOLEAN VirtIoMSInterruptWorker(IN PVOID DeviceExtension, IN ULONG MessageId);
+
+VOID CompleteDpcRoutine(IN PSTOR_DPC Dpc, IN PVOID Context, IN PVOID SystemArgument1, IN PVOID SystemArgument2);
+
+VOID CompletePendingRequests(IN PVOID DeviceExtension);
+
+VOID DispatchQueue(IN PVOID DeviceExtension, IN ULONG MessageId);
+
 BOOLEAN
 VirtIoResetBus(IN PVOID DeviceExtension, IN ULONG PathId);
 
 SCSI_ADAPTER_CONTROL_STATUS
 VirtIoAdapterControl(IN PVOID DeviceExtension, IN SCSI_ADAPTER_CONTROL_TYPE ControlType, IN PVOID Parameters);
+
+SCSI_UNIT_CONTROL_STATUS
+VirtIoUnitControl(IN PVOID DeviceExtension, IN SCSI_UNIT_CONTROL_TYPE ControlType, IN PVOID Parameters);
 
 UCHAR
 RhelScsiGetInquiryData(IN PVOID DeviceExtension, IN OUT PSRB_TYPE Srb);
@@ -105,56 +131,45 @@ RhelScsiVerify(IN PVOID DeviceExtension, IN OUT PSRB_TYPE Srb);
 UCHAR
 RhelScsiReportLuns(IN PVOID DeviceExtension, IN OUT PSRB_TYPE Srb);
 
-VOID FORCEINLINE CompleteSRB(IN PVOID DeviceExtension, IN PSRB_TYPE Srb);
-
-VOID FORCEINLINE DeviceChangeNotification(IN PVOID DeviceExtension, IN BOOLEAN bLun);
-
-BOOLEAN
-FORCEINLINE
-SetSenseInfo(IN PVOID DeviceExtension, IN PSRB_TYPE Srb);
+VOID FORCEINLINE DeviceChangeNotification(IN PVOID DeviceExtension,
+                                          IN BOOLEAN bLun,
+                                          IN INL_FUNC_IDX idx_ICN,
+                                          IN INL_FUNC_IDX idx_IFN);
 
 BOOLEAN
 FORCEINLINE
-CompleteDPC(IN PVOID DeviceExtension, IN ULONG MessageID);
+SetSenseInfo(IN PVOID DeviceExtension, IN PSRB_TYPE Srb, IN INL_FUNC_IDX idx_ICN, IN INL_FUNC_IDX idx_IFN);
+
+UCHAR DeviceToSrbStatus(IN UCHAR status);
 
 UCHAR FirmwareRequest(IN PVOID DeviceExtension, IN PSRB_TYPE Srb);
 
 VOID ReportDeviceIdentifier(IN PVOID DeviceExtension, IN PSRB_TYPE Srb);
 
-#ifdef EVENT_TRACING
-VOID WppCleanupRoutine(PVOID arg1)
-{
-    RhelDbgPrint(TRACE_LEVEL_INFORMATION, " WppCleanupRoutine\n");
-
-    WPP_CLEANUP(NULL, TraceContext);
-}
-#endif
-
 ULONG
-DriverEntry(IN PVOID DriverObject, IN PVOID RegistryPath)
+DriverEntry(PVOID DriverObject, PVOID RegistryPath)
 {
 
     HW_INITIALIZATION_DATA hwInitData;
     ULONG initResult;
+    ANSI_STRING aRegistryPath;
+    NTSTATUS u2a_status;
 
 #ifdef EVENT_TRACING
     STORAGE_TRACE_INIT_INFO initInfo;
-#else
-// FIXME
-#ifdef DBG
+#endif
+
     InitializeDebugPrints((PDRIVER_OBJECT)DriverObject, (PUNICODE_STRING)RegistryPath);
-#endif
-#endif
 
     RhelDbgPrint(TRACE_LEVEL_ERROR, " Viostor driver started...built on %s %s\n", __DATE__, __TIME__);
     IsCrashDumpMode = FALSE;
     if (RegistryPath == NULL)
     {
-        RhelDbgPrint(TRACE_LEVEL_INFORMATION, " DriverEntry: Crash dump mode\n");
         IsCrashDumpMode = TRUE;
+        RhelDbgPrint(TRACE_LEVEL_INFORMATION, " DriverEntry: Crash dump mode\n");
     }
 
-    memset(&hwInitData, 0, sizeof(HW_INITIALIZATION_DATA));
+    RtlZeroMemory(&hwInitData, sizeof(HW_INITIALIZATION_DATA));
 
     hwInitData.HwInitializationDataSize = sizeof(HW_INITIALIZATION_DATA);
 
@@ -164,7 +179,9 @@ DriverEntry(IN PVOID DriverObject, IN PVOID RegistryPath)
     hwInitData.HwInterrupt = VirtIoInterrupt;
     hwInitData.HwResetBus = VirtIoResetBus;
     hwInitData.HwAdapterControl = VirtIoAdapterControl;
+    hwInitData.HwUnitControl = VirtIoUnitControl;
     hwInitData.HwBuildIo = VirtIoBuildIo;
+
     hwInitData.NeedPhysicalAddresses = TRUE;
     hwInitData.TaggedQueuing = TRUE;
     hwInitData.AutoRequestSense = TRUE;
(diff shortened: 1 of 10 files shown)


YanVugenfirer:

benyamin-codez:
Some performance pics:

win10_random_4k_o32_t16_r_iops_variance_count

win10_random_4k_o32_t16_r_iops_variance_latency

win11_random_4k_o32_t16_r_iops_variance_latency

win11_random_64k_o32_t16_r_iops_variance_count

win10_random_64k_o32_t16_r_iops_variance_count

win10_random_64k_o32_t16_r_iops_variance_latency

win10_sequential_1m_o8_t1_s_iops_variance_count

win10_sequential_1m_o32_t8_T4k_s8192k_iops_variance_count

viostor/virtio_stor.h
```diff
@@ -152,10 +167,11 @@ typedef struct virtio_blk_config
      */
     u8 write_zeroes_may_unmap;
 
-    u8 unused1[3];
+    u8 unused1[7];
```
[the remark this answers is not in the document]
benyamin-codez: Comment added.

viostor/virtio_stor_hw_helper.c
```diff
@@ -40,525 +41,398 @@
...
+                                 BlockDescrIndex,
+                                 BlockDescrCount,
+                                 blockDescrStartingLba,
+                                 blockDescrLbaCount);
+                    adaptExt->blk_discard[BlockDescrIndex].sector = blockDescrStartingLba *
+                                                                    (adaptExt->info.blk_size / SECTOR_SIZE);
+                    adaptExt->blk_discard[BlockDescrIndex].num_sectors = blockDescrLbaCount *
+                                                                         (adaptExt->info.blk_size / SECTOR_SIZE);
+                    adaptExt->blk_discard[BlockDescrIndex].flags = 0;
+                }
```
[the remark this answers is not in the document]
benyamin-codez: This was lifted from `RhelDoUnMap()`.
Will address in child / separately...

kevmw:
@benyamin-codez I expected a series of smaller cleanups to wait for or rebase on, but I didn't realise that you have effectively a whole driver rewrite consisting of unreviewed code. I'm afraid you have manoeuvred yourself into a dead end here, having so much out-of-tree code is bound to end in pain. There is no way we can wait for all of this to land before we touch the driver again, and even if we waited, review of each step would probably cause changes and constant rebasing pains, too.

So I think we should go ahead with all the other fixes and cleanups and basically treat your tree as a source of ideas, but not as something to be merged as-is. And then we can work together on the current state in master, with small incremental self-contained changes.

benyamin-codez:
Yeah, that's fine. That was the expectation.

benyamin-codez:
@kevmw 

I still think it worthwhile grabbing the MSI-X parts sooner rather than later. In particular, I see some value to retaining the type of mechanism in `vioscsi` as to mapping QUEUENUMBER_TO_MESSAGE and MESSAGE_TO_QUEUENUMBER, per:

```c
#define QUEUE_TO_MESSAGE(QueueId)            ((QueueId) + VIRTIO_BLK_MSIX_VQ_OFFSET)
#define MESSAGE_TO_QUEUE(MessageId)          ((MessageId)-VIRTIO_BLK_MSIX_VQ_OFFSET)
#define QUEUE_TO_MESSAGE_1_VECTOR(QueueId)   ((QueueId) + VIRTIO_BLK_MSIX_VQ_1_VCTR_OFFSET)
#define MESSAGE_TO_QUEUE_1_VECTOR(MessageId) ((MessageId)-VIRTIO_BLK_MSIX_VQ_1_VCTR_OFFSET)
```

This is one reason why I've been developing a clearer PR for your consideration. I've progressed work on this but as you can imagine the number of commits I understand you would prefer me to provide isn't insignificant. I'm quickly running out of time before I must attend to other matters for about 24 hours. Perhaps on the other side of the weekend I will have something more appropriate for you to review...

Does this work for you, @XanClic and anyone else interested..?

viostor/virtio_stor_hw_helper.c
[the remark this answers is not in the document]
benyamin-codez: @XanClic

Hanna, did you already catch this one in your review of master...?

cc: @kevmw @YanVugenfirer

benyamin-codez:
Looks like I've got a little cleanup of the registry stuff... 8^d

viostor/virtio_stor_hw_helper.c
```diff
@@ -754,77 +656,92 @@ VOID RhelGetDiskGeometry(IN PVOID DeviceExtension)
...
-            if (adaptExt->num_queues > 1)
+            RhelDbgPrint(TRACE_LOCKS, " Processing StorPortAcquireMSISpinLock() for MessageId : %d \n", MessageId);
+            ULONG oldIrql = 0;
+            StorPortAcquireMSISpinLock(DeviceExtension, MessageId, &oldIrql);
+            LockHandle->Context.OldIrql = (KIRQL)oldIrql;
+        }
+        else
+        {
+            // Unused legacy pathway for when adaptExt->msix_sync_mode == InterruptSynchronizeAll
+            ULONG QueueNumber;
```
kevmw, line 695: Don't go through QueueNumbers here, DPCs are primarily about messages, not queues (and yes, the code is currently a bit confused about this). After #1503, this should just call `MessageToDpcIdx()` instead.

viostor/virtio_stor_hw_helper.c
```diff
@@ -754,77 +656,92 @@ VOID RhelGetDiskGeometry(IN PVOID DeviceExtension)
...
+    PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
+    if (adaptExt->msix_enabled)
     {
-        if (adaptExt->msix_enabled)
+        if (adaptExt->msix_sync_mode == InterruptSynchronizePerMessage)
         {
-            if (adaptExt->num_queues > 1)
+            RhelDbgPrint(TRACE_LOCKS, " Processing StorPortAcquireMSISpinLock() for MessageId : %d \n", MessageId);
+            ULONG oldIrql = 0;
+            StorPortAcquireMSISpinLock(DeviceExtension, MessageId, &oldIrql);
```
kevmw, line 689: This is also a question about the preexisting code: What is the advantage of `StorPortAcquireMSISpinLock()` in our case? Generally speaking it seems to be that you can selectively lock only for a specific message, but the way we use the `adaptExt->dpc` array per message with `StorPortAcquireSpinLock()` effectively achieves the same thing.

Could we simplify and just use the method that is always available?

viostor/virtio_stor_hw_helper.c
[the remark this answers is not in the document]
XanClic: Yes: https://github.com/virtio-win/kvm-guest-drivers-windows/commit/7ab33cf492fbec197ee838abb1fc99e8e98746ae

(Longer list of potential issues here: https://github.com/mstsirkin/kvm-guest-drivers-windows/commits/review1, and here: https://github.com/mstsirkin/kvm-guest-drivers-windows/pull/1/commits)

viostor/virtio_stor_hw_helper.c
benyamin-codez: Yeah, I think it's a legacy holdover.

viostor/virtio_stor_hw_helper.c
benyamin-codez: The advantage is primarily improved SMP performance due to the way StorPort calls the interrupt callback.

So even if we call `StorPortAcquireSpinLock()` with `DpcType`, StorPort still acquires an interrupt spin lock on ALL messages and then calls the `HW_MESSAGE_SIGNALED_INTERRUPT_ROUTINE` at DIRQL. It is not sufficient to set the `InterruptSynchronizationMode` to `InterruptSynchronizePerMessage` because even in this mode, if we call `StorPortAcquireSpinLock()`, StorPort will still acquire a interrupt lock on all messages. This is the mechanism used to synchronise with all messages when the need arises even in `InterruptSynchronizePerMessage` mode.

To take advantage of the performance benefit of `InterruptSynchronizePerMessage` mode, we need to exclusively call spinlocks using `StorPortAcquireMSISpinLock()`. In the edge case when we need to synchronise with all messages - and I'm not aware of any in the current implementation because we now always address the DPC context anyway - we would then use `StorPortAcquireSpinLock()` directly rather than via `VioStorVQLock()`.

I included the `StorPortAcquireSpinLock()` conditional fork for completeness and in case we wanted to use the `INTERRUPT_SYNCHRONIZATION_MODE` enumerator. down the track, by say exposing it to a registry switch.

As it stands I just did this:

https://github.com/virtio-win/kvm-guest-drivers-windows/blob/7229eaba0e524fe2866093340db35f9b4d97b259/viostor/virtio_stor.c#L413-L429

...where the only alternative is for our `InterruptLock` type fallback.

It's possible we originally implemented `InterruptSynchronizePerMessage` mode hoping to get the performance gains but then stumbled. It could date back to early MSI-X implementation. I think it was the right choice but we didn't nail the implementation.

viostor/virtio_stor.c
```diff
@@ -1392,16 +2234,18 @@ VirtIoBuildIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
                      lba,
                      adaptExt->lastLBA);
         CompleteRequestWithStatus(DeviceExtension, (PSRB_TYPE)Srb, SRB_STATUS_BAD_SRB_BLOCK_LENGTH);
+        EXIT_FN_SRB();
         return FALSE;
     }
-    if ((lba + blocks) > adaptExt->lastLBA)
+    if ((lba + blocks) > adaptExt->lastLBA - 1)
```
benyamin-codez, line 2240: @XanClic 

I note 86394da (Issue B) touched this part.

I included this part as reported by @EnergyFaith in commentary on 67f64d0 for issue #829.

Did you want to include it in your lists?

viostor/virtio_stor_hw_helper.c
benyamin-codez: > I included the `StorPortAcquireSpinLock()` conditional fork for completeness and in case we wanted to use the `INTERRUPT_SYNCHRONIZATION_MODE` enumerator, down the track, by say exposing it to a registry switch.

@kevmw 

Kevin, I should also mention that I retained the pathway because I also wanted to experiment with spinlocks again in `vioscsi`. My previous attempts to get `StorPortAcquireMSISpinLock()` working there were not terribly successful, but I now suspect that this was more about `msix_one_vector` and  the `MessageNumber` maybe being off...

It works quite well here in `viostor` - in combination with other changes in this PR - using 4 vCPUs the average latency for 4KiB random reads reduced by over 50% to about 1ms, and 64KiB went down to a hair over 6ms.

cc: @YanVugenfirer

kevmw: > It is not sufficient to set the InterruptSynchronizationMode to InterruptSynchronizePerMessage because even in this mode, if we call StorPortAcquireSpinLock(), StorPort will still acquire a interrupt lock on all messages.

If it does indeed acquire an interrupt lock on all messages, that sounds significant. I tried to find this behaviour in the documentation, but couldn't find it. Instead, `StorPortAcquireSpinLock()` documents that you can explicitly take an `InterruptLock` after acquiring a `DpcLock`, which implies that it won't already be held. Do you have a pointer for me where this is explained?

benyamin-codez: Apologies, Kevin. I wasn't especially clear and conflated my response.

With regard to the `HW_MESSAGE_SIGNALED_INTERRUPT_ROUTINE` callback, which is `HwMSIInterruptRoutine` and `VirtIoMSInterruptRoutine()` in our `viostor` driver. In this routine we don't actually call any spin locks, but the `InterruptSynchronizationMode` affects how this operates.

If you look at the [documentation for `StorPortAquireSpinlock()`](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nf-storport-storportacquirespinlock), in the tables you will see for `HwMSIInterruptRoutine` that the port driver (StorPort) acquires and holds an `InterruptLock` type spin lock and that you cannot acquire a spin lock in the miniport driver (in this case `viostor`), i.e. Allowed spin locks = None.

The [`HW_MESSAGE_SIGNALED_INTERRUPT_ROUTINE` documentation](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/ne-storport-_interrupt_synchronization_mode) gives a more concise overview too.

Generally, when we call `StorPortAquireSpinlock()` we must honour the tables per the documentation, but if you look at the [documentation for `StorPortAquireMSISpinlock()`](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nf-storport-storportacquiremsispinlock) you will see that it can be called at any IRQL. This is because StorPort will manage the level hierarchy.

Again, the advantage is better SMP performance.

kevmw: I'm not sure how `HW_MESSAGE_SIGNALED_INTERRUPT_ROUTINE` is even relevant here. Yes, it holds the per-message interrupt spinlock, but in the common case, we don't do anything interesting there anyway, but just issue a DPC. In particular, we don't call the code in `VioStorVQLock()` that you're changing here.

What does change is just that `VirtIoMSInterruptRoutine()` doesn't get called while you hold the MSISpinlock. If you only hold the `DpcLock`, the interrupt handler would still be called and only the DPC it issues would have to wait. If this is our case, the question is why we even get additional interrupts while we're already processing the queue. My understanding is that `virtqueue_disable_cb()` is supposed to prevent this.

In either case, I don't see anything that would affect all messages if `InterruptSynchronizationMode` is `InterruptSynchronizePerMessage`.

benyamin-codez: > I'm not sure how HW_MESSAGE_SIGNALED_INTERRUPT_ROUTINE is even relevant here...

The relevance is only in which `InterruptSynchronizationMode` you choose.

> What does change is just that `VirtIoMSInterruptRoutine()` doesn't get called while you hold the MSISpinlock. If you only hold the `DpcLock`, the interrupt handler would still be called and only the DPC it issues would have to wait. If this is our case, the question is why we even get additional interrupts while we're already processing the queue. My understanding is that `virtqueue_disable_cb()` is supposed to prevent this.

Is that true? I thought it was the opposite, i.e. that `VirtIoMSInterruptRoutine()` doesn't get called while you hold `StorPortAcquireSpinLock()`, and moreover that `VirtIoMSInterruptRoutine()` can be called for other vectors when you use `StorPortAcquireMSISpinLock()`. Also `virtqueue_disable_cb()` is per virtqueue, so should only block for the vector currently under lock.

Additional interrupts would normally be for other vectors, trying to process their own DPC dispatch queues on other CPUs. My understanding is that Storport won't call `VirtIoMSInterruptRoutine()` if any vector acquires a `StorPortAcquireSpinLock()`, because this is the mechanism used to synchronise with all messages when in `InterruptSynchronizePerMessage` mode, per the relevant [documentation](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/storport/nf-storport-storportacquiremsispinlock):

> _When a miniport needs to synchronize with all messages, it can use one call to **StorPortAcquireSpinLock** which will acquire a lock for each message in the proper order._

Further to the above, when preparing a `DpcLock` type spin lock for `StorPortAcquireSpinLock()` the additional steps required also have a performance overhead. From what I have seen `StorPortAcquireMSISpinLock()` is more efficient and will provide better SMP performance.

Also, as previously mentioned, from the perspective of convenience, when using `StorPortAcquireMSISpinLock()`, one also doesn't need to be concerned with the IRQL you are in or from which routine you are acquiring the lock.

benyamin-codez: I should mention, that in any case, it is better to test for:
```c
if (adaptExt->msix_sync_mode == InterruptSynchronizePerMessage)
```
...or similar, rather than:
```c
if (adaptExt->num_queues <= 1)
```
...in deciding whether to use `StorPortAcquireMSISpinLock()`...

It seems quite counter-intuitive the current way, yes?

kevmw: > Is that true? I thought it was the opposite, i.e. that VirtIoMSInterruptRoutine() doesn't get called while you hold StorPortAcquireSpinLock(), and moreover that VirtIoMSInterruptRoutine() can be called for other vectors when you use StorPortAcquireMSISpinLock(). Also virtqueue_disable_cb() is per virtqueue, so should only block for the vector currently under lock.

Yes, I'm always talking only about the same vector here.

`VirtIoMSInterruptRoutine()` doesn't get called for the same vector while you hold an MSISpinLock for that message (either explicitly in `VioStorVQLock()` or automatically around `VirtIoMSInterruptRoutine()` with `InterruptSynchronizePerMessage`).

It also doesn't get called while you hold an `InterruptLock` (which may again be explicitly through the non-MSI path in `VioStorVQLock()`, or automatically around `VirtIoMSInterruptRoutine()` with `InterruptSynchronizeAll`, or automatically around `VirtIoInterrupt()`).

I don't see `DpcLock` documented as interacting with any of this, and it would also be unexpected to me if it did interact. So my understanding is that holding that one will still allow the interrupt handler to run, and only running the DPC will be affected. (If I'm not misreading, this actually means that in the case when `CompleteDPC()` returns `FALSE`, we're ignoring the `DpcLock` that the submission code may be holding.)

> It seems quite counter-intuitive the current way, yes?

Sure, my intuition says that we should use `StorPortAcquireMSISpinLock()` whenever we can and a global `InterruptLock` as a fallback when it's not available. But obviously someone (in fact @vrozenfe) wrote the code with an additional case using `DpcLock`, so I assume that my intuition might be off and they had reasons for it.

This is why I'm asking all of these questions, because while intuitively always using `StorPortAcquireMSISpinLock()` seems simpler, I also don't really understand why it should improve performance so much compared to `DpcLock`. We don't seem to have a good theory for this yet, and I like to fully understand changes that I support.

But given that you do see the performance improvements, we should probably just a PR that does just this one change and properly benchmark it.

benyamin-codez: Ok, so I think we (mostly) are on the same page.

iirc, I think the historical decisions were during initial MSI-X implementation away from `InterruptLock` and because `msix_one_vector` would introduce the ASSERT and BSOD when using `StorPortAcquireMSISpinLock()`.

I'm fairly certain the reason why `StorPortAcquireMSISpinLock()` performs better in SMP systems is because `VirtIoMSInterruptRoutine()` can run for other vectors if `StorPortAcquireSpinLock()` isn't holding a lock.

In any case, I think your proposal for a separate PR and further benchmarking is a good idea. As I previously mentioned, I only saw the performance increase along with other changes, so it would be best to separate them first to have confidence in the claim.

It would also be prudent to rebase on PRs #1493 and #1503 first methinks...

kevmw: > I'm fairly certain the reason why StorPortAcquireMSISpinLock() performs better in SMP systems is because VirtIoMSInterruptRoutine() can run for other vectors if StorPortAcquireSpinLock() isn't holding a lock.

When I said above that my understanding is that `DpcLock` doesn't interact with interrupts, this scenario is precisely what I had in mind. Yes, holding an `InterruptLock` would definitiely prevent `VirtIoMSInterruptRoutine()` from being called for another message. But holding a `DpcLock` shouldn't - and it shouldn't even prevent the DPC for another message to run because the `DpcLock` is not global, but per DPC (and therefore per message in our design).

> It would also be prudent to rebase on PRs https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1493 and https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1503 first methinks...

I agree.

benyamin-codez: Yep, I get what you're saying. That being said, it's somewhat hard to test whether `VirtIoMSInterruptRoutine()` is blocked (as in Storport just doesn't call it) when using `DpcLock` type (we both agree `InterrupLock` type is a problem, but in this case we don't use MSI-X anyway), without stepping through in the debugger and in this instance that would take some time - perhaps too long (before timeouts occur). Testing for this isn't especially clear via tracing either.

Performance testing should reveal something, but in the end you could well be right about `DpcLock` type and any improvement could be due to some other reason.

viostor/virtio_stor.c
XanClic: I mean, as long as it’s fixed, it doesn’t need to be part of the list :)

But it would be separate issue from issue B, right.

benyamin-codez: Sure, it's the boundary check itself rather than the unit mismatch.

My thoughts were that if this part is being refactored to address the unit mismatch, then maybe this off-by-one issue could be addressed at the same time.

Refactoring might also consolidate the two tests to:
```c
 if (lba >= adaptExt->lastLBA || (lba + blocks) >= adaptExt->lastLBA)
```
Or even eliminate the first test...
...because `lba` + xfer_length (`blocks`) will always be > `lastLBA` if `lba` > `lastLBA`...

kevmw: `lastLBA` is actually the capacity (i.e. first LBA after the end of the disk), so it's a bit of a misnomer. Should it be renamed?

The current condition seems right to me (apart from the unit confusion): `lba + blocks` is the first LBA after the end of the request. It is allowed for this to be the first LBA after the end of the disk (`lba + blocks == adaptExt->lastLBA`), this happens when the request accesses the very end of the disk. Am I missing something?

What you could additionally take into account is integer overflows, if Windows doesn't already catch them before calling us. `if (lba >= adaptExt->lastLBA || blocks > adaptExt->lastLBA - lba)` should take care of that.

benyamin-codez: > `lastLBA` is actually the capacity (i.e. first LBA after the end of the disk), so it's a bit of a misnomer. Should it be renamed?

Maybe `totalLBAs`...?
iinm, `blk_config.capacity` would be a count...

> The current condition seems right to me (apart from the unit confusion): `lba + blocks` is the first LBA after the end of the request. It is allowed for this to be the first LBA after the end of the disk (`lba + blocks == adaptExt->lastLBA`), this happens when the request accesses the very end of the disk. Am I missing something?

Likely not.
That makes sense.
This is probably why the comment by @EnergyFaith went unanswered...

> What you could additionally take into account is integer overflows, if Windows doesn't already catch them before calling us. `if (lba >= adaptExt->lastLBA || blocks > adaptExt->lastLBA - lba)` should take care of that.

Looks good to me.

If someone is addressing the unit mismatch, can they pick up this consolidation along with it?

PR #1502, opened 2026-01-26
[viostor] Use ULONG for num_queues

benyamin-codez:
Refactors num_queues to use ULONG rather than USHORT.

The adapter extension struct contains a `num_queues` member of type `USHORT`. Here we refactor it to use `ULONG` as all mathematical operations are with other `ULONG` variables and this removes the need to cast as `ULONG` / `USHORT`. This change prepares for future PRs, which increase the number of these mathematical operations.

Changing to `ULONG` (rather than using `USHORT` for others) also helps with memory alignment with associated structures.

**NOTE:** The `blk_config` struct contains a `num_queues` member which remains 16 bits in length per the VirtIO v1.1 standard. When reading this value we ensure we only read the first 16 bits by using new `USHORT` variable `num_queues_u16` and then casting this to the corresponding `ADAPTER_EXTENSION` member. This method also preserves bi-endian system support, should it ever be needed.

We also refactor a for loop in `VirtIoStartIo()` that we touched when changing `num_queues` to use `ULONG`. The loop used index `i` which we refactor to the semantically mnemonic `vq_idx` instead.

The for loop is used in debug only. Future use may include ETW too.

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -1066,11 +1066,11 @@ VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
                              adaptExt->srb_cnt,
                              adaptExt->inqueue_cnt,
                              status ? "SRB_STATUS_SUCCESS" : "SRB_STATUS_ERROR");
-                for (USHORT i = 0; i < adaptExt->num_queues; i++)
+                for (ULONG vq_idx = 0; vq_idx < adaptExt->num_queues; vq_idx++)
                 {
-                    if (adaptExt->vq[i])
+                    if (adaptExt->vq[vq_idx])
                     {
-                        RhelDbgPrint(TRACE_LEVEL_ERROR, "%d indx %d\n", i, adaptExt->vq[i]->index);
+                        RhelDbgPrint(TRACE_LEVEL_ERROR, "%d indx %d\n", vq_idx, adaptExt->vq[vq_idx]->index);
                     }
                 }
 #endif


YanVugenfirer:

benyamin-codez:
Should have mentioned that we use ULONG in `vioscsi` too...

benyamin-codez:
Missed a definition...

viostor/virtio_stor.c
XanClic: As far as I understand, the length passed here must match the config space attribute size. `blk_config.num_queues` is 16-bit, so this read should remain 16-bit, too, or we’ll read the low 16 bits of `max_discard_sectors` into the high 16 bits of `num_queues`.

viostor/virtio_stor.c
```diff
@@ -981,11 +981,11 @@ VirtIoStartIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
                              SRB_FUNCTION(Srb),
                              adaptExt->srb_cnt,
                              adaptExt->inqueue_cnt);
-                for (ULONG i = 0; i < adaptExt->num_queues; i++)
+                for (ULONG vq_req_idx = VIRTIO_BLK_REQUEST_QUEUE_0; vq_req_idx < adaptExt->num_queues; vq_req_idx++)
```
XanClic: I assume this PR is supposed to go on top of #1499, where `VIRTIO_BLK_REQUEST_QUEUE_0` is defined?

Bike-shedding: `vq_req_idx` to me sounds like a request index inside of a specific virtqueue. Maybe instead just `vq_idx`, or `req_vq_idx`?

viostor/virtio_stor.c
benyamin-codez: Ouch...!
Of course it is...!
Thanks for spotting that one.
8^{d>~

viostor/virtio_stor.c
benyamin-codez: It was supposed to be under #1499, but I missed the constant.

I was dragging `vq_req_idx` from `vioscsi` with a view to creating a common code base down the track.
So, yes, your understanding of the mnemonic was correct.

I hadn't considered `req_vq_idx`. That could work well...
Still thinking some of this over whilst the other work progresses...

benyamin-codez: In considering the RHS of the statement, `vq_idx` is likely sufficient:
```c
ULONG vq_idx = VIRTIO_BLK_REQUEST_QUEUE_0;
```

It matters less here in `viostor`, but in `vioscsi` the queues are defined with:
```c
#define VIRTIO_SCSI_CONTROL_QUEUE            0
#define VIRTIO_SCSI_EVENTS_QUEUE             1
#define VIRTIO_SCSI_REQUEST_QUEUE_0          2
```

As a `vioscsi` example we have for queue allocation (noting the arithmetic is probably wrong):
```c
for (index = VIRTIO_SCSI_CONTROL_QUEUE; index < max_queues + VIRTIO_SCSI_REQUEST_QUEUE_0; ++index)
{...}
```
In this intance, it would also make sense to use `vq_idx` and let the mnemonic context work itself out by way of the macro constant:
```c
for (vq_idx = VIRTIO_SCSI_CONTROL_QUEUE; vq_idx < max_queues + VIRTIO_SCSI_REQUEST_QUEUE_0; ++vq_idx)
{...}
```

Yan - and I think Kevin - have suggested the same or similar elsewhere, so I think I will go with the consensus.

cc: @YanVugenfirer @kevmw

viostor/virtio_stor.c
benyamin-codez: It looks like I missed it in tracing because of the subsequent:
```c
adaptExt->num_queues = min(adaptExt->num_queues, num_cpus);
```

I ended up using:
```c
virtio_get_config(&adaptExt->vdev,
                  FIELD_OFFSET(blk_config, num_queues),
                  &adaptExt->num_queues,
                  sizeof(adaptExt->info.num_queues));
```

viostor/virtio_stor.c
kevmw: Yes, virtio-scsi is different, though the loop you quoted still looks more confusing than it probably needs to be.

In virtio-blk, I don't think `VIRTIO_BLK_REQUEST_QUEUE_0` has any use. I know you envision some code being shared between both drivers in the future, but then introducing that shared code is when the different queue numbers need to be taken into account. While the drivers are fully separate, it only adds unnecessary complexity. Code that we add should make sense in the codebase as it is today, not only in some speculative future state.

viostor/virtio_stor.c
XanClic: That will only work on little endian systems, no?

I think it would be better to read it into a local `u16` variable and assign `adaptExt->num_queues` from that.

benyamin-codez: Sure.

I didn't realise we were catering for bi-endian systems in this project. In MS development, we usually don't bother  even for modern ARM systems, and just use the native byte ordering, which is invariably little-endian. We really only use big-endian for specific data structures, e.g. Network Byte Order, GUIDs, etc.

If we should be catering for system endianness, we should probably include that in the dev wiki somewhere...

cc: @YanVugenfirer @kostyanf14

viostor/virtio_stor.c
benyamin-codez: I am inclined to agree tbh. I've done some further work in a WIP a few phases ahead of master on consolidation and the easiest mechanism to make use of the common code really requires these parts to be generic, which makes a lot of sense in hindsight... 8^{d>~

Thanks for the feedback here and elsewhere everyone.

So I'll revert to using:
```c
for (ULONG vq_idx = 0; vq_idx < adaptExt->num_queues; vq_idx++)
```
...and push an update.

YanVugenfirer:
ok to test

benyamin-codez:
@kostyanf14 Should we pause the builds here and rebase on PR #1493 before continuing...?

kostyanf14:
@benyamin-codez You can just force push, and Jenkins should (I hope it even does) terminate the old job

benyamin-codez:
@kostyanf14 - Did I miss the boat...? Can I still force push while the HCK runs are underway?

benyamin-codez:
> @kostyanf14 - Did I miss the boat...? Can I still force push while the HCK runs are underway?

I will try... :crossed_fingers:
Apologies in advance if it borks... :pray:

benyamin-codez:
@kostyanf14

Looks like Jenkins coped ok, but there aren't any runners to do the clang-format checks.
I will retry them a bit later...

benyamin-codez:
@kostyanf14

GitHub reported that the issue with Actions was resolved at 23:50 UTC. 
I reran the failed jobs in my fork but no perms here. I'll leave it to you. Let me know if you need another push.

kostyanf14:
> @kostyanf14 - Did I miss the boat...? Can I still force push while the HCK runs are underway?

You can

PR #1503, opened 2026-01-27
RHEL-149878: viostor: Cleanups around msix_one_vector and MessageId

kevmw:
As previous discussions have shown, the way msix_one_vector works (and especially that it uses MessageId = 1 in the code while it actually uses vector 0) is quite confusing. In fact, it almost looks like it works only accidentally, not by design. This PR cleans up the code to make it more obvious how things work. A functional change is not intended.

There is a clear next step after this: Add support for mapping multiple queues to a single vector (primarily to `VioStorCompleteRequest()`) and update `VirtIoHwInitialize()` to make use of all available vectors (and give up the config vector first) instead of immediately falling back to a single queue mode if they don't match. However, as this is essentially feature work, I decided to keep it separate from the cleanups. If you'd prefer to have a single PR for the full thing, let me know.

This is a draft because it's based on the current, non-final version of #1493

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -703,13 +703,14 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
     {
         /* initialize queues with a MSI vector per queue */
         RhelDbgPrint(TRACE_LEVEL_INFORMATION, (" Using a unique MSI vector per queue\n"));
-        adaptExt->msix_one_vector = FALSE;
+        adaptExt->msix_has_config_vector = TRUE;
     }
     else
     {
         /* if we don't have enough vectors, use one for all queues */
+        RhelDbgPrint(TRACE_LEVEL_INFORMATION, (" Disabling MSI config vector\n"));
         RhelDbgPrint(TRACE_LEVEL_INFORMATION, (" Using one MSI vector for all queues\n"));
-        adaptExt->msix_one_vector = TRUE;
+        adaptExt->msix_has_config_vector = FALSE;
         /* TODO Allow using multiple queues even with a single vector */
         adaptExt->num_queues = 1;
     }
@@ -764,7 +765,7 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
                 if (CHECKFLAG(perfData.Flags, STOR_PERF_INTERRUPT_MESSAGE_RANGES))
                 {
                     adaptExt->perfFlags |= STOR_PERF_INTERRUPT_MESSAGE_RANGES;
-                    perfData.FirstRedirectionMessageNumber = adaptExt->msix_one_vector ? 0 : 1;
+                    perfData.FirstRedirectionMessageNumber = adaptExt->msix_has_config_vector ? 1 : 0;
                     perfData.LastRedirectionMessageNumber = perfData.FirstRedirectionMessageNumber +
                                                             adaptExt->num_queues - 1;
                     ASSERT(perfData.lastRedirectionMessageNumber < adaptExt->num_affinity);
@@ -1253,9 +1254,9 @@ VirtIoInterrupt(IN PVOID DeviceExtension)
     intReason = virtio_read_isr_status(&adaptExt->vdev);
     if (intReason == 1 || adaptExt->dump_mode)
     {
-        if (!CompleteDPC(DeviceExtension, !adaptExt->msix_one_vector))
+        if (!CompleteDPC(DeviceExtension, adaptExt->msix_has_config_vector))
         {
-            VioStorCompleteRequest(DeviceExtension, !adaptExt->msix_one_vector, TRUE);
+            VioStorCompleteRequest(DeviceExtension, adaptExt->msix_has_config_vector, TRUE);
         }
         isInterruptServiced = TRUE;
     }
@@ -1578,7 +1579,7 @@ VirtIoMSInterruptRoutine(IN PVOID DeviceExtension, IN ULONG MessageID)
         return FALSE;
     }
 
-    if (!adaptExt->msix_one_vector && MessageID == VIRTIO_BLK_MSIX_CONFIG_VECTOR)
+    if (adaptExt->msix_has_config_vector && MessageID == VIRTIO_BLK_MSIX_CONFIG_VECTOR)
     {
         RhelGetDiskGeometry(DeviceExtension);
         adaptExt->sense_info.senseKey = SCSI_SENSE_UNIT_ATTENTION;
@@ -2173,7 +2174,7 @@ VOID VioStorCompleteRequest(IN PVOID DeviceExtension, IN ULONG MessageID, IN BOO
 {
     unsigned int len = 0;
     PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
-    ULONG QueueNumber = adaptExt->msix_one_vector ? MessageID : MessageID - 1;
+    ULONG QueueNumber = MessageID - adaptExt->msix_has_config_vector;
     STOR_LOCK_HANDLE queueLock = {0};
     struct virtqueue *vq = NULL;
     ULONG_PTR srbId = 0;
--- a/viostor/virtio_stor.h
+++ b/viostor/virtio_stor.h
@@ -231,7 +231,8 @@ typedef struct _ADAPTER_EXTENSION
     BOOLEAN dump_mode;
     ULONG msix_vectors;
     BOOLEAN msix_enabled;
-    BOOLEAN msix_one_vector;
+    /* Standard C type because we do arithmetics with it */
+    _Bool msix_has_config_vector;
     ULONGLONG features;
     CHAR sn[BLOCK_SERIAL_STRLEN];
     BOOLEAN sn_ok;
--- a/viostor/virtio_stor_hw_helper.h
+++ b/viostor/virtio_stor_hw_helper.h
@@ -73,7 +73,7 @@ static inline ULONG QueueToMessageId(PVOID DeviceExtension, ULONG QueueNumber)
 {
     PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
 
-    if (!adaptExt->msix_one_vector)
+    if (adaptExt->msix_has_config_vector)
     {
         QueueNumber++;
     }
@@ -90,7 +90,7 @@ static inline ULONG MessageToDpcIdx(PVOID DeviceExtension, ULONG MessageId)
 {
     PADAPTER_EXTENSION adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
 
-    if (!adaptExt->msix_one_vector)
+    if (adaptExt->msix_has_config_vector)
     {
         /* The config vector doesn't have a slot in adaptExt->dpc */
         MessageId--;
--- a/viostor/virtio_pci.c
+++ b/viostor/virtio_pci.c
@@ -231,20 +231,13 @@ static u16 vdev_get_msix_vector(void *context, int queue)
         /* queue interrupt */
         if (adaptExt->msix_enabled)
         {
-            if (adaptExt->msix_one_vector)
-            {
-                vector = 0;
-            }
-            else
-            {
-                vector = queue + 1;
-            }
+            vector = adaptExt->msix_has_config_vector ? queue + 1 : queue;
         }
     }
     else
     {
         /* on-device-config-change interrupt */
-        if (!adaptExt->msix_one_vector)
+        if (adaptExt->msix_has_config_vector)
         {
             vector = VIRTIO_BLK_MSIX_CONFIG_VECTOR;
         }


YanVugenfirer:

viostor/virtio_stor.c
```diff
@@ -839,54 +844,119 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
...
+     * are currently in progress in DPC. Each new DPC call will have no affect because
+     * VioStorCompleteRequest will exits for any non-zero reset_in_progress_count value
+     */
+    for (ULONG QueueNumber = 0; QueueNumber < adaptExt->num_queues; QueueNumber++)
+    {
+        ULONG MessageId = QueueToMessageId(DeviceExtension, QueueNumber);
+        STOR_LOCK_HANDLE LockHandle = {0};
+        VioStorVQLock(DeviceExtension, MessageId, &LockHandle, FALSE);
+        adaptExt->reset_in_progress_count++;
+        VioStorVQUnlock(DeviceExtension, MessageId, &LockHandle, FALSE);
```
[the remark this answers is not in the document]
kevmw: This refers only to the base PR and is invalid anyway.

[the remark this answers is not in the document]
kostyanf14: unfortunatly Gemini is incorrect, using one global lock for DPC will cause performance regression

benyamin-codez:
I like the approach here, Kevin.
It got past me, so only just noticed it now.
Maybe it is indeed best for me to rebase some pickings of mine on top of this.

Could you consider my `VioStorVQLock()`/`VioStorVQUnlock()` routines in PR #1499...?
We generally use `InterruptSynchronizePerMessage` so should prefer `StorPort*MSISpinLock()`...

https://github.com/virtio-win/kvm-guest-drivers-windows/blob/7229eaba0e524fe2866093340db35f9b4d97b259/viostor/virtio_stor_hw_helper.c#L678-L747

Anyway, I'm off to the land of nod...

kevmw:
@benyamin-codez I don't think your changes are in conflict with mine, except that the `QueueNumber` part should just be a `MessageToDpcIdx()` call now. They are also unrelated to my cleanups, so I think it's best to have them in a separate PR. But let's discuss the details in #1499.

viostor/virtio_stor.c
```diff
@@ -760,7 +765,7 @@ VirtIoHwInitialize(IN PVOID DeviceExtension)
                 if (CHECKFLAG(perfData.Flags, STOR_PERF_INTERRUPT_MESSAGE_RANGES))
                 {
                     adaptExt->perfFlags |= STOR_PERF_INTERRUPT_MESSAGE_RANGES;
-                    perfData.FirstRedirectionMessageNumber = 1;
+                    perfData.FirstRedirectionMessageNumber = adaptExt->msix_config_vector ? 1 : 0;
```
benyamin-codez: I don't think we hit this when `msix_config_vector = FALSE` as then `num_queues = 1` and we don't need Storport performance assistance.

viostor/virtio_stor.c
benyamin-codez, line 744: This is probably best to be:
```c
if ((adaptExt->msix_vectors > 1) && (adaptExt->perfFlags == 0))
```
btw, I'm thinking it's best I rebase PR #1300 and others on top of this PR.

viostor/virtio_stor.c
```diff
@@ -1493,22 +1578,15 @@ VirtIoMSInterruptRoutine(IN PVOID DeviceExtension, IN ULONG MessageID)
...
-            return TRUE;
-        }
+        RhelGetDiskGeometry(DeviceExtension);
+        adaptExt->sense_info.senseKey = SCSI_SENSE_UNIT_ATTENTION;
+        adaptExt->sense_info.additionalSenseCode = SCSI_ADSENSE_PARAMETERS_CHANGED;
+        adaptExt->sense_info.additionalSenseCodeQualifier = SCSI_SENSEQ_CAPACITY_DATA_CHANGED;
+        adaptExt->check_condition = TRUE;
+        DeviceChangeNotification(DeviceExtension, TRUE);
+        return TRUE;
     }
```
benyamin-codez, line 1591: I've been very curious how we do this when `msix_config_vector = FALSE`.
Is it just not possible? A limitation of running with only a single vector?

kevmw:
I pushed a rebased version now that #1493 has been merged and it's no longer just a draft.

viostor/virtio_stor.c
kevmw: I didn't explicitly test it, but I think it will just fall back to the non-MSI codepath in `VirtIoInterrupt()`.

viostor/virtio_stor.c
kevmw: Maybe it could even be unconditional, or at least lose the first condition. I don't think setting the performance options would hurt even with a single queue or vector. But it's only tangentially related to this PR, so it should probably be a separate cleanup.

viostor/virtio_stor.c
kevmw: Good point, after this patch, that's probably currently true.

The next step I mentioned in the PR description would be to allow multiple queues while `msix_config_vector == FALSE` instead of immediately falling back to single queue, and then the code would become active again.

viostor/virtio_stor.c
benyamin-codez: Ok. I'll do some explicit testing to confirm.

viostor/virtio_stor.c
benyamin-codez: Right, that makes more sense now. Good idea really...
I'm working on the `num_queues` <--> `msix_vectors` logic now.
So I'll be mindful of your proposed change.

viostor/virtio_stor.c
benyamin-codez: Probably best as:
```c
if ((adaptExt->msix_enabled > 1) && (adaptExt->perfFlags == 0))
```
...to avoid in the corner case when there's no MSI-X.

I'll add it to mine.

viostor/virtio_stor.c
benyamin-codez: This appears broken for me, even on master - in `vioscsi` too (nothing hits the events queue)...

When trying to do a size change, I don't get an interrupt at all, definitely not on MessageId 0 anyway. Disk refresh and rescan in `diskmgmt.msc` don't produce any changes. The only way I can hit `DeviceChangeNotification()` is via the `VirtIoStartIo()` pathway on surprise removal. On reattaching the disk, the capacity changes.

This is on Proxmox at least... I noticed the same thing on a FreeBSD guest too in the last 12 hours.
```bash
# pveversion -v | grep qemu | grep -v backup
pve-qemu-kvm: 10.1.2-5
qemu-server: 9.1.3

# kvm --version
QEMU emulator version 10.1.2 (pve-qemu-kvm_10.1.2-5)

# qm showcmd <VMID> --pretty | grep machine
  -machine 'hpet=off,type=pc-q35-10.1+pve0' \
```
Maybe this is a QEMU bug. Any chance you could please check on a RH flavour?

benyamin-codez: @kevmw - Just to be clear, this was with the config vector enabled.

viostor/virtio_stor.h
```diff
@@ -231,7 +231,7 @@ typedef struct _ADAPTER_EXTENSION
     BOOLEAN dump_mode;
     ULONG msix_vectors;
     BOOLEAN msix_enabled;
-    BOOLEAN msix_one_vector;
+    BOOLEAN msix_config_vector;
```
YanVugenfirer: +1 for PR in general.
I have a question about this semantic change.

What are the cases when we have only one vector? Is this because hypervisor/HW don't have configuration interrupt or we are also covering for the case when there could be a configuration interrupt but for some reason OS allocated only one vector to the device? In this is a second case - the name is confusing. 

https://learn.microsoft.com/en-gb/windows-hardware/drivers/kernel/introduction-to-message-signaled-interrupts
_"Drivers that are designed to use multiple messages should check that the expected number of messages is allocated. If the Plug and Play (PnP) manager cannot allocate the requested number of messages, it instead allocates exactly one message to the device. Drivers can check the number of messages that are actually allocated in one of the following ways:"_

benyamin-codez: I've refactored this as `ULONG msix_cfg_vector_cnt;` and am using it in calculations...

YanVugenfirer: But "count" also doesn't make sense. There is only one config interrupt according to the spec.

viostor/virtio_stor.c
YanVugenfirer: If config vector is enabled and MSIx is used and only one vector is available - this code will be broken.

viostor/virtio_stor.h
benyamin-codez: When count = 0 there is no config vector; when count = 1 there is a config vector.
It really doesn't matter that there will never be 2, 3, 4, etc.
It does permit arithmetic operations rather than logical operations, which simplifies the code.

In either case, I think it is ok to be PnP agnostic here. We know that if the table cannot be allocated with the requested size, we will end up with only one vector, and thus no MSI-X vector for _**exclusive**_ configuration change notifications. When we have any more than one vector, we will have one. This also keeps it VM and hypervisor agnostic. I think it is a pretty neat solution that covers all edge cases.

I should be able to force push PR #1300 shortly so you can have a look.
It works quite well, just updating my commit messages.

kevmw: So to come back to Yan's original question, in the case of virtio-blk, the "hardware" just provides a number of queues and a number of MSI-X vectors, both independently configurable on the QEMU command line. There is no inherent mapping between queues and vectors, the mapping is configured at runtime by the driver (see `vdev_get_msix_vector()`). In other words, the decision whether a config vector should be used is the decision of the driver. I think it's the lowest priority one, so if we don't have at least `num_queues + 1` vectors, the config vector is the first thing we should switch to non-MSI.

The code so far uses `msix_one_vector` for all cases in which we don't have enough vectors to both have a config vector and give every queue a separate vector. This has two effects: The first is that config changes switch to non-MSI (as suggested above), but it also means that we effectively fall back to single queue in this case, even if we have multiple queues and vectors available (justifying the old name).

The primary idea behind the switch to `msix_config_vector` is that instead, we could just disable MSI-X for the config vector, but still make use all of the remaining vectors for queues. If you have 8 queues and 4 vectors, you want to make use of at least 4 queues instead of only a single queue and leaving three vectors unused. Implementing this would be the next step after this PR.

But even without this, all the places that check `msix_one_vector` only care about the presence of the config vector (because this changes the index of the vectors used for queues) and not the fact that only a single queue is used (which may still be true with `msix_one_vector == TRUE` and `num_queues == 1`). So it's the more accurate name even now.

viostor/virtio_stor.c
kevmw: @YanVugenfirer I don't think this is a setup we'll ever want to use, but if we did set `msix_config_vector = TRUE` while having only a single vector, then what would need to happen is that the queue (there would be only one in use) uses non-MSI interrupts. There is a problem in `vdev_get_msix_vector()` with this (it should check that the vectors it tries to assign even exist and return `VIRTIO_MSI_NO_VECTOR` otherwise), but I think this code in `VirtIoMSInterruptRoutine()` is fine.

Not sure if making the scenario work is worth it, though. We'll never configure things like this.

viostor/virtio_stor.h
YanVugenfirer: @kevmw  we cannot decide in the driver that we are not using MSI interrupts (or at least this is a usual case in Windows, might be different in storport mini port). 
Meaning if we get only one MSI interrupt vector, we cannot switch to legacy.

`    hwInitData.HwInterrupt = VirtIoInterrupt; ` - is set during DriverEntry
` ConfigInfo->HwMSInterruptRoutine = VirtIoMSInterruptRoutine;` - is set in VirtIoFindAdapter. At this stage pci bus driver did all the PCI resource setup for the device and if the device has MSI vectors this callback should be set.

We can fail VirtIoFindAdapter with SP_RETURN_BAD_CONFIG error code. But that will mean the driver will not be functional is such scenario.

kevmw: You're right, I assumed that mixing interrupt types is possible, but it doesn't seem to work like this. In this case, the existing `msix_one_vector` code is already broken and config changes will never be honoured. I think this means that without a separate config vector, config updates need to share a vector with a queue and we need to check the ISR register even in `VirtIoMSInterruptRoutine()` like it's done in `VirtIoInterrupt()`.

We'll also need to check why it is that Ben didn't see config updates working at all, even if enough vectors were available. So we can probably expect some additional fix in this area either way.

benyamin-codez: It's really set after HW_INITIALIZE, i.e. `VirtIoHwInitialize()` when we set the performance options. Then Storport knows to use MSI-X instead. I've never seen mixed types before, but there might be a magic config... In `vioscsi` we use the events queue, but in my test, that wasn't even receiving anything. We probably need to fix that first so we can do testing...

viostor/virtio_stor.c
kevmw: @benyamin-codez It works for me with a config vector. I used `block_resize` in the HMP monitor and the new size was visible in the Disk Management program after either going to the properties dialog for the disk or pressing F5. Without a config vector, the notification goes nowhere and the size doesn't update.

I tested this on an upstream qemu.git close to 10.2.0-rc3.

benyamin-codez: Looking at the relevant part of the [VirtIO standard (s.4.1.5.4)](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-1760004), if we set `VIRTIO_MSI_NO_VECTOR` then _"the device <ins>**MUST NOT**</ins> deliver an interrupt for device configuration space changes (s.4.1.5.4.1)"_, and if we do not _"the driver <ins>**MUST**</ins> handle the case where the same interrupt is used to indicate both device configuration space change and one or more virtqueues being used. (s.4.1.5.4.2)"_

However, how do we do the latter in an inexpensive manner when we only have one vector? Surely we cannot do a full check for configuration changes with every message... The ISR Status register is still present in the VirtIO common configuration structure and it's probably worth checking to see if it gets updated anyway. However,  presently it doesn't get updated in normal operation with legacy interrupts. Similarly, the config vector never gets fired when MSI-X is enabled.

> Not sure if making the scenario work is worth it, though. We'll never configure things like this.

If PnP cannot - for whatever reason - supply the number of vectors required - we will end up with one vector.

If we cannot do the above, we should revert to legacy interrupt instead in `VirtIoHwInitialize()`, as we cannot set `VIRTIO_MSI_NO_VECTOR` on the single vector without disabling the queue vector too (per s.4.1.5.4.2), and we cannot handle the case where the interrupt services both config and queue roles (per s.4.1.5.4.2).

First we need to test the latter case, and to work that out, first we must fix the lack of notifications...

benyamin-codez: > It works for me with a config vector. I used `block_resize` in the HMP monitor and the new size was visible in the Disk Management program after either going to the properties dialog for the disk or pressing F5. Without a config vector, the notification goes nowhere and the size doesn't update.
> 
> I tested this on an upstream qemu.git close to 10.2.0-rc3.

Apologies Kevin. Didn't see this come in. I'll have a try with that method...

YanVugenfirer: @benyamin-codez In my opinion having one MSI vector is problematic configuration to begin with and we can live with performance hit in such case.

benyamin-codez: @kevmw @YanVugenfirer 

Thanks for the hint Kevin.
I couldn't get HMP to work but QMP did.
It appears `qemu-img resize ...` + `qm rescan` does not.

Yan, I think that would have been a reasonable fallback, i.e. checking with every interrupt.

As it stands, hitting the ISR Status register seems to work just fine. Performance is more than acceptable.

Here is a trace snap of the change notification coming through:

chk_isr_status_for_change

...
Should I write that into PR #1300, a separate PR, or did you maybe want to put in this one (_sans_ Dbg prints)?

```c
BOOLEAN
VirtIoMSInterruptRoutine(IN PVOID DeviceExtension, IN ULONG MessageID)
{
    ULONG intReason;

    if (MessageID > adaptExt->num_queues || adaptExt->removed == TRUE || adaptExt->stopped == TRUE)
    {
        RhelDbgPrint(TRACE_LEVEL_ERROR, " MessageID = %d\n", MessageID);
        return FALSE;
    }

    if (adaptExt->msix_cfg_vector_cnt && MessageID == VIRTIO_STOR_MSIX_CONFIG_VECTOR)
    {
        RhelGetDiskGeometry(DeviceExtension);
        adaptExt->sense_info.senseKey = SCSI_SENSE_UNIT_ATTENTION;
        adaptExt->sense_info.additionalSenseCode = SCSI_ADSENSE_PARAMETERS_CHANGED;
        adaptExt->sense_info.additionalSenseCodeQualifier = SCSI_SENSEQ_CAPACITY_DATA_CHANGED;
        adaptExt->check_condition = TRUE;
        DeviceChangeNotification(DeviceExtension, TRUE);
        return TRUE;
    }
    
    if (!adaptExt->msix_cfg_vector_cnt)
    {
        intReason = virtio_read_isr_status(&adaptExt->vdev);
        RhelDbgPrint(TRACE_LEVEL_FATAL, "---------------------------------------------------- intReason : %d \n", intReason);
        if (intReason == 3)
        {
            RhelDbgPrint(TRACE_LEVEL_FATAL, "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! intReason : %d \n", intReason);
            RhelGetDiskGeometry(DeviceExtension);
            adaptExt->sense_info.senseKey = SCSI_SENSE_UNIT_ATTENTION;
            adaptExt->sense_info.additionalSenseCode = SCSI_ADSENSE_PARAMETERS_CHANGED;
            adaptExt->sense_info.additionalSenseCodeQualifier = SCSI_SENSEQ_CAPACITY_DATA_CHANGED;
            adaptExt->check_condition = TRUE;
            DeviceChangeNotification(DeviceExtension, TRUE);
            return TRUE;
        }
    }


    if (!CompleteDPC(DeviceExtension, MessageID))
    {
        VioStorCompleteRequest(DeviceExtension, MessageID, TRUE);
    }

    return TRUE;
}
```

benyamin-codez: @YanVugenfirer @kevmw 

I started a draft... Here's what I have so far: 90c39c6

benyamin-codez: Next cut: 816cf1a  Works well.

Just need to split for review, and decide whether to rebase on this PR, or 1300 or master...

Any preference?

viostor/virtio_stor_hw_helper.h
```diff
@@ -69,6 +69,39 @@ FORCEINLINE VOID SrbGetPnpInfo(_In_ PVOID Srb, ULONG *PnPFlags, ULONG *PnPAction
     }
 }
 
+static inline QueueToMessageId(PVOID DeviceExtension, ULONG QueueNumber)
```
benyamin-codez: The bot pulled me up on these when rebasing on this PR to use an explicit return type:
```c
static inline ULONG QueueToMessageId(PVOID DeviceExtension, ULONG QueueNumber)
```

I note we typically use FORCEINLINE rather than letting the compiler decide whether to inline or use a static copy for each use. Perhaps a TODO note for future performance profiling work...?

viostor/virtio_stor_hw_helper.h
```diff
@@ -69,6 +69,39 @@ FORCEINLINE VOID SrbGetPnpInfo(_In_ PVOID Srb, ULONG *PnPFlags, ULONG *PnPAction
...
+    else
+    {
+        NT_ASSERT(QueueNumber == 0);
+    }
+
+    /* TODO Limit to msix_vectors to allow multiple queues per vector */
+    return QueueNumber;
+}
+
+static inline MessageToDpcIdx(PVOID DeviceExtension, ULONG MessageId)
```
benyamin-codez: The bot pulled me up on these when rebasing on this PR to use an explicit return type:
```c
static inline ULONG MessageToDpcIdx(PVOID DeviceExtension, ULONG MessageId)
```
I note we typically use FORCEINLINE rather than letting the compiler decide whether to inline or use a static copy for each use. Perhaps a TODO note for future performance profiling work...?

viostor/virtio_stor.h
benyamin-codez: I note [the comment](https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1517#discussion_r2779132436) by @YanVugenfirer in PR #1517 to use `msi_one_vector_used` and seeing that part is just because I rebased on top of this PR, I thought it best to discuss here.

There's two issues I guess:

On semantics:
Yan, was dropping the "x" from `msi_*` intentional?
If not, I presume `msix_one_vector_used` is ok.

On function:
As mentioned above, I refactored this as a ULONG to use arithmetically in PR #1300.
In this PR, @kevmw makes use of some double bangs which work well with either BOOLEAN or integers and there were a couple of parts where integers simplified the code, e.g. in `vdev_get_msix_vector()` and the  `perfData.FirstRedirectionMessageNumber` definition. The parts were:

```diff
- BOOLEAN msix_config_vector;
+ ULONG msix_cfg_vector_cnt;
...
- vector = adaptExt->msix_config_vector ? queue + 1 : queue;
+ vector = (int)adaptExt->msix_cfg_vector_cnt + queue;
...
- perfData.FirstRedirectionMessageNumber = adaptExt->msix_config_vector ? 1 : 0;
+ perfData.FirstRedirectionMessageNumber = adaptExt->msix_cfg_vector_cnt;
```

As you can see, there is only a minor simplification. So the change may not be worth it to avoid confusion. Some guidance on whether I should leave this as a ULONG in my PR or revert this to use a BOOLEAN would be helpful. If ULONG, would `msix_cfg_vectors_used` be ok?

Cheers,
Ben

kevmw: If you don't like the double bangs, I think they are not strictly needed because at least at the moment we always explicitly set `TRUE` or `FALSE`. I only put them there to be extra defensive because in theory a `BOOLEAN` could have other values. Or we could just use standard C `bool` and avoid the whole mess of Windows boolean types (it wouldn't be the first place to use it in the repository).

I can also rename to `msix_has_cfg_vector` if that's clearer.

viostor/virtio_stor_hw_helper.h
kevmw: `static inline` is just kind of the default in header files, I didn't mean to imply any point on performance. I would trust the compiler to do the right thing and don't expect that `FORCEINLINE` would make a major difference.

The missing return type is a good point, though. Will fix.

viostor/virtio_stor.h
YanVugenfirer: @kevmw @benyamin-codez 

- I think it is OK to use standard C bool
- Regarding the naming: I think msix_has_cfg_vector is clearer. "msix_cfg_vectors_used" - is more confusing, because we don't suppose to have more than one configuration vector anyway, and it takes us further away from explaining the case for this variable - only one vector is used for the configuration and the queue.

kevmw: Ugh, I saw only now that osdep.h has `#define bool u8`, which conflicts with the stdbool.h definition. So `_Bool` it is for a real standard C boolean that is guaranteed to be 0 or 1.

viostor/virtio_stor.c
kevmw: It seems like a lot of noise around the real change. But more importantly, does this really work? You still seem to use NO_VECTOR for config changes, so I don't think we'd ever get the config interrupt. Ah, or does it work only accidentally because the next queue interrupt might still see the old ISR bit and do the config update then?

This is the fix I have: https://github.com/kevmw/kvm-guest-drivers-windows/compare/msg-cleanup...notify-config (based on this PR, so I haven't created a PR yet - still not sure how to handle dependent PRs best in Github, it doesn't seem to be supported very well).

viostor/virtio_stor.h
benyamin-codez: I did't mind the double bangs...!!
They certainly have their place - especially with integer conversions to bool.
All good. It looks like we might as well run with a bool anyway.
I'm not sure those two simplifications I did above are worth the bool confusion they probably introduce.
I'll refactor my PRs to use `msix_has_cfg_vector` too and leave it as a bool.

I presume `_Bool` is implemented as 1 byte.
The standard never specified the size other than being large enough to store a bit.
It's usually implemented as a byte these days...
... but I'm minded to consider why we saw fit to previously implement `bool` as a ~CHAR~ UCHAR.
The standard MS `BOOL` is 4 bytes (`typedef int`)...
So... I assume the reason is for efficient byte boundary alignment in structs...
Or maybe just to align with `_Bool` as `bool` is just an alias for `_Bool` per `stdbool.h`...
As long as `_Bool` is 1 byte, it's probably fine.
Otherwise `BOOLEAN` (1-byte - `typedef BYTE`) or `bool` (1 byte) should be fine for arithmetic too.
iinm, all of these - including `_Bool` - are `u8` anyway....

Boolean types are a bit of a can of worms. Maybe we should exercise caution with changing it...?

kevmw: Standard C `_Bool` (and therefore `bool` with stdbool.h, but not with our osdep.h) is different from `u8` in that it really only allows 0 and 1 as values. You can assign `_Bool foo = 42;` and then print it and you get 1 as the result. This is how a proper boolean type should behave.

You're right that the legacy Windows types seem to be bit of a can of worms... `BOOLEAN` pretends to be boolean, but as you say, it's effectively a `u8` that can take any value, so doing arithmetics with that one is a bit risky because you never know if your true is 1 or 42 or something entirely different.

Alignment shouldn't be a problem, this is a purely internal type and not ABI. We could group booleans together to allow the compiler to keep the struct as compact as possible, but `ADAPTER_EXTENSION` is a huge struct anyway, so I wouldn't be overly concerned about a few bytes more or less.

benyamin-codez: Apologies Kevin, I meant arithmetic with `BOOLEAN` (or `bool` in our `osdep.h`) in the context of double bangs would have been fine. The `BOOLEAN` type was introduced by the OS/2 NT team and remains in security and driver subsystems, which is why we see it here. If testing against FALSE - even `var != FALSE` to test for TRUE - then it is also reliable. In most cases it works anyway, but if testing with `var == 1` then you might have trouble.

Whilst `ADAPTER_EXTENSION` isn't an issue, an auditor will likely see your inline comment and check every other `BOOLEAN` and `bool` for arithmetic and suggest a change to `_Bool`. This could potentially have unintended consequences for `#pragma` packed or otherwise properly aligned structures (I'm about to suggest this for many of the structs in `virtio_stor.h`). That being said, I'm fairly certain that `_Bool` is usually compiled as a `BYTE PTR` and the smallest size available in any case is 1 byte, so this is probably a moot point - provided `_Bool` packs well with `#pragma pack(1)`.

viostor/virtio_stor.c
benyamin-codez: @kevmw 

Yeah, mine works too because it still is `MessageId = 0`...
... and then I only skip config update if the ISR bit-field says it's for a queue vector.

In any case, I like your proposal:

 * Proper implementation of the ISR Status bit-field checks avoids potential `0x01b` \ `0x10b` problems.
 * The `u8` for `isr` in `VirtIoMSI...()` is also the right type (although we seem to prefer `UCHAR` here).
 * The new `VirtIoConfigUpdated()` routine you carved out cleans it up too.

If you want to run with this, I might do a relocation refactor (to group the ISR routines together) plus introduce an MSI-X ISR worker routine on top of your PR. I can also update `intReason` in `VirtIoInterrupt()` to use the UCHAR instead of ULONG - unless you want to do it here. Some standardisation on relevant local variable names would be good, e.g. I used `isr_status` for `intReason` / `isr` and then `irq_serviced` rather than `isInterruptServiced`.

Alternatively, I could introduce your improvements into mine and do it in one step.
Let me know what you prefer.

viostor/virtio_stor_hw_helper.h
benyamin-codez: The only problem with inlining is that if the routine does get compiled inline it introduces an issue with ETW if there are any trace messages. We use a standard prefix in the WPP that includes the function name and this will become - for all traces - the first function in which the routine was inlined. For some reason, it will also introduce a small latency overhead. I will  introduce a tracing PR soon that adds Inline functions to address this, but it cannot cater to routines that are dynamically inlined by the compiler.

So whilst that's not an issue here as there are no traces, it is something to be mindful of if any tracing is added.

viostor/virtio_stor.c
kevmw: I think I'd like to go with my notify-config branch, once this PR is merged.

For the additional changes, maybe let's ask @YanVugenfirer first, because I understood that he doesn't like commits just for renaming things etc. My own approach is to just clean up the code that I'm touching anyway. But if there is something small we agree on (like moving both interrupt handlers next to each other), I can include it in my branch, too. In this case, you could rebase and then I could cherry-pick from your branch so that you're still credited as the author, but we get everything done in a single PR.

As for `VirtIoMSInterruptWorker()`, I'm not sure what the purpose of that one is. To me it looks like an arbitrary split of `VirtIoMSInterruptRoutine()`, which wasn't even a very long function?

In general, instead of trying to make the code a bit prettier here and there, I feel that maybe we should focus more on the things that have practical effects e.g. on performance. The locking changes you mentioned the other day would be something that seems interesting to get separated into a commit and benchmarked. Maybe we should try and coordinate on Slack which things to address next?

viostor/virtio_stor.h
kevmw: Yes, that's why I had the double bangs initially.

Anyway, I don't think there's a reason why `_Bool` should not be a single byte, and a quick web search confirms that this is what MSVC does.

viostor/virtio_stor.c
YanVugenfirer: I am OK with commits that renames stuff, but let's make them part of something useful. I agree regarding the coordination first.

kevmw:
The only active conversations here don't seem to be about this PR itself any more, but more about possible future work on top of it.

Does this mean we should move ahead and merge this?

viostor/virtio_stor.c
benyamin-codez: I will rebase ASAP, and now this has merged, my other PRs too.

PR #1443, opened 2025-10-24
[vioscsi] vioscsi misses last target and last LUN #1442

hrosenfeld:
As mandated by the virtio spec, adjust the check in VioScsiBuildIo() to allow TargetId and Lun being equal to max_target and max_lun, respectively.

Should fix https://github.com/virtio-win/kvm-guest-drivers-windows/issues/1442, but as I don't have a way to rebuild the drivers for Windows I can't test this change. If someone else would rebuild the driver for me with this change, I can test them on bhyve on illumos and FreeBSD. Alternatively, I can help setting up an appropriate bhyve test environment.

PR #1511, opened 2026-01-31
[vioscsi] Fix off-by one errors in PORT_CONFIGURATION_INFORMATION members

### LINKED ISSUE #1442: vioscsi misses last target and last LUN
I'm currently working on porting the virtio-scsi device from FreeBSD bhyve to illumos bhyve, improving it in the process by adding support for multiple targets. During my tests with a Windows VM using virtio-win-0.1.285, I noticed that the target with the highest target id isn't detected.

I presume this is caused by this check right at the beginning of VioScsiBuildIo():

```
    if ((SRB_PATH_ID(Srb) > (UCHAR)adaptExt->num_queues) || (TargetId >= adaptExt->scsi_config.max_target) ||
        (Lun >= adaptExt->scsi_config.max_lun) || adaptExt->bRemoved)
    {
        SRB_SET_SRB_STATUS(Srb, SRB_STATUS_NO_DEVICE);
        SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
        StorPortNotification(RequestComplete, DeviceExtension, Srb);
        return FALSE;
    }
```

To quote from the virtio spec (https://docs.oasis-open.org/virtio/virtio/v1.2/csd01/virtio-v1.2-csd01.html#x1-3470004), section 5.6.4:

> max_channel, max_target and max_lun
>    can be used by the driver as hints to constrain scanning the logical units on the host to channel/target/logical unit numbers that are less than or equal to the value of the fields. max_channel SHOULD be zero. max_target SHOULD be less than or equal to 255. max_lun SHOULD be less than or equal to 16383.

In accordance with the spec, the virtio-scsi device of bhyve sets max_target to the highest target id configured. On all released versions of FreeBSD, this means that max_target is set to 0 as only one target with id=0 is supported, leading to a situation where the virtio-win vioscsi driver would not detect any target at all. With the multi-target support I'm currently developing for virtio-scsi for bhyve, all but the last targets are detected for the same reason.

Obviously, the check against max_target should be `>`, not `>=`, as `TargetID == max_target` is perfectly valid. The same is true for the check against max_lun.

A similar bug was present in the vioscsi driver in illumos, which I've fixed recently: https://www.illumos.org/issues/17609

benyamin-codez:
Fixes off-by one errors in `VioScsiFindAdapter()` for `PORT_CONFIGURATION_INFORMATION` members `MaximumNumberOfTargets` and `MaximumNumberOfLogicalUnits`, and introduces a similar method used by these to also define the `NumberOfBuses` member. Includes minor refactoring in separate commits.

**_Commit summary:_**

<ins>_Rename `TargetId` variable to `Target`_</ins>

Renames `TargetId` variable to `Target` in `VioScsiBuildIo()`.

`TargetId` is a member of the legacy `SCSI_REQUEST_BLOCK` struct, as was `PathId` and `Lun`. Since porting from the SCSI port driver to the Storport driver, we now use the `STORAGE_REQUEST_BLOCK` struct instead, which contains an `AddressOffset` member which is used to locate a `STOR_ADDRESS` struct. The `STOR_ADDRESS` struct
contains an `AddressData` array member containing a `STOR_ADDR_BTL8` struct, which in-turn contains `Path`, `Target` and `Lun` members. The term Path is interchangeable with the terms Bus and Channel. Here we update the use of the `TargetId` variable to use `Target` instead. `Lun` remains the same.

Per `Lun` variable and proposed `Channel` variable which do not use the Id suffix.

<ins>_Fix off-by-one error in `PORT_CONFIGURATION_INFORMATION` members_</ins> 

Fixes off-by one errors in the following `PORT_CONFIGURATION_INFORMATION` members:

 * `MaximumNumberOfTargets`
 * `MaximumNumberOfLogicalUnits`

These members are respectively derived from the `scsi_config.max_target` and `scsi_config.max_lun` `ADAPTER_EXTENSION` struct members via the `VirtIOSCSIConfig` struct, and represent the highest Target and Lun with a zero-based index rather than the number of Targets and Luns. Therefore, we need to add one to this value in order to properly populate the `PORT_CONFIGURATION_INFORMATION` `*NumberOf*` members in `VioScsiFindAdapter()`.

This commit also update the initial check in `VioScsiBuildIo()` to permit a maximum `Target` and `Lun` of zero (0) as can be reported by some hypervisors, e.g. bhyve. More specifically, the previous behaviour would not permit the last Target and Lun to proceed. Some hypervisors will report the installed maximum Target and Lun rather than the theoretical maximum. When the last Target and Lun are processed, e.g. T:L = 0:0 (the common case), the existing check will return early after setting the SRB status to `SRB_STATUS_NO_DEVICE`.

This commit also adds relevant comments.

Fixes #1442

Credit: @hrosenfeld - Hans Rosenfeld

<ins>_Populate `NumberOfBuses` with `scsi_config.max_channel` value_</ins>

Populates the `PORT_CONFIGURATION_INFORMATION` member `NumberOfBuses` found in `VioScsiFindAdapter()` with the value from `scsi_config.max_channel` + 1 as this is a zero-based index of the maximum channel rather than the maximum number of channels. This is the same behaviour as now used for Target and Lun.

Whilst Storport permits `SCSI_MAXIMUM_BUSES_PER_ADAPTER` (255) channels, we must limit this to `VIRTIO_SCSI_MAX_BUSES_PER_HBA` (1), which we add in this commit to `vioscsi.h`. This is presently the maximum permitted number of buses in the VirtIO standard. To facilitate the comparison we introduce the new variable `max_channels` of type `UCHAR`.

This commit also updates the initial check in `VioScsiBuildIo()` to use the new variable `Channel` of type `UCHAR`, which we define with `SRB_PATH_ID(Srb)` and compare it with `VIRTIO_SCSI_MAX_BUSES_PER_HBA` rather than `num_queues`, which we previously used in error. Using `Channel` here makes it clearer that we are performing a check to ensure the C:T:L is within bounds (plus a check to ensure the adapter has not been removed).

The reason for the previous use of `num_queues` as a check against the Channel / Bus / Path identified by Storport is unknown. Speculation as to the reason is perhaps due to a confusion with Storport `ConcurrentChannels` or as a result of porting from another Storport driver. Using `num_queues` would also accommodate use of the PDA `BUS IDENTIFIER` per the SAM and the description in the VirtIO standard to set the first byte (representing the Bus) to 1 when used to address a Target and LUN. This is the PDA method for addressing Targets and LUNs in a SCSI domain beyond the current level and is necessary when addressing targets other than LUN 0, i.e. 0:0:0, e.g. 0:1:0. However, Storport never calls the `VioScsiBuildIo()` routine with a `Path` greater than zero, but rather we use this technique to create SCSI commands for addressable targets in the initial steps to build the I/O request.

Trace messages are added for when the adapter has been removed or the device is not present. The Storport addressing jargon BTL8 is used, noting C:T:L and BTL8 addressing are synonymous with each other.

This commit also updates relevant comments.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SCSI device discovery by correctly honoring the bus, target, and LUN limits reported by the VirtIO configuration.
  * Added validation for channel, target, and LUN values to prevent unsupported requests.
  * Unsupported requests continue to complete with the appropriate “no device” status.

### DIFF
--- a/vioscsi/vioscsi.c
+++ b/vioscsi/vioscsi.c
@@ -1303,17 +1303,17 @@ VioScsiBuildIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
     PSRB_EXTENSION srbExt;
     PSTOR_SCATTER_GATHER_LIST sgList;
     VirtIOSCSICmd *cmd;
-    UCHAR TargetId;
+    UCHAR Target;
     UCHAR Lun;
 
     ENTER_FN_SRB();
     cdb = SRB_CDB(Srb);
     srbExt = SRB_EXTENSION(Srb);
     adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
-    TargetId = SRB_TARGET_ID(Srb);
+    Target = SRB_TARGET_ID(Srb);
     Lun = SRB_LUN(Srb);
 
-    if ((SRB_PATH_ID(Srb) > (UCHAR)adaptExt->num_queues) || (TargetId >= adaptExt->scsi_config.max_target) ||
+    if ((SRB_PATH_ID(Srb) > (UCHAR)adaptExt->num_queues) || (Target >= adaptExt->scsi_config.max_target) ||
         (Lun >= adaptExt->scsi_config.max_lun) || adaptExt->bRemoved)
     {
         SRB_SET_SRB_STATUS(Srb, SRB_STATUS_NO_DEVICE);
@@ -1332,7 +1332,7 @@ VioScsiBuildIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
     cmd = &srbExt->cmd;
     cmd->srb = (PVOID)Srb;
     cmd->req.cmd.lun[0] = 1;
-    cmd->req.cmd.lun[1] = TargetId;
+    cmd->req.cmd.lun[1] = Target;
     cmd->req.cmd.lun[2] = 0;
     cmd->req.cmd.lun[3] = Lun;
     cmd->req.cmd.tag = (ULONG_PTR)(Srb);
--- a/vioscsi/vioscsi.c
+++ b/vioscsi/vioscsi.c
@@ -394,9 +394,25 @@ VioScsiFindAdapter(IN PVOID DeviceExtension,
     SetGuestFeatures(DeviceExtension);
 
     ConfigInfo->NumberOfBuses = 1;
-    ConfigInfo->MaximumNumberOfTargets = min((UCHAR)adaptExt->scsi_config.max_target,
-                                             255 /*SCSI_MAXIMUM_TARGETS_PER_BUS*/);
-    ConfigInfo->MaximumNumberOfLogicalUnits = min((UCHAR)adaptExt->scsi_config.max_lun, SCSI_MAXIMUM_LUNS_PER_TARGET);
+    /* The following *NumberOf* PORT_CONFIGURATION_INFORMATION members are set by values with a zero-based index,
+     * so add one (1) to each: MaximumNumberOfTargets and MaximumNumberOfLogicalUnits.
+     * -----------------------------------------------------------------------------------------------------------
+     **  MaximumNumberOfTargets:
+     **  Most hypervisors will report the max_target value as 255 per the VirtIO standard,
+     **  and whilst SCSI port used SCSI_MAXIMUM_TARGETS_PER_BUS (128), in Storport the limit is 255.
+     **  To avoid an integer wrap-around following scsi_config.max_target + 1, we also initially cast
+     **  max_target as ULONG.
+     **
+     **  MaximumNumberOfLogicalUnits:
+     **  Most hypervisors will report the max_lun value as 16383 per the VirtIO standard,
+     **  but the Storport limit is SCSI_MAXIMUM_LUNS_PER_TARGET (255). We also need to cast to UCHAR
+     **  as MaximumNumberOfLogicalUnits is UCHAR but max_lun is ULONG. To avoid an integer wrap-around
+     **  following scsi_config.max_lun + 1, we also initially cast max_lun as ULONGLONG.
+     */
+    ConfigInfo->MaximumNumberOfTargets = (UCHAR)min((ULONG)adaptExt->scsi_config.max_target + 1, 255);
+    ConfigInfo->MaximumNumberOfLogicalUnits = (UCHAR)min((ULONGLONG)adaptExt->scsi_config.max_lun + 1,
+                                                         SCSI_MAXIMUM_LUNS_PER_TARGET);
+
     ConfigInfo->MaximumTransferLength = SP_UNINITIALIZED_VALUE;  // Unlimited
     ConfigInfo->NumberOfPhysicalBreaks = SP_UNINITIALIZED_VALUE; // Unlimited
 
@@ -1313,8 +1329,8 @@ VioScsiBuildIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
     Target = SRB_TARGET_ID(Srb);
     Lun = SRB_LUN(Srb);
 
-    if ((SRB_PATH_ID(Srb) > (UCHAR)adaptExt->num_queues) || (Target >= adaptExt->scsi_config.max_target) ||
-        (Lun >= adaptExt->scsi_config.max_lun) || adaptExt->bRemoved)
+    if ((SRB_PATH_ID(Srb) > (UCHAR)adaptExt->num_queues) || (Target > adaptExt->scsi_config.max_target) ||
+        (Lun > adaptExt->scsi_config.max_lun) || adaptExt->bRemoved)
     {
         SRB_SET_SRB_STATUS(Srb, SRB_STATUS_NO_DEVICE);
         SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
--- a/vioscsi/vioscsi.c
+++ b/vioscsi/vioscsi.c
@@ -348,6 +348,7 @@ VioScsiFindAdapter(IN PVOID DeviceExtension,
     ULONG num_cpus;
     ULONG max_cpus;
     ULONG max_queues;
+    UCHAR max_channels;
 
     UNREFERENCED_PARAMETER(HwContext);
     UNREFERENCED_PARAMETER(BusInformation);
@@ -393,10 +394,19 @@ VioScsiFindAdapter(IN PVOID DeviceExtension,
     GetScsiConfig(DeviceExtension);
     SetGuestFeatures(DeviceExtension);
 
-    ConfigInfo->NumberOfBuses = 1;
     /* The following *NumberOf* PORT_CONFIGURATION_INFORMATION members are set by values with a zero-based index,
-     * so add one (1) to each: MaximumNumberOfTargets and MaximumNumberOfLogicalUnits.
+     * so add one (1) to each: NumberOfBuses, MaximumNumberOfTargets and MaximumNumberOfLogicalUnits.
      * -----------------------------------------------------------------------------------------------------------
+     **  NumberOfBuses:
+     **  Most hypervisors will report the max_channel value as zero (0) per the VirtIO standard.
+     **  The supported limit of virtio-scsi is VIRTIO_SCSI_MAX_BUSES_PER_HBA (1). The Storport limit
+     **  is SCSI_MAXIMUM_BUSES_PER_ADAPTER (255). To retain similar logic to max_target and max_lun
+     **  as shown below, and provide for additional channels should the VIrtIO standard ever accommodate
+     **  more buses per adapter, we define max_channels as the lesser of SCSI_MAXIMUM_BUSES_PER_ADAPTER
+     **  and VIRTIO_SCSI_MAX_BUSES_PER_HBA. We then use max_channels and scsi_config.max_channel + 1 as
+     **  limits. To avoid an integer wrap-around following scsi_config.max_channel + 1, we also initially
+     **  cast max_channel as ULONG.
+     **
      **  MaximumNumberOfTargets:
      **  Most hypervisors will report the max_target value as 255 per the VirtIO standard,
      **  and whilst SCSI port used SCSI_MAXIMUM_TARGETS_PER_BUS (128), in Storport the limit is 255.
@@ -409,6 +419,8 @@ VioScsiFindAdapter(IN PVOID DeviceExtension,
      **  as MaximumNumberOfLogicalUnits is UCHAR but max_lun is ULONG. To avoid an integer wrap-around
      **  following scsi_config.max_lun + 1, we also initially cast max_lun as ULONGLONG.
      */
+    max_channels = (UCHAR)min(VIRTIO_SCSI_MAX_BUSES_PER_HBA, SCSI_MAXIMUM_BUSES_PER_ADAPTER);
+    ConfigInfo->NumberOfBuses = (UCHAR)min((ULONG)adaptExt->scsi_config.max_channel + 1, max_channels);
     ConfigInfo->MaximumNumberOfTargets = (UCHAR)min((ULONG)adaptExt->scsi_config.max_target + 1, 255);
     ConfigInfo->MaximumNumberOfLogicalUnits = (UCHAR)min((ULONGLONG)adaptExt->scsi_config.max_lun + 1,
                                                          SCSI_MAXIMUM_LUNS_PER_TARGET);
@@ -1319,6 +1331,7 @@ VioScsiBuildIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
     PSRB_EXTENSION srbExt;
     PSTOR_SCATTER_GATHER_LIST sgList;
     VirtIOSCSICmd *cmd;
+    UCHAR Channel;
     UCHAR Target;
     UCHAR Lun;
 
@@ -1326,12 +1339,27 @@ VioScsiBuildIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
     cdb = SRB_CDB(Srb);
     srbExt = SRB_EXTENSION(Srb);
     adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
+    Channel = SRB_PATH_ID(Srb);
     Target = SRB_TARGET_ID(Srb);
     Lun = SRB_LUN(Srb);
 
-    if ((SRB_PATH_ID(Srb) > (UCHAR)adaptExt->num_queues) || (Target > adaptExt->scsi_config.max_target) ||
+    if ((Channel >= VIRTIO_SCSI_MAX_BUSES_PER_HBA) || (Target > adaptExt->scsi_config.max_target) ||
         (Lun > adaptExt->scsi_config.max_lun) || adaptExt->bRemoved)
     {
+        if (adaptExt->bRemoved)
+        {
+            RhelDbgPrint(TRACE_LEVEL_ERROR,
+                         " Storport attempted to build an I/O request but the adapter is no longer present.\n");
+        }
+        else
+        {
+            RhelDbgPrint(TRACE_LEVEL_ERROR,
+                         " Storport attempted to build an I/O request for a device that is not present."
+                         " The BTL8 address is %d:%d:%d.\n",
+                         Channel,
+                         Target,
+                         Lun);
+        }
         SRB_SET_SRB_STATUS(Srb, SRB_STATUS_NO_DEVICE);
         SRB_SET_DATA_TRANSFER_LENGTH(Srb, 0);
         StorPortNotification(RequestComplete, DeviceExtension, Srb);
--- a/vioscsi/vioscsi.h
+++ b/vioscsi/vioscsi.h
@@ -119,6 +119,8 @@ typedef struct VirtIOBufferDescriptor VIO_SG, *PVIO_SG;
 #define VIRTIO_SCSI_REQUEST_QUEUE_0          2
 #define VIRTIO_SCSI_QUEUE_LAST               VIRTIO_SCSI_REQUEST_QUEUE_0 + MAX_CPU
 
+#define VIRTIO_SCSI_MAX_BUSES_PER_HBA        1
+
 /* MSI messages and virtqueue indices are offset by 1, MSI 0 is not used */
 #define QUEUE_TO_MESSAGE(QueueId)            ((QueueId) + 1)
 #define MESSAGE_TO_QUEUE(MessageId)          ((MessageId)-1)


YanVugenfirer:

benyamin-codez:
Alternative to PR #1443.

vioscsi/vioscsi.c
```diff
@@ -442,6 +439,36 @@ VioScsiFindAdapter(IN PVOID DeviceExtension,
...
+     **  as MaximumNumberOfLogicalUnits is UCHAR but max_lun is ULONG.
+     */
+    if (adaptExt->num_queues > 255)
+    {
+        max_channels = SCSI_MAXIMUM_BUSES_PER_ADAPTER;
+    }
+    else
+    {
+        max_channels = (UCHAR)adaptExt->num_queues;
+    }
```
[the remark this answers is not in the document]
benyamin-codez: Done.

vioscsi/vioscsi.c
```diff
@@ -442,6 +439,36 @@ VioScsiFindAdapter(IN PVOID DeviceExtension,
...
+    ConfigInfo->NumberOfBuses = min(adaptExt->scsi_config.max_channel + 1, max_channels);
+    ConfigInfo->MaximumNumberOfTargets = min(adaptExt->scsi_config.max_target + 1, 255);
+    ConfigInfo->MaximumNumberOfLogicalUnits = (UCHAR)min(adaptExt->scsi_config.max_lun + 1,
+                                                         SCSI_MAXIMUM_LUNS_PER_TARGET);
```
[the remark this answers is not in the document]
benyamin-codez: Done.
Also did the same for `scsi_config.max_channel` and `scsi_config.max_target` but casting with `ULONG`.

vioscsi/vioscsi.c
```diff
@@ -1782,7 +1811,7 @@ VOID LogError(IN PVOID DeviceExtension, IN ULONG ErrorCode, IN ULONG UniqueId)
 
 VOID TransportReset(IN PVOID DeviceExtension, IN PVirtIOSCSIEvent evt)
 {
-    UCHAR TargetId = evt->lun[1];
+    UCHAR Target = evt->lun[1];
     UCHAR Lun = (evt->lun[2] << 8) | evt->lun[3];
```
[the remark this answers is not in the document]
benyamin-codez: Proposed solution in PR #1512.

vioscsi/vioscsi.c
```diff
@@ -1802,7 +1831,7 @@ VOID TransportReset(IN PVOID DeviceExtension, IN PVirtIOSCSIEvent evt)
 
 VOID ParamChange(IN PVOID DeviceExtension, IN PVirtIOSCSIEvent evt)
 {
-    UCHAR TargetId = evt->lun[1];
+    UCHAR Target = evt->lun[1];
     UCHAR Lun = (evt->lun[2] << 8) | evt->lun[3];
```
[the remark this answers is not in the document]
benyamin-codez: Proposed solution in PR #1512.

benyamin-codez:
@YanVugenfirer 

Yan, could I please have your initial take on this to determine if there's any further drafting required?

I note my comments [here](https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1443#issuecomment-3829302144) in the related PR.

Best regards,
Ben

cc: @vrozenfe

YanVugenfirer:
Second thought regarding the  "+1" changes.

I don't think they are needed at all. Theoretically a device with zero LUNs can exists (it is not functional), but still.
Adding +1 unconditionally is definitely a mistake as it creates a discrepancy between actual amount of Luns, channels and what Windows will think is the actual amount. 
So I don't think this code should be changed at all.

Regarding limit by channel and not by queues - let's discuss it first in the issue without the noise of other code changes.

hrosenfeld:
These values aren't counts, these are the maximum address values for channel, target, and LUN. A virtio-scsi emulation reporting a maximum target of 0 supports exactly one target with address 0. A target with a maximum LUN of 0 has only one LUN, which is LUN 0 (Which, btw., is mandated to exist by the T10 specs).

Thus, if you want to convert the maximum into a count, you add +1.

benyamin-codez:
@YanVugenfirer 

Yan, given the clarification @hrosenfeld provided regarding the `VirtIOSCSIConfig` struct members `max_channel`, `max_target` and `max_lun` being the upper bound of C:T:L address IDs - a zero-based index, which should make the need for these changes clear, is it ok to discuss here the limit by channel rather than by queues part of the PR rather than in Hans' issue...? His issue doesn't touch on that part, so my thoughts were, here might be better.

I should really preface any discussion on that part by stating the following:

I still limit by queues, per:
```c
max_channels = (UCHAR)min(adaptExt->num_queues, SCSI_MAXIMUM_BUSES_PER_ADAPTER);
```
... but, I only did this to retain the function of the current code.

I'm not convinced the relationship established here between SCSI channels and virtqueues is correct.
There might be a confusion with the `ChannelNumber` in `_STARTIO_PERFORMANCE_PARAMETERS`.
In that context there is a relationship between `ConcurrentChannels` and `num_queues`...
... but here "channels" is synonymous with "buses" - via the H:C:T:L<-->BTL8 nomenclature relationship...
... and the virtqueues could likely be used for any bus.

Regardless, in every known case to me, the value of of `max_channels = 0` and thus `NumberOfBuses =  1`.

So I am minded to drop this part out with appropriate commentary in the commit message.

Let me know what you think.

Best regards,
Ben

YanVugenfirer:
@benyamin-codez I suggest @vrozenfe and @kevmw to voice their ideas about channel VS queues relations.

YanVugenfirer:
> These values aren't counts, these are the maximum address values for channel, target, and LUN. A virtio-scsi emulation reporting a maximum target of 0 supports exactly one target with address 0. A target with a maximum LUN of 0 has only one LUN, which is LUN 0 (Which, btw., is mandated to exist by the T10 specs).
> 
> Thus, if you want to convert the maximum into a count, you add +1.

@hrosenfeld @benyamin-codez Sorry, you are right. After another review and looking into Linux kernel where the virtio spec used concepts originated I understand the difference between the spec and Windows exceptions.

vioscsi/vioscsi.c
```diff
@@ -1317,17 +1329,19 @@ VioScsiBuildIo(IN PVOID DeviceExtension, IN PSCSI_REQUEST_BLOCK Srb)
...
     VirtIOSCSICmd *cmd;
+    UCHAR Channel;
     UCHAR Target;
     UCHAR Lun;
 
     ENTER_FN_SRB();
     cdb = SRB_CDB(Srb);
     srbExt = SRB_EXTENSION(Srb);
     adaptExt = (PADAPTER_EXTENSION)DeviceExtension;
+    Channel = SRB_PATH_ID(Srb);
```
YanVugenfirer, line 1342: I suggest to remove this commit from PR so we can move forward with merging for the off by one error. and we can discuss channel VS queue relations later.

YanVugenfirer (changes requested):

vioscsi/vioscsi.c
```diff
@@ -394,9 +394,23 @@ VioScsiFindAdapter(IN PVOID DeviceExtension,
...
+     **  max_target as ULONG.
+     **
+     **  Most hypervisors will report the max_lun value as 16383 per the VirtIO standard,
+     **  but the Storport limit is SCSI_MAXIMUM_LUNS_PER_TARGET (255). We also need to cast to UCHAR
+     **  as MaximumNumberOfLogicalUnits is UCHAR but max_lun is ULONG. To avoid an integer wrap-around
+     **  following scsi_config.max_lun + 1, we also initially cast max_lun as ULONGLONG.
+     */
+    ConfigInfo->MaximumNumberOfTargets = (UCHAR)min((ULONG)adaptExt->scsi_config.max_target + 1, 255);
+    ConfigInfo->MaximumNumberOfLogicalUnits = (UCHAR)min((ULONGLONG)adaptExt->scsi_config.max_lun + 1,
+                                                         SCSI_MAXIMUM_LUNS_PER_TARGET);
```
kevmw, line 426: The `ULONG` cast is unnecessary. The addition is performed on `int`, which is enough to avoid an integer overflow because max_target is `u16`. The `ULONGLONG` cast could be replaced by `max_lun + 1ULL`.

While the cast to `UCHAR` is technically redundant, keeping it for documentation purposes may make sense.

vioscsi/vioscsi.c
```diff
@@ -407,6 +417,8 @@ VioScsiFindAdapter(IN PVOID DeviceExtension,
      **  as MaximumNumberOfLogicalUnits is UCHAR but max_lun is ULONG. To avoid an integer wrap-around
      **  following scsi_config.max_lun + 1, we also initially cast max_lun as ULONGLONG.
      */
+    max_channels = (UCHAR)min(adaptExt->num_queues, SCSI_MAXIMUM_BUSES_PER_ADAPTER);
```
kevmw: I'm not sure where this one comes from. I don't think the number of buses a SCSI HBA exposes depends on the number of virtqueues it uses. Did you find anything to that effect in the virtio spec? On the other hand, the spec also doesn't really say how to use multiple buses (nor does this commit add any code to enable it, which would be untestable anyway), so maybe keeping the hardcoded `NumerOfBuses = 1` is best.

kevmw:
> WARNING: This commit breaks driver operation. To restore operation the new
>         section must be relocated to after num_queues is properly defined.
>         This is accomplished by the next commit in this series.

I don't know if the virtio-win project allows this, but breaking bisectability isn't good practice. In this specific case it should be easy to fix by reordering the patches. Of course, my other comment suggests that it shouldn't even depend on `num_queues`, in which case this is moot.

vioscsi/vioscsi.c
hrosenfeld: It certainly doesn't depend on the virtqueues, and I agree with keeping this hardcoded at 1 for now. The virtio spec not only says that the number of channels should be 0, it also doesn't explain how to address a channel other than 0.

While in theory the LUN format used by virtio-scsi would allow for addressing channels/buses using the hierarchical addressing scheme as specified by SAM-5, the virtio spec is very clear on the limited subset of hierarchical addressing it supports without ever even calling it hierarchical addressing. (The bus address must be hardcoded to 1 in each LUN address for it to be valid according to the virtio spec.)

Anyway, as of now there's no spec-conforming way for a virtio-scsi device to have more than one bus, and there's no spec-conforming way for a virtio-scsi driver to address more than 1 bus.

vioscsi/vioscsi.c
benyamin-codez: > The `ULONG` cast is unnecessary. The addition is performed on `int`, which is enough to avoid an integer overflow because max_target is `u16`. 

Some legacy and embedded systems still implement `int` as 2 bytes, so a theoretical value of 0xFFFFh for `max_target` will still overflow on addition in those scenarios. 

> ... The `ULONGLONG` cast could be replaced by `max_lun + 1ULL`.

This format is really for static analysis, which seem to prefer the explicit cast rather than using arithmetic promotion techniques. We could maybe try `1ULL + max_lun`, but I'm not convinced the static analysis tools employed will even necessarily catch that.

> While the cast to `UCHAR` is technically redundant, keeping it for documentation purposes may make sense.

All the members of `NumberOf*` `PORT_CONFIGURATION_INFORMATION` members subject to the definition are UCHAR. Static analysis tools pick up on this too.

vioscsi/vioscsi.c
benyamin-codez: Looks like I was a bit slow and this might be resolved anyway.

vioscsi/vioscsi.c
benyamin-codez: I think we all have the same opinion here regarding a dependency on `num_queues`, i.e. there is none.

> ...(The bus address must be hardcoded to 1 in each LUN address for it to be valid according to the virtio spec.)
>
> Anyway, as of now there's no spec-conforming way for a virtio-scsi device to have more than one bus, and there's no spec-conforming way for a virtio-scsi driver to address more than 1 bus.

I'm not sure that is accurate because the VirtIO verbiage is SHOULD and not MUST...

The real test should be if Storport runs with `NumOfBuses > 1`, and if so, let it.
If not, we should limit this to 1, with either the existing hard-coded technique or a conditional limiter.
The latter allows for an easy change at a later date if the need ever arises and will allow for maximum interoperability in the event a hypervisor should wish to employ multiple buses.
So I will run some quick tests to see what Storport does...

In any case, I will clean this up ASAP.

hrosenfeld: Oh, I'm not going to argue about the verbiage being SHOULD and not MUST w.r.t. max_channel, that part is obvious.

But if you want to allow (even just for the future) to have a value > 0 (translating to NumOfBuses > 1), then I'd ask where in a virtio-scsi request you'll set the bus number so the virtio-scsi device knows what bus the request addresses.

benyamin-codez: You raise a very good point, Hans.

I guess that is the grey area, because the `lun[8]` member is not terribly well defined:

> <ins>**When used to address a target and logical unit**</ins>, the only supported format for _lun_ is: <ins>**first byte set to 1**</ins>, second byte set to target, third and fourth byte representing a single level LUN structure, followed by four zero bytes. _[Emphasis mine]_

If we assume the peripheral device addressing method is used, per the SAM - as _appears_ to be the case - it does seem to suggest the first byte here is for the bus, but at the same time the standard only refers to addressing the target and LUN...

Having said that, I do think it is fair to say that the words _"only supported format"_ indicate only a single bus is supported.

So based on the above, if we agree to hard code `NumberOfBuses = 1` but maybe add a relevant comment...
...and we agree there is no dependency on `num_queues`...
...so this is meaningless: `SRB_PATH_ID(Srb) > (UCHAR)adaptExt->num_queues`
==> Should we still retain the proposed sanity check `Channel > adaptExt->scsi_config.max_channel`...
...where `Channel = SRB_PATH_ID(Srb)`...
...or just delete it...?

I'm inclined to keep the proposed sanity check as is.

cc: @kevmw

kevmw: In theory, we should be able to rely on storport passing us a valid address, but it can't hurt to be a bit more defensive and check the value even if we don't use it in the foreseeable future.

However, if we're ignoring `scsi_config.max_channel` and keep the hardcoded `NumberOfBuses = 1` (which I believe we should do), then it doesn't make sense to compare anything against an arbitrary `scsi_config.max_channel` that we never validated and that might not match the driver's capabilities. Instead, we should probably check `SRB_PATH_ID(Srb) > 0` (assuming it's zero-based, might be 1 instead if it follows the SCSI `BUS IDENTIFIER` definition).

benyamin-codez: > ... we should probably check `SRB_PATH_ID(Srb) > 0` (assuming it's zero-based, might be 1 instead if it follows the SCSI `BUS IDENTIFIER` definition).

This is where I settled too, i.e. the SCSI `BUS IDENITFIER` PDA mechanism, might be why this ended up being checked against `num_queues` (in combination with some other misunderstanding or improper port)...

I'm now fairly confident the VirtIO bus / channel implementation is sufficiently elucidated to resolve this part. I have an alternate proposal in my test pipeline, which is presently being checked to see if we ever observe a BUS ID of 1.

Once I'm done with that I'll push updated commits here and if that get's up I move this to ready for review.
Otherwise I'll split this bit off per Yan's suggestion in order to keep it moving.

benyamin-codez: Storport never reported a Path of 1. This is expected. We use a `BUS IDENTIFIER` of 1 when creating the SCSI command in the initial steps of building the I/O request. I updated the commit message to discuss this briefly.

benyamin-codez:
@kevmw @hrosenfeld 

Kevin and Hans, if you are happy with this latest cut, esp. wrt to the change to `max_channel` / `NumberOfBuses`, I will move this to ready for review.

I think the intro of a macro definition to cover the hard coding is a good compromise and covers the bases for any future changes in the standard re number of buses. I also cleaned up irrelevant touching of other routines and added some trace messages for out-of-bounds devices, etc.

Let me know...

Best regards,
Ben

hrosenfeld:
Looks good to me.

hrosenfeld:
Ping? Will this fix be committed any time soon?

benyamin-codez:
> Ping? Will this fix be committed any time soon?

Thanks for the follow up Hans. Apologies for dropping the ball. I've had a lot going on of late.
I'll switch this to ready for review...

@YanVugenfirer 

Yan, I think the issues around channels has been thoroughly discussed and is now well understood.
I'm happy to capture this by leaving in the proposed changes as-is.

cc: @kevmw @vrozenfe @kostyanf14

hrosenfeld:
Ping?

YanVugenfirer:
re-run tests

vioscsi/vioscsi.c
```diff
@@ -393,10 +394,37 @@ VioScsiFindAdapter(IN PVOID DeviceExtension,
...
+     **  Most hypervisors will report the max_lun value as 16383 per the VirtIO standard,
+     **  but the Storport limit is SCSI_MAXIMUM_LUNS_PER_TARGET (255). We also need to cast to UCHAR
+     **  as MaximumNumberOfLogicalUnits is UCHAR but max_lun is ULONG. To avoid an integer wrap-around
+     **  following scsi_config.max_lun + 1, we also initially cast max_lun as ULONGLONG.
+     */
+    max_channels = (UCHAR)min(VIRTIO_SCSI_MAX_BUSES_PER_HBA, SCSI_MAXIMUM_BUSES_PER_ADAPTER);
+    ConfigInfo->NumberOfBuses = (UCHAR)min((ULONG)adaptExt->scsi_config.max_channel + 1, max_channels);
+    ConfigInfo->MaximumNumberOfTargets = (UCHAR)min((ULONG)adaptExt->scsi_config.max_target + 1, 255);
+    ConfigInfo->MaximumNumberOfLogicalUnits = (UCHAR)min((ULONGLONG)adaptExt->scsi_config.max_lun + 1,
+                                                         SCSI_MAXIMUM_LUNS_PER_TARGET);
```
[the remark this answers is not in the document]
benyamin-codez: iinm, after writing the `ConfigInfo` Storport limits these IDs in the SRB, so we don't expect SRBs with out-of-bounds IDs. The sanity check in `VioScsiBuildIo` looks for the edge case when a hypervisor actually reports a lower ID, i.e. the actual ID maxima rather than that of the VirtIO standard.

Methinks nothing to see here....

benyamin-codez:
Looks like we are right to go again...

benyamin-codez:
Yan, the failing check is clang-format for . in vioserial, I presume from your merges...

YanVugenfirer (approved):

YanVugenfirer:
@benyamin-codez Sorry, some issues with CI and Github API. Can you resend the PR as new PR on top of master branch?

benyamin-codez:
No problem, Yan. 
Should be able to do that in the next 12 hours or so.
With kind regards, Ben

benyamin-codez:
New PR is #1654...

YanVugenfirer:
@benyamin-codez Please drop my merge commit "[df94d8b](https://github.com/virtio-win/kvm-guest-drivers-windows/commit/df94d8bd3d8cea90ecd67f071bf83abcf7ff139b)" from PR, it was done by mistake. Thanks.

benyamin-codez:
@YanVugenfirer, methinks this one can be closed.
I raised PR #1654 on top of your commit b5cc02f to replace this PR as requested...

PR #1520, opened 2026-02-12
RHEL-149866: viostor: Fix device configuration updates

kevmw:
This fixes viostor to handle Device Configuration Interrupts more correctly.

In the case of `msix_has_config_vector == false`, the driver didn't even try to get the notification, but it disabled the config vector entirely. This is changed to share vector 0 between a queue and config updates. This is easy to reproduce: After resizing the disk (with `block_resize` in the QEMU monitor), the change would not be picked up by Windows if the disk had `vectors < num_queues + 1` until the device is disabled and reenabled. With the fixed driver, the new size is recognised and shown e.g. in Disk Management after hitting F5.

The fixes in `VirtIoInterrupt()` make the driver more compliant with the spec, but are more theoretical in nature.

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -1270,7 +1270,7 @@ VirtIoInterrupt(IN PVOID DeviceExtension)
         return FALSE;
     }
     intReason = virtio_read_isr_status(&adaptExt->vdev);
-    if (intReason == 1 || adaptExt->dump_mode)
+    if (intReason & 0x1 || adaptExt->dump_mode)
     {
         if (!CompleteDPC(DeviceExtension, adaptExt->msix_has_config_vector))
         {
@@ -1278,7 +1278,7 @@ VirtIoInterrupt(IN PVOID DeviceExtension)
         }
         isInterruptServiced = TRUE;
     }
-    else if (intReason == 3)
+    if (intReason & VIRTIO_PCI_ISR_CONFIG)
     {
         VirtIoConfigUpdated(adaptExt);
         isInterruptServiced = TRUE;


kostyanf14 (approved):

YanVugenfirer:
@kevmw Sorry for nitpicking, next time please include Jira key at least in the PR title (based on https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/contributing.md for Red Hat contributors - that was already our practice before Scrum).

benyamin-codez:
Doh, too slow again...
Presently running performance checks between this, previous master and my branch (yet to push)...
Will post any remarkable outcomes here...

YanVugenfirer:
While having performance numbers is nice. It is pretty much all or nothing scenario that works like legacy interrupt case.

kevmw:
@YanVugenfirer I had read that, but didn't realise I was supposed to create new Jira tickets for new upstream work as opposed to just referencing already existing ones I was addressing. Do we need the ticket overhead for every single PR or only user-visible features/bug fixes? Does it have to be a separate ticket for each PR or can multiple PRs share one? In other words, are fewer large PRs easier to handle than many small and self-contained PRs?

(I also don't think I should be treated any differently from other community contributors, given that I'm not part of your team, or in fact any difference should be made between Red Hat employees and other contributors in an upstream community, but if it's the project policy, I'll try to follow it.)

YanVugenfirer:
@kevmw The policy was originated many years ago to help QE to identify what should they test during acceptance cycle and enable them to understand the changes and decide if some changes need to trigger test plan updates. 

I prefer to have smaller PRs in general, as they are easier to review and it prevents large code changes handing for relatively small changes in one of the commits. 

Before Scrum one ticket for multiple PRs was OK. Unfortunately after Scrum, I think it is more strict as QE tasks are split from the story\bugs (previously we didn't have QE tasks at all). Still if small PRs are expected to be accepted is short time it is OK to create one ticket.

A fast way to create Jira ticket from upstream issues or PR is to add GitHub PR\issue URL to "Summary" field and let the automation bot to create the issue.

PR #1527, opened 2026-02-25
Remove an oversight from the mergeable RX buffer PR.

zjmletang:
## Background

The `MAX_MERGED_PHYSICAL_PAGES` macro (set to 18) and its associated comment
were left behind from an earlier design iteration.

In the early design, physical page aliasing was used to preserve
`ParaNdis_BindRxBufferToPacket` compatibility: the existing MDL binding
function always started from `PhysicalPages[1]` (legacy design for traditional
mode). By creating an alias where both `PhysicalPages[0]` and `[1]` pointed to
the same physical memory, the function could work correctly for mergeable
buffers without modification, reducing regression risk. This led to the
calculation: first buffer (2 logical pages) + 16 additional buffers = 18 pages.

Later, this approach was replaced by introducing the `FirstRxDataPage` field
(instead of the hardcoded `PARANDIS_FIRST_RX_DATA_PAGE` macro), which provided
a cleaner solution. However, the redundant macro and outdated comment were
overlooked during that change.

## Changes

Remove `MAX_MERGED_PHYSICAL_PAGES` and use `VIRTIO_NET_MAX_MRG_BUFS` (17)
directly, since the final implementation uses exactly 1 page per buffer.

### DIFF
--- a/NetKVM/Common/ParaNdis-RX.h
+++ b/NetKVM/Common/ParaNdis-RX.h
@@ -68,12 +68,9 @@ class CParaNdisRX : public CParaNdisTemplatePath<CVirtQueue>, public CNdisAlloca
 
 // Maximum mergeable packet size per VirtIO spec: 65562 bytes (including 12-byte header)
 // Required buffers: ceil(65562 / 4096) = 17 PAGE-sized buffers maximum
-#define VIRTIO_NET_MAX_MRG_BUFS   17
+#define VIRTIO_NET_MAX_MRG_BUFS 17
 
-// Merge buffer context structure - pre-allocated to avoid hot-path allocation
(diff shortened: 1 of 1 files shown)


YanVugenfirer:

YanVugenfirer:
ok to test

zjmletang:
There are 2 failures in total, and preliminary analysis suggests that neither of them appear to be rooted in the driver itself:

HLK Win2022x64 Failure Analysis:
NICStrictPropertyValidation: The test suite has expired
image


HLK Win10 22H2 x64 Failure Analysis:
NDISTest 6.0 - [2 Machine] - 2c_Mini6RSSSendRecv (Multi-Group Win8+): Based on local testing, this failure seems to be eligible for an HLK filter.
image

YanVugenfirer:
@zjmletang Thanks for the failure review!

Merging.

PR #1549, opened 2026-03-21
netkvm: treat temporary TX queue full as retry, not drop

zjmletang:
This series tries to clarify the TX-side handling of SUBMIT_NO_PLACE_IN_QUEUE and align the code flow with that meaning.

My understanding is that SUBMIT_NO_PLACE_IN_QUEUE represents temporary resource pressure (queue full / retry later), not a final transmit failure.
If that interpretation is off, I would really appreciate corrections.

Commit 1
netkvm: do not count temporary TX queue full as packet drop

This commit separates the SUBMIT_NO_PLACE_IN_QUEUE path from drop accounting, so temporary backpressure is not reported as dropped TX packets.

The goal is to avoid misleading TX statistics when packets are retried later instead of being lost.

Commit 2
netkvm: reorder SendMapped switch cases as success/retry/failure

This commit is intended as a readability/maintainability cleanup only (no functional change intended).
It handles SUBMIT_SUCCESS explicitly and orders the switch logic as:

success
retry
failure
This makes the control flow easier to follow and reduces the chance of mixing retry semantics with failure semantics in future changes.

Why these two commits are in one patch series
Both commits address the same semantic point from two angles:

Commit 1 fixes the observable behavior (statistics)
Commit 2 makes the related control flow clearer and easier to maintain
I grouped them together because they are tightly related in intent.
That said, if maintainers prefer the cleanup commit to be split further, I’m happy to reorganize accordingly.

### DIFF
--- a/NetKVM/Common/ParaNdis_TX.cpp
+++ b/NetKVM/Common/ParaNdis_TX.cpp
@@ -893,6 +893,24 @@ bool CParaNdisTX::SendMapped(bool IsInterrupt, CRawCNBLList &toWaitingList)
 
                 switch (result)
                 {
+                    case SubmitTxPacketResult::SUBMIT_SUCCESS:
+                        // if this NBL finished?
+                        if (!NBLHolder->HaveMappedBuffers())
+                        {
+                            /* We use PeekMappedNBL method to get the current NBL
+                             * that should be processed from the queue, when we finish
+                             * sending all it's NBs, we should pop it from the queue.
+                             */
+                            PopMappedNBL();
+                            toWaitingList.Push(NBLHolder);
+                        }
+                        else
+                        {
+                            // no, keep it in the queue
+                        }
+                        SentOutSomeBuffers = true;
+                        break;
+
                     case SubmitTxPacketResult::SUBMIT_NO_PLACE_IN_QUEUE:
                         NBLHolder->PushMappedNB(NBHolder);
                         HaveBuffers = false;
@@ -900,8 +918,6 @@ bool CParaNdisTX::SendMapped(bool IsInterrupt, CRawCNBLList &toWaitingList)
                         break;
 
                     case SubmitTxPacketResult::SUBMIT_FAILURE:
-                        __fallthrough;
-                    case SubmitTxPacketResult::SUBMIT_SUCCESS:
                         __fallthrough;
                     case SubmitTxPacketResult::SUBMIT_PACKET_TOO_LARGE:
                         // if this NBL finished?
@@ -918,16 +934,8 @@ bool CParaNdisTX::SendMapped(bool IsInterrupt, CRawCNBLList &toWaitingList)
                         {
                             // no, keep it in the queue
                         }
-
-                        if (result == SubmitTxPacketResult::SUBMIT_SUCCESS)
-                        {
-                            SentOutSomeBuffers = true;
-                        }
-                        else
-                        {
-                            CNB::Destroy(NBHolder);
-                            NBLHolder->NBComplete();
-                        }
+                        CNB::Destroy(NBHolder);
+                        NBLHolder->NBComplete();
                         break;
                     default:
                         NETKVM_ASSERT(false);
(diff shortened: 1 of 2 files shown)


kostyanf14:
ok to test

zjmletang:
CI Failed Test Items – Initial Analysis
NetKVM-Win10_22H2x64 — NDISTest 6.0 [2 Machine] 2c_Mini6RSSSendRecv (Multi-Group Win8+): Known environment issue 
NetKVM-Win2022x64 — NICStrictPropertyValidation: Known environment issue 
NetKVM-Win2022x64 — NDISTest 6.5 [1 Machine] OffloadParity: local Rerun passed; still suspected to be environment-related.
NetKVM-Win2025x64_host — NDISTest 6.5 [2 Machine] GlitchFreeDevice:  Primary focus — not understood yet; fails locally as well, and also fails on the master branch.
image

kostyanf14:
@zjmletang 

> NetKVM-Win2025x64_host — NDISTest 6.5 [2 Machine] GlitchFreeDevice: Primary focus — not understood yet; fails locally as well, and also fails on the master branch.

Did you enable "GlitchFreeDevice_NdisPollMode_workaround" in local test? 
Explanation: https://github.com/HCK-CI/AutoHCK/commit/42a67096db897400f08d59de11d3f4e39b080a17

kostyanf14:
For some reason, Server 2025 on CI was run without "GlitchFreeDevice_NdisPollMode_workaround". I rerun with it

kostyanf14:
@zjmletang NetKVM-Win2025x64_host with GlitchFreeDevice_NdisPollMode_workaround pass

zjmletang:
> @zjmletang NetKVM-Win2025x64_host with GlitchFreeDevice_NdisPollMode_workaround pass

@kostyanf14 thanks.I didn’t use this workaround locally—I’ll deploy it later as well

zjmletang:
hi @YanVugenfirer @ybendito It looks like there are no blocking HLK issues for this series now.

When you have a chance, could you please review it?

YanVugenfirer:
@zjmletang Sorry, a lot of activities this week, I will give additional look next week

ybendito:
@zjmletang Regarding first commit:  If I understand correctly this is a logical bug in the code. I think we never face it in any real scenario (due to proper configuration of number of descriptors according to queue size). Can you please confirm (or disconfirm) that?

zjmletang:
> @zjmletang Regarding first commit: If I understand correctly this is a logical bug in the code. I think we never face it in any real scenario (due to proper configuration of number of descriptors according to queue size). Can you please confirm (or disconfirm) that?

hi @ybendito Actually I believe this is not just a theoretical bug — I've reproduced it in practice. When VIRTIO_RING_F_INDIRECT_DESC is not negotiated (direct mode), each packet consumes multiple ring descriptors per submission. Under sustained small-packet TCP traffic with Nagle's algorithm enabled, ring descriptors get exhausted, hitting this path and incorrectly counting a retried packet as a drop.

zjmletang:
> @zjmletang Sorry, a lot of activities this week, I will give additional look next week

hi @YanVugenfirer No worries at all, there’s no rush. This patch doesn’t address any major issue, so a lower priority is totally fine. Please just take a look whenever you have time

ybendito:
@zjmletang Without indirect, When we do not have enough entries in the queue I think the flow drops earlier, at https://github.com/virtio-win/kvm-guest-drivers-indows/blob/master/NetKVM/Common/ParaNdis_VirtQueue.cpp#L376
I'm thinking whether we need more changes to prevent the packet queued in TX path forever when it can't be sent.

zjmletang:
hi @ybendito 
Thanks for the L376 pointer — that is exactly the line I hit in my reproduction (direct mode, back-to-back small TCP segments with Nagle on).

On the scope of this commit: pre-fix, SUBMIT_NO_PLACE_IN_QUEUE fell through to SUBMIT_FAILURE and called NB.Report(0, false), which only increments extraStatistics.droppedTxPackets. SubmitPacket still returns SUBMIT_NO_PLACE_IN_QUEUE, and the NB is still pushed back via PushMappedNB and retried. So the data path is identical before and after — only the drop counter changes.

On the concern about packets queued forever: since the data path is unchanged here, the retry loop (DPC / Poll → DoPendingTasks → ProcessTXCompletions → SendMapped) behaves the same as before. Do you think it warrants further hardening? If so, I'd be glad to address it as a follow-up patch — open to your suggestions on the direction.

ybendito (approved):

PR #1550, opened 2026-03-22
netkvm: Add ntttcp performance test scripts

zjmletang:
NTTTCP Performance Test Scripts
===============================

Prerequisites
-------------
- Python 3.6+
- ntttcp.exe (download from https://github.com/microsoft/ntttcp)
- Two Windows machines connected via network

Setup
-----
1. Copy ntttcp_sender.py and ntttcp.exe to the SENDER machine
2. Copy ntttcp_receiver.py and ntttcp.exe to the RECEIVER machine
3. Edit ntttcp_receiver.py to set MY_IP to the receiver's IP address

Running Tests
-------------
Step 1: Start the sender service on the SENDER machine:
    python ntttcp_sender.py

Step 2: Run the receiver on the RECEIVER machine:
    python ntttcp_receiver.py <SENDER_IP> [TEST_MODE]

Test Modes:
    all          - Run all tests (default)
    tcp-nonagle  - TCP with Nagle disabled only
    tcp-nagle    - TCP with Nagle enabled only
    udp          - UDP tests only

Examples:
    python ntttcp_receiver.py 192.168.1.100           # Run all tests
    python ntttcp_receiver.py 192.168.1.100 udp      # UDP only
    python ntttcp_receiver.py 192.168.1.100 tcp-nonagle

Output
------
- Console: Markdown-formatted table with real-time results
- CSV file: perf_results_YYYYMMDD_HHMMSS.csv
[perf_results_20260322_164202.csv](https://github.com/user-attachments/files/26164695/perf_results_20260322_164202.csv)

Test Duration
-------------
Each test runs for 300 seconds (Microsoft recommended for formal testing).
Total time depends on test mode:
    all:         ~2.75 hours (33 test cases)
    tcp-nonagle: ~1.1 hours (13 test cases)
    tcp-nagle:   ~1.1 hours (13 test cases)
    udp:         ~35 minutes (7 test cases)

## Summary by CodeRabbit

- **New Features**
  - Added cross-platform network performance testing for TCP and UDP scenarios.
  - Tests automatically adapt to available network interfaces, packet sizes, and CPU capacity.
  - Added coordinated sender and receiver operation for remote throughput testing.
  - Added collection of throughput, CPU usage, and packets-per-second results.
  - Added CSV and Markdown result reporting with clear failure classifications.
  - Test runs now return a failure status when any scenario does not produce valid results.

### DIFF
--- a/NetKVM/Tests/ntttcp_receiver.py
+++ b/NetKVM/Tests/ntttcp_receiver.py
@@ -1,0 +1,281 @@
+import socket
+import subprocess
+import signal
+import time
+import sys
+import csv
+import os
+import re
+import platform
+import xml.etree.ElementTree as ET
+from datetime import datetime
+
+# === Platform Detection ===
+IS_WINDOWS = platform.system() == "Windows"
+NTTTCP_CLIENT = "ntttcp.exe" if IS_WINDOWS else "ntttcp"
+
+# === Configuration ===
+SENDER_IP = sys.argv[1] if len(sys.argv) > 1 else "192.168.122.220"
+TEST_MODE = sys.argv[2] if len(sys.argv) > 2 else "all"  # all, tcp, udp
+CONTROL_PORT = 9999
+TEST_TIME = 10  # Set to 300 for formal testing (Microsoft recommended duration)
+CROSS_PLATFORM = False  # Auto-detected: True when sender and receiver are different OS
+
+def get_my_ip():
+    """Auto-detect local IP by checking which interface routes to SENDER_IP."""
+    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+        s.connect((SENDER_IP, 1))  # UDP, no actual data sent
+        return s.getsockname()[0]
+
+MY_IP = get_my_ip()
+
+def detect_interface(target_ip):
+    """Auto-detect network interface name for a given target IP via routing table."""
+    try:
+        out = subprocess.check_output(["ip", "route", "get", target_ip],
+                                      text=True, stderr=subprocess.DEVNULL)
+        m = re.search(r'dev\s+(\S+)', out)
+        return m.group(1) if m else ""
+    except Exception:
+        return ""
+
+INTERFACE_NAME = "" if IS_WINDOWS else detect_interface(SENDER_IP)
+
+# Thread counts to test (iterate like iperf_wrapper's PARALLEL_THREADS)
+THREAD_COUNTS = [1, 2, 4, 8, 16]
+
+# Generate results filename with timestamp
+TIMESTAMP = datetime.now().strftime("%Y%m%d_%H%M%S")
+CSV_FILENAME = f"ntttcp_results_{TIMESTAMP}.csv"
+LOG_FILENAME = f"ntttcp_results_{TIMESTAMP}.log"
+
+# Packet sizes to test
+PACKET_SIZES = [32, 64, 128, 256, 512, 1024, 1460, 2048, 4096, 8192, 16384, 32768, 65536]
+UDP_SIZES = [32, 64, 128, 256, 512, 1024, 1472]  # MTU-safe sizes
+
+# Build scenarios based on test mode
+def build_scenarios():
+    scenarios = []
+    if TEST_MODE in ("all", "tcp"):
+        for threads in THREAD_COUNTS:
+            for size in PACKET_SIZES:
+                scenarios.append(("TCP", size, threads))
+    if TEST_MODE in ("all", "udp"):
+        for threads in THREAD_COUNTS:
+            for size in UDP_SIZES:
+                scenarios.append(("UDP", size, threads))
+    return scenarios
+
+SCENARIOS = build_scenarios()
+
+log_file = None
+
+def log(msg=""):
+    """Print to both console and log file."""
+    print(msg)
+    if log_file:
+        log_file.write(msg + "\n")
+        log_file.flush()
+
+def wait_for_sender():
+    """Wait for sender service to come online, detect its platform."""
+    global CROSS_PLATFORM
+    log(f"Waiting for sender service at {SENDER_IP}:{CONTROL_PORT}...")
+    while True:
+        try:
+            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+                s.settimeout(3)
+                s.connect((SENDER_IP, CONTROL_PORT))
+                s.sendall(b"PLATFORM")
+                sender_platform = s.recv(64).decode().strip()
+            break
+        except (ConnectionRefusedError, socket.timeout, OSError):
+            time.sleep(2)
+    my_platform = platform.system()
+    CROSS_PLATFORM = (sender_platform != my_platform) and sender_platform != ""
+    if CROSS_PLATFORM:
+        log(f"Sender is online ({sender_platform}). Local: {my_platform}, IP: {MY_IP}")
+        log(f"Cross-platform detected, using no-sync mode.\n")
+    else:
+        log(f"Sender is online. Local IP: {MY_IP}\n")
+
+def signal_sender(proto, size, threads, duration):
+    """Signals the remote sender with test parameters and duration."""
+    try:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.settimeout(5)
+            s.connect((SENDER_IP, CONTROL_PORT))
+            no_sync = "1" if CROSS_PLATFORM else "0"
+            s.sendall(f"{proto}:{size}:{threads}:{MY_IP}:{duration}:{no_sync}".encode())
+        return True
+    except Exception:
+        return False
+
+def get_sender_results():
+    """Connect to sender and retrieve last test results (throughput:cpu:pps)."""
+    try:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.settimeout(30)
+            s.connect((SENDER_IP, CONTROL_PORT))
+            s.sendall(b"RESULTS")
+            data = s.recv(1024).decode().strip()
+            parts = data.split(':')
+            return float(parts[0]), float(parts[1]), int(parts[2])
+    except Exception:
+        return 0.0, 0.0, 0
+
+def parse_stats_xml(xml_file):
+    """Parse ntttcp XML output for throughput, CPU, and PPS."""
+    mbps, cpu, pps = 0.0, 0.0, 0
+    try:
+        tree = ET.parse(xml_file)
+        root = tree.getroot()
+        if IS_WINDOWS:
+            for elem in root.findall('throughput'):
+                if elem.get('metric') == 'mbps':
+                    mbps = round(float(elem.text), 2)
+                    break
+            for elem in root.findall('realtime'):
+                if elem.get('metric') == 's':
+                    realtime = float(elem.text)
+                    break
+            else:
+                realtime = 0
+            pkts = int(root.findtext('packets_received', '0'))
+            if pkts > 0 and realtime > 0:
+                pps = int(pkts / realtime)
+            for elem in root.findall('cpu'):
+                if elem.get('metric') == '%':
+                    cpu = round(float(elem.text), 2)
+                    break
+        else:
+            realtime = float(root.findtext('realtime', '0'))
+            for elem in root.findall('throughput'):
+                if elem.get('metric') == 'mbps':
+                    mbps = round(float(elem.text), 2)
+                    break
+            pkts = int(root.findtext('packets_received', '0'))
+            if pkts > 0 and realtime > 0:
+                pps = int(pkts / realtime)
+            user = float(root.findtext('user', '0'))
+            softirq = float(root.findtext('softirq', '0'))
+            system = float(root.findtext('system', '0'))
+            cpu = round(user + system + softirq, 2)
+    except Exception as e:
+        log(f"  XML parse error: {e}")
+    return mbps, cpu, pps
+
+# Main Execution
+start_time = datetime.now()
+wait_for_sender()
+log_file = open(LOG_FILENAME, mode='w')
+with open(CSV_FILENAME, mode='w', newline='') as csv_file:
+    csv_writer = csv.writer(csv_file)
+    csv_writer.writerow(["Protocol", "Size(B)", "Threads",
+                         "Recv_Thr(Mbps)", "Recv_PPS", "Recv_CPU(%)",
+                         "Send_Thr(Mbps)", "Send_PPS", "Send_CPU(%)"])
+
+    log(f"### NetKVM Performance Report (Mode: {TEST_MODE}, Duration: {TEST_TIME}s)")
+    log(f"### Results saved to: {CSV_FILENAME}")
+    hdr = "| {:8} | {:7} | {:4} | {:>10} | {:>7} | {:>6} | {:>10} | {:>7} | {:>6} |"
+    sep = "| {:8} | {:7} | {:4} | {:>10} | {:>7} | {:>6} | {:>10} | {:>7} | {:>6} |"
+    log(hdr.format("Protocol", "Size(B)", "Thr",
+                    "R_Mbps", "R_PPS", "R_CPU%", "S_Mbps", "S_PPS", "S_CPU%"))
+    log(sep.format(":---", ":---", ":--", ":---", ":---", ":---", ":---", ":---", ":---"))
+
+    def kill_proc(proc):
+        if IS_WINDOWS:
+            subprocess.run(["taskkill", "/f", "/pid", str(proc.pid)],
+                           capture_output=True, timeout=10)
+        else:
+            os.killpg(os.getpgid(proc.pid), signal.SIGKILL)
+
+    timeout_s = TEST_TIME + 45
+    failed = 0
+    for idx, (proto, size, threads) in enumerate(SCENARIOS):
+        try:
+            is_udp = "-u" if proto == "UDP" else ""
+            recv_l = max(size, 65536) if proto == "UDP" else size
+            xml_file = f"ntttcp_recv_{os.getpid()}.xml"
+            if os.path.exists(xml_file):
+                os.remove(xml_file)
+            if idx > 0:
+                time.sleep(2)
+
+            if IS_WINDOWS:
+                timeout_ms = (TEST_TIME + 30) * 1000
+                rcv_cmd = [NTTTCP_CLIENT, "-r"]
+                if is_udp:
+                    rcv_cmd.append(is_udp)
+                rcv_cmd += ["-m", f"{threads},*,{MY_IP}", "-l", str(recv_l),
+                            "-t", str(TEST_TIME), "-to", str(timeout_ms)]
+                if CROSS_PLATFORM:
+                    rcv_cmd.append("-ns")
+                rcv_cmd += ["-xml", xml_file]
+            else:
+                rcv_cmd = f"{NTTTCP_CLIENT} -r {is_udp} -b {recv_l} -P {threads} -N"
+                if INTERFACE_NAME:
+                    rcv_cmd += f" --show-nic-packets {INTERFACE_NAME}"
+                rcv_cmd += f" -x {xml_file} -t {TEST_TIME}"
+
+            if IS_WINDOWS:
+                proc = subprocess.Popen(rcv_cmd, stdout=subprocess.PIPE,
+                                        stderr=subprocess.PIPE, universal_newlines=True)
+            else:
+                proc = subprocess.Popen(rcv_cmd, shell=True, stdout=subprocess.PIPE,
+                                        stderr=subprocess.PIPE, universal_newlines=True,
+                                        preexec_fn=os.setsid)
+            time.sleep(2)
+
+            if signal_sender(proto, size, threads, TEST_TIME):
+                try:
+                    stdout, stderr = proc.communicate(timeout=timeout_s)
+                except subprocess.TimeoutExpired:
+                    kill_proc(proc)
+                    stdout, stderr = proc.communicate()
+                    log(f"  WARNING: ntttcp timed out")
+
+                if os.path.exists(xml_file):
+                    r_thr, r_cpu, r_pps = parse_stats_xml(xml_file)
+                else:
+                    log(f"  WARNING: XML not generated. stderr: {stderr.strip()}")
+                    r_thr, r_cpu, r_pps = 0.0, 0.0, 0
+
+                s_thr, s_cpu, s_pps = get_sender_results()
+
+                row = "| {:8} | {:7} | {:4} | {:>10} | {:>7} | {:>6} | {:>10} | {:>7} | {:>6} |"
+                log(row.format(proto, size, threads,
+                               r_thr, r_pps, f"{r_cpu}%", s_thr, s_pps, f"{s_cpu}%"))
+                csv_writer.writerow([proto, size, threads,
+                                     r_thr, r_pps, r_cpu, s_thr, s_pps, s_cpu])
+                csv_file.flush()
+            else:
+                log(f"| ERROR: Sender signal failed for {proto} {size}B |")
+                kill_proc(proc)
+                proc.communicate()
+                failed += 1
+        except Exception as e:
+            log(f"| ERROR: {proto} {size}B {threads}T failed: {e} |")
+            failed += 1
+            try:
+                kill_proc(proc)
+                proc.communicate()
+            except Exception:
+                pass
+
+    # Signal sender to exit
+    try:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.settimeout(5)
+            s.connect((SENDER_IP, CONTROL_PORT))
+            s.sendall(b"QUIT")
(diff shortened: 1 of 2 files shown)


NetKVM/Tests/ntttcp_receiver.py
```diff
@@ -0,0 +1,113 @@
...
+NUM_CPUS = os.cpu_count()
+AUTO_THREADS = NUM_CPUS * 2
+
+# Generate results filename with timestamp
+TIMESTAMP = datetime.now().strftime("%Y%m%d_%H%M%S")
+CSV_FILENAME = f"perf_results_{TIMESTAMP}.csv"
+
+# Packet sizes to test
+PACKET_SIZES = [32, 64, 128, 256, 512, 1024, 1460, 2048, 4096, 8192, 16384, 32768, 65536]
+UDP_SIZES = [32, 64, 128, 256, 512, 1024, 1472]  # MTU-safe sizes
```
YanVugenfirer: Don't you want to test USO? https://learn.microsoft.com/en-us/windows-hardware/drivers/network/udp-segmentation-offload-uso-

heywji:
@zjmletang Hi Jianming,

  Thanks for the great patch! Using ntttcp is indeed the approach highly recommended by Microsoft. I have reviewed the patch and have a few comments and suggestions:

##  1. Execution Issue
  I couldn't run it successfully on my end. The command prompts (CMD windows) and the CSV file were created, but there was no actual test data/content written to the CSV file.  

download from: https://github.com/microsoft/ntttcp/releases/download/v5.40/ntttcp.exe

image

##  2. Code Implementations
   * Parsing compatibility: Besides the issues pointed out by AI, the `parse_stats` function doesn't seem to capture the output information correctly across all versions of ntttcp.
   * Logging: It would be better to implement streaming logs rather than just writing everything to the CSV file at the very end. (Also, as AI mentioned, calling `datetime.now()`  twice isn't idel. )
   * Process cleanup: Just using proc.kill() doesn't seem robust enough. It might not cleanly terminate all related processes.

##  3. Test Iterations
  In our internal tests using netperf (TCP_STREAM, TCP_MAERTS, TCP_RR), we typically run the test multi-times to get an average. Your script appears to run the test only once, which might not be sufficient. I'm not entirely sure if Microsoft implements an internal averaging mechanism in ntttcp over a long test duration. 

Refer: 
 * https://github.com/autotest/tp-qemu/blob/master/generic/tests/netperf.py
 * https://github.com/autotest/tp-qemu/blob/master/provider/netperf_base.py
 * https://github.com/autotest/tp-qemu/blob/master/generic/tests/cfg/netperf.cfg

##  4. Automation Topology Thoughts
  Our internal testing matrix runs from a Windows VM (netperf-win) to the Linux Host (netperf-linux), which makes automation relatively simple. Your approach requires running tests between two VMs across two hosts. How do we handle network latency variations and manage exceptions in this setup? 

Plus: I noticed there is a Linux version of ntttcp available. If we could use that to replicate our internal "VM to Host" topology, that would be fantastic.

Refer:
 * https://github.com/microsoft/ntttcp
 * https://github.com/microsoft/ntttcp-for-linux

zjmletang:
@heywji Thank you for your suggestions.  They are all very helpful. I will change it to the vm-host topology. In addition, I will incorporate the other suggestions you mentioned in version 2.

zjmletang:
Sorry for the delay. Based on the review feedback, here is V2 — which is essentially a full rewrite rather than an incremental fix.

Automated cross-platform network throughput testing for NetKVM virtio-net driver using Microsoft's ntttcp.

- **ntttcp_sender.py** — Runs on the sending machine, auto-exits when receiver finishes
- **ntttcp_receiver.py** — Runs on the receiving machine, orchestrates all tests and writes CSV/log reports

Cross-platform mode (Linux ↔ Windows) is **auto-detected** at startup.

## V2 Changes (vs V1)

V2 is a major redesign. Key differences:

| # | Area | V1 | V2 |
|---|------|----|----|
| 1 | Platform | Windows-only | Cross-platform (Linux ↔ Windows, auto-detected) |
| 2 | Linux ntttcp | N/A | Full integration: `-b`, `-P`, `-N`, `--show-nic-packets`, `ip route get` auto-detect interface |
| 3 | Test matrix | `tcp-nagle` / `tcp-nonagle` / `udp` | `tcp` / `udp` only — simplified for manageable runtime |
| 4 | CPU metric | Windows `<cpu metric="%">` only | Linux: `user + system + softirq`; aligned with Windows total CPU% |
| 5 | PPS | Not extracted | `packets_sent`/`packets_received` from XML, both platforms |
| 6 | Output | CSV only | CSV + log (full console markdown table) |
| 7 | Duration | Hardcoded 300s | `TEST_TIME = 10` for iteration, comment recommends 300s for formal runs |


---

## Prerequisites

- Python 3.6+ on both machines
- `ntttcp.exe` (Windows) or `ntttcp` (Linux) in the working directory or PATH
- Windows Firewall must allow ntttcp (add rule or disable for testing)
- Control port `9999/TCP` and data ports `5001+/TCP` must be reachable between machines

## Usage

**Step 1:** Start sender (on Windows use `python`, on Linux use `python3`):

```bash
python3 ntttcp_sender.py
```

**Step 2:** Run receiver:

```bash
python3 ntttcp_receiver.py <SENDER_IP> [MODE]
```

MODE: `tcp` (65 tests), `udp` (35 tests), or `all` (100 tests). Default: `all`.

## Configuration

Edit constants at the top of `ntttcp_receiver.py`:

```python
TEST_TIME = 10          # Seconds per test (use 300 for formal testing)
THREAD_COUNTS = [1, 2, 4, 8, 16]
PACKET_SIZES = [32, 64, 128, 256, 512, 1024, 1460, 2048, 4096, 8192, 16384, 32768, 65536]
UDP_SIZES = [32, 64, 128, 256, 512, 1024, 1472]
```

### Estimated Duration

| Mode | Tests | @ 10s each | @ 300s each |
|------|-------|------------|-------------|
| tcp  | 65    | ~15 min    | ~6.5 hours  |
| udp  | 35    | ~9 min     | ~3.5 hours  |
| all  | 100   | ~25 min    | ~10 hours   |

## Output

Each run produces two files in the working directory:

- `ntttcp_results_YYYYMMDD_HHMMSS.csv` — machine-readable results
- `ntttcp_results_YYYYMMDD_HHMMSS.log` — console output (markdown table)

Sample console output:
```
| Protocol | Size(B) | Thr  |     R_Mbps |   R_PPS | R_CPU% |     S_Mbps |   S_PPS | S_CPU% |
| TCP      |   65536 |    4 |  16009.19  |   26962 | 100.0% |  16018.42  |   26966 |   2.2% |
```

## Control Protocol (port 9999)

| Message | Direction | Description |
|---------|-----------|-------------|
| `PLATFORM` | recv → send | Query sender OS (returns `Windows` or `Linux`) |
| `TCP:1024:4:IP:10:0` | recv → send | Start test: `proto:size:threads:target_ip:duration:no_sync` |
| `RESULTS` | recv → send | Request last test results (returns `mbps:cpu:pps`) |
| `QUIT` | recv → send | Signal sender to exit (sent after all tests complete) |

NetKVM/Tests/ntttcp_receiver.py
```diff
@@ -0,0 +1,281 @@
...
+import platform
+import xml.etree.ElementTree as ET
+from datetime import datetime
+
+# === Platform Detection ===
+IS_WINDOWS = platform.system() == "Windows"
+NTTTCP_CLIENT = "ntttcp.exe" if IS_WINDOWS else "ntttcp"
+
+# === Configuration ===
+SENDER_IP = sys.argv[1] if len(sys.argv) > 1 else "192.168.122.220"
```
satishnaidu77, line 18: get_my_ip() and detect_interface() run at import time with no if __name__ == "__main__" guard (sender has one at L171, receiver doesn't). Running with no args silently hangs trying to route to the hardcoded IP. Should require the argument and print usage on missing input.

NetKVM/Tests/ntttcp_receiver.py
```diff
@@ -0,0 +1,281 @@
...
+                                      text=True, stderr=subprocess.DEVNULL)
+        m = re.search(r'dev\s+(\S+)', out)
+        return m.group(1) if m else ""
+    except Exception:
+        return ""
+
+INTERFACE_NAME = "" if IS_WINDOWS else detect_interface(SENDER_IP)
+
+# Thread counts to test (iterate like iperf_wrapper's PARALLEL_THREADS)
+THREAD_COUNTS = [1, 2, 4, 8, 16]
```
satishnaidu77, line 45: Hardcoded regardless of VM size. On 4 vCPU guest: 54% of TCP rows and 60% of UDP rows have Send_CPU > 95%. At 8T and 16T, throughput is CPU-limited, not driver-limited. Should cap at vCPU count or make configurable.

NetKVM/Tests/ntttcp_receiver.py
```diff
@@ -0,0 +1,281 @@
...
+    print(msg)
+    if log_file:
+        log_file.write(msg + "\n")
+        log_file.flush()
+
+def wait_for_sender():
+    """Wait for sender service to come online, detect its platform."""
+    global CROSS_PLATFORM
+    log(f"Waiting for sender service at {SENDER_IP}:{CONTROL_PORT}...")
+    while True:
```
satishnaidu77, line 84: Infinite retry loop with no maximum. If sender is unreachable, hangs forever. Hung 5m26s in test run 3 until SIGINT. Add a deadline or max retry count and exit non-zero if exceeded.

NetKVM/Tests/ntttcp_receiver.py
```diff
@@ -0,0 +1,281 @@
...
+    try:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.settimeout(30)
+            s.connect((SENDER_IP, CONTROL_PORT))
+            s.sendall(b"RESULTS")
+            data = s.recv(1024).decode().strip()
+            parts = data.split(':')
+            return float(parts[0]), float(parts[1]), int(parts[2])
+    except Exception:
+        return 0.0, 0.0, 0
```
satishnaidu77, line 125: Three independent zero-return paths feed the same CSV writer with no validation. parse_stats_xml() inits to zeros (L129) and catches all exceptions (L165). get_sender_results() returns 0.0, 0.0, 0 on any exception (L124). Neither file checks proc.returncode. Result: zeros written as valid measurement rows, script exits 0. Confirmed: Recv_PPS=0 on all 100 rows, exit code 0. A test suite that can't fail is worse than no test suite.

NetKVM/Tests/ntttcp_receiver.py
```diff
@@ -0,0 +1,281 @@
...
+                if elem.get('metric') == 'mbps':
+                    mbps = round(float(elem.text), 2)
+                    break
+            for elem in root.findall('realtime'):
+                if elem.get('metric') == 's':
+                    realtime = float(elem.text)
+                    break
+            else:
+                realtime = 0
+            pkts = int(root.findtext('packets_received', '0'))
```
satishnaidu77, line 144: Receiver reads packets_received (L144, L157), sender reads packets_sent (L46, L59). Linux ntttcp v5.33 doesn't populate packets_received in its XML — Recv_PPS=0 on all 100 rows. Windows ntttcp v5.40 populates packets_sent correctly (26K–5.9M). Two separate parsers reading different fields for the same metric. Should be one shared parser that tries both fields with a fallback.

NetKVM/Tests/ntttcp_receiver.py
```diff
@@ -0,0 +1,281 @@
...
+        else:
+            realtime = float(root.findtext('realtime', '0'))
+            for elem in root.findall('throughput'):
+                if elem.get('metric') == 'mbps':
+                    mbps = round(float(elem.text), 2)
+                    break
+            pkts = int(root.findtext('packets_received', '0'))
+            if pkts > 0 and realtime > 0:
+                pps = int(pkts / realtime)
+            user = float(root.findtext('user', '0'))
```
satishnaidu77, line 160: Recv_CPU(%) and Send_CPU(%) use completely different measurement methods placed in the same CSV columns. Linux: user + system + softirq from /proc/stat across 64 cores. Windows: ntttcp-reported single % on 4 vCPU. Mean Recv=18.33% vs Send=86.91%. Not comparable. At minimum, use distinct column names to prevent misleading downstream analysis.

NetKVM/Tests/ntttcp_receiver.py
```diff
@@ -0,0 +1,281 @@
...
+            recv_l = max(size, 65536) if proto == "UDP" else size
+            xml_file = f"ntttcp_recv_{os.getpid()}.xml"
+            if os.path.exists(xml_file):
+                os.remove(xml_file)
+            if idx > 0:
+                time.sleep(2)
+
+            if IS_WINDOWS:
+                timeout_ms = (TEST_TIME + 30) * 1000
+                rcv_cmd = [NTTTCP_CLIENT, "-r"]
```
satishnaidu77, line 207: Receiver always -r, sender always -s. All 100 rows are guest(sender)→host(receiver). No reverse direction. NetKVM has separate TX and RX code paths — different virtio queues, different interrupt handling. A driver perf test covering one direction only misses half the driver

NetKVM/Tests/ntttcp_receiver.py
```diff
@@ -0,0 +1,281 @@
...
+                rcv_cmd += f" -x {xml_file} -t {TEST_TIME}"
+
+            if IS_WINDOWS:
+                proc = subprocess.Popen(rcv_cmd, stdout=subprocess.PIPE,
+                                        stderr=subprocess.PIPE, universal_newlines=True)
+            else:
+                proc = subprocess.Popen(rcv_cmd, shell=True, stdout=subprocess.PIPE,
+                                        stderr=subprocess.PIPE, universal_newlines=True,
+                                        preexec_fn=os.setsid)
+            time.sleep(2)
```
satishnaidu77, line 228: In cross-platform mode, -ns means independent timers. Receiver starts ntttcp, sleeps 2s, then signals sender. sleep(2) is a guess, not a synchronization mechanism. Sender can still be running when receiver finishes. 2/65 tests timed out in our run. Fix: sender should ACK when ntttcp is listening, receiver waits for ACK before starting its own timer.

NetKVM/Tests/ntttcp_receiver.py
```diff
@@ -0,0 +1,281 @@
...
+
+            if signal_sender(proto, size, threads, TEST_TIME):
+                try:
+                    stdout, stderr = proc.communicate(timeout=timeout_s)
```
satishnaidu77, line 232: proc.returncode never checked in either file — grep -n 'returncode' returns 0 matches in both. A failed ntttcp (segfault, bad args, bind failure) exits non-zero but code proceeds to parse XML and write CSV, produces zero-filled rows that look like valid data. Check proc.returncode != 0 after every communicate()

NetKVM/Tests/ntttcp_sender.py
```diff
@@ -0,0 +1,172 @@
...
+                                    stderr=subprocess.PIPE)
+            try:
+                proc.communicate(timeout=timeout_s)
+            except subprocess.TimeoutExpired:
+                subprocess.run(["taskkill", "/f", "/pid", str(proc.pid)],
+                               capture_output=True, timeout=10)
+                proc.communicate()
+                print(f"[!] WARNING: ntttcp timed out after {duration}s + 60s margin")
+        else:
+            proc = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE,
```
satishnaidu77, line 87: cmd is an f-string built from unsanitized socket input (L156) — shell=True here means full command injection. echo 'TCP:64:1:$(whoami>/tmp/pwned):10:0' | nc <guest> 9999 executes arbitrary commands. Windows path (L77) correctly uses list + no shell=True. Do the same here: build cmd as a list, drop shell=True.

NetKVM/Tests/ntttcp_sender.py
```diff
@@ -0,0 +1,172 @@
...
+        run_done.set()
+
+def start_sender_service():
+    global tests_run
+    PORT = 9999
+    start_time = datetime.now()
+    xml_file = f"ntttcp_send_{os.getpid()}.xml"
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        s.bind(('0.0.0.0', PORT))
```
satishnaidu77, line 113: Service binds all interfaces — any host on the network can connect to port 9999. No authentication, no IP filtering (addr from s.accept() at L117 is never checked). Bind to the test interface IP

NetKVM/Tests/ntttcp_sender.py
```diff
@@ -0,0 +1,172 @@
...
+
+                if data == "PLATFORM":
+                    conn.sendall(platform.system().encode())
+                    continue
+
+                parts = data.split(':')
+                if len(parts) < 5:
+                    continue
+
+                proto, size, threads, target_ip, duration = parts[:5]
```
satishnaidu77, line 141: Five fields unpacked from unauthenticated socket input with zero validation. No type checks, no range checks, no whitelist. int(duration) at L75/L146 crashes on non-numeric input. proto is never checked against TCP/UDP. target_ip is never validated with ipaddress.ip_address(). All five values are used directly in command construction.

NetKVM/Tests/ntttcp_sender.py
```diff
@@ -0,0 +1,172 @@
...
+                else:
+                    nic = detect_interface(target_ip)
+                    cmd = f"{NTTTCP_CLIENT} -s{target_ip} {is_udp} -b {size} -P {threads} -n 1 -N"
+                    if nic:
+                        cmd += f" --show-nic-packets {nic}"
+                    cmd += f" -t {duration} -x {xml_file}"
+
+                print(f"[!] Executing: {cmd}")
+                tests_run += 1
+                run_done.clear()
```
satishnaidu77, line 163: Single XML filename for all tests. run_done.clear() at L163 doesn't wait for previous thread to finish. Normal receiver serializes via RESULTS → run_done.wait(), but direct socket access can trigger parallel runs sharing one file. Use a unique filename per test or add run_done.wait() before run_done.clear().

satishnaidu77 (changes requested):

zjmletang:
@satishnaidu77 Thanks for your review. I will analyze each comment over the next few days

NetKVM/Tests/ntttcp_receiver.py
zjmletang: Fixed. All receiver logic now runs inside `main()` behind an `if __name__ == "__main__"` guard, matching the sender's structure.

NetKVM/Tests/ntttcp_receiver.py
zjmletang: Fixed, plus one extra case your data surfaced. The receiver's default thread matrix is now capped at the local CPU count
[Optional] If you'd prefer folding the sender's CPU count into the control protocol (cap at the smaller of the two) or a CLI option, we're happy to follow up.

NetKVM/Tests/ntttcp_receiver.py
zjmletang: Fixed. `wait_for_sender()` now has a 300-second wall-clock deadline;

NetKVM/Tests/ntttcp_receiver.py
zjmletang: fixed
1. Every CSV row now carries a Status column: `OK` / `FAILED(rc=)` / `FAILED(no xml)` / `FAILED(no data)` / `FAILED(no sender data)` — the fourth exists precisely to catch the zero-masquerade path you described.
2. `get_sender_results()` returns None on failure instead of a zero tuple, so "measured zeros" and "nothing fetched" are explicitly distinct.
3. The script exits `1 if failed > 0 else 0` — the exit code always matches the CSV.

NetKVM/Tests/ntttcp_sender.py
zjmletang: fixed

NetKVM/Tests/ntttcp_receiver.py
zjmletang: I think both directions already covered: the two scripts are role-symmetric, and testing the reverse direction needs no code changes — just swap which machine runs which script (the guest runs receiver.py, the host runs sender.py) and run again

NetKVM/Tests/ntttcp_receiver.py
zjmletang: Worth noting, my original intent was for users to adjust the thread counts themselves by editing the script — how many threads to measure is something the user cares about — so I deliberately kept it simple rather than adding a CLI option or protocol negotiation

NetKVM/Tests/ntttcp_receiver.py
zjmletang: fixed as you suggested, two layers: the receiver polls its local data ports (netstat/ss, PID-matched) and signals the sender only once ntttcp is really listening; the sender ACKs after validation, before spawning its ntttcp — the receiver uses that ACK as its alignment point.

NetKVM/Tests/ntttcp_receiver.py
zjmletang: Fixed.

NetKVM/Tests/ntttcp_sender.py
zjmletang: Fixed exactly as you suggested

NetKVM/Tests/ntttcp_sender.py
zjmletang: fiexed

NetKVM/Tests/ntttcp_receiver.py
zjmletang: My intent with these two columns is simply to report the CPU usage on each test machine - same pattern as R_Mbps / S_Mbps: each side measures its own machine, and the numbers naturally differ when the machines differ in size. Checking both tools' sources, they are in fact the same kind of metric (system-wide average utilization): the Windows value is a per-CPU average from NtQuerySystemInformation (not ntttcp's own process CPU%), and the Linux /proc/stat categories are mathematically equivalent to a per-CPU average. Your comment did surface one real fix: the Linux formula is now 100-idle (covers hard-irq), with both formulas stated in the report header. That said, I'd like to understand which specific scenario you find not comparable. If it's comparing the two columns horizontally (86.91% vs 18.33%): a cross-machine comparison is indeed meaningless when core counts differ; the intended usage is vertical - same column, across rows.

NetKVM/Tests/ntttcp_receiver.py
zjmletang: Direct answer first: the Linux ntttcp you tested (5.33-linux, i.e. ntttcp-for-linux) does populate packets_received — provided the command line carries `--show-nic-packets <iface>`. This works in my local runs: same old code, same binary, A/B on interface detection only — detection OK → Recv_PPS normal (TCP 89,541 / UDP 139,357); detection failed → Recv_PPS = 0 while Send_PPS stays populated, exit 0, matching your observation exactly. Same binary, opposite results: not a version issue. The source confirms it (oscounter.c): packets_* read the sysfs counters of the given interface and return 0 when the interface name is empty.

So why were all 100 rows zero in your run? Most likely interface detection failed and the old code silently dropped the flag (it resolved `dev <iface>` from `ip route get <sender_ip>` and only appended the flag on success, with no output on failure). The easiest way to find out: rerun with the fixed scripts — on detection failure they now print a WARNING with the raw `ip route get` output (or the exception), visible on the first screen. If the detection turns out fine yet the field is still 0, please share the XML and I'll dig into the version difference.

I didn't adopt the fallback parser, for three reasons. First, both fields share the same sysfs gate: without the flag both are 0, so switching fields changes nothing; with the flag both are populated and the fallback never triggers. Second, R_PPS and Send_PPS are direction-specific metrics — each machine reports its own direction — so the two parsers reading different fields is the intended role semantics, not an inconsistency. Third, where only one side is populated, packets_sent on the receiver would count local sends (mostly ACKs on TCP) — a misleading number is worse than an explicit 0. Instead, the 0 stays explicit and the diagnostics above explain why it is 0.

zjmletang:
@satishnaidu77 I’ve submitted a revised version based on your suggestions. There are still a few comments that I haven’t marked as resolved, as I’d like to discuss them with you further.

NetKVM/Tests/ntttcp_sender.py
zjmletang: You're right — it was hard-bound. To be precise, this is a partial adoption: the default is still 0.0.0.0, and binding a specific IP is now an option:

```
python ntttcp_sender.py [bind_ip]
```

(no argument = listens on all interfaces as before; the startup banner prints what it bound to).

I kept the loose default following the convention of perf-test servers: iperf3's server also listens on all interfaces unless you pass `-B/--bind`; classic netserver doesn't even have a bind-address option (its man page lists only -p/-n/-h), so it always listens on all interfaces; and the netperf workflow referenced above in this thread (autotest/tp-qemu netperf.py) starts netserver on the guest with zero arguments. With the other fixes in place (shell-free commands + input validation), a stray connection can at most trigger a performance test or get an ERR reply — no more capability than an open iperf3 server.

PR #1553, opened 2026-03-27
viogpudo: fix hardware cursor when HWCursor=1

lbwanghr:
When HWCursor=1, three issues prevent hardware cursor from working correctly:

**1. MaskedColor declared but not handled → double cursor**

`PointerCaps.MaskedColor = 1` tells Windows the driver supports MaskedColor format, but `UpdateCursor` only handles `Flags.Color`. Windows sends I-Beam, text selection, and other cursors as MaskedColor, which falls through to `STATUS_UNSUCCESSFUL`. This causes Windows to fall back to software cursor rendering while the previous hardware cursor shape remains visible — resulting in a double cursor.

Fix: Remove `MaskedColor` from `PointerCaps`. Windows then converts all cursors to Color (A8R8G8B8) with proper alpha channel, which `UpdateCursor` already handles correctly via `BltBits`.

**2. SetPointerPosition ignores Visible flag → cursor stuck at (0,0)**

`resource_id` is always set to `m_pCursorBuf->GetId()` regardless of the `Visible` flag. When `Visible=FALSE`, the cursor should be hidden per VirtIO GPU spec ("If resource_id is set to 0, the cursor is hidden."), but instead it moves to position (0,0) and remains visible.

Fix: Set `resource_id = 0` when `Visible=FALSE`, and `resource_id = m_pCursorBuf->GetId()` when `Visible=TRUE`.

**3. HWCursor defaults to 0**

With the above two fixes, hardware cursor works correctly. Change the default to 1 to enable it out of the box.

**Testing:**
- Tested on Windows 11 ARM64 (26200.8037) with VirtIO GPU 2D
- Standard arrow, I-Beam, resize handles, web custom cursors (CSS cursor) all render correctly
- Cursor hide/show transitions work properly (e.g., fullscreen video, text input)
- No double cursor artifacts

Signed-off-by: lonefondness.com <support@lonefondness.com>

### DIFF
--- a/viogpu/viogpudo/viogpudo.inx
+++ b/viogpu/viogpudo/viogpudo.inx
@@ -67,7 +67,7 @@ HKR,,EventMessageFile,%REG_EXPAND_SZ%,"%%SystemRoot%%\System32\IoLogMsg.dll"
 HKR,,TypesSupported,%REG_DWORD%,7
 
 [VioGpuDod_DeviceSettings]
-HKR,, HWCursor,                    %REG_DWORD%, 0
+HKR,, HWCursor,                    %REG_DWORD%, 1
 HKR,, FlexResolution,              %REG_DWORD%, 1
 HKR,, UsePhysicalMemory,           %REG_DWORD%, 0
 HKR,, UsePresentProgress,          %REG_DWORD%, 0
--- a/viogpu/viogpudo/viogpudo.cpp
+++ b/viogpu/viogpudo/viogpudo.cpp
@@ -482,7 +482,6 @@ NTSTATUS VioGpuDod::QueryAdapterInfo(_In_ CONST DXGKARG_QUERYADAPTERINFO *pQuery
                     pDriverCaps->MaxPointerWidth = POINTER_SIZE;
                     pDriverCaps->MaxPointerHeight = POINTER_SIZE;
                     pDriverCaps->PointerCaps.Color = 1;
-                    pDriverCaps->PointerCaps.MaskedColor = 1;
                 }
                 pDriverCaps->SupportNonVGA = IsVgaDevice();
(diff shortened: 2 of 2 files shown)


viogpu/viogpudo/viogpudo.inx
```diff
@@ -67,7 +67,7 @@ HKR,,EventMessageFile,%REG_EXPAND_SZ%,"%%SystemRoot%%\System32\IoLogMsg.dll"
 HKR,,TypesSupported,%REG_DWORD%,7
 
 [VioGpuDod_DeviceSettings]
-HKR,, HWCursor,                    %REG_DWORD%, 0
```
kostyanf14, line 70: Please move this to a separate commit; it does not fix the issue, but enables the usage of the HW cursor

kostyanf14:
ok to test

viogpu/viogpudo/viogpudo.inx
lbwanghr: Thanks for the review! Separated into two commits: the first one fixes the rendering and visibility bugs, the second one enables HWCursor by default.

YanVugenfirer:
@6-dehan Please test , the binary was built by CI

6-dehan:
Hi @YanVugenfirer @lbwanghr, I tested this issue on win11_x86 VM, but there is wrong mouse problem. The cursor is no longer an 'arrow'; it has become an orange loading state 'circle' or a non-standard mouse state.  cc @kostyanf14 

image

image

lbwanghr:
@6-dehan Thank you for presenting these incredible visuals! I’ve just successfully tested this on Windows 11 (ARM). It looks like there’s still a lot of work to be done!

PR #1555, opened 2026-03-31
viosock: Fix spec compliance issues in WSP library and socket driver

irudakov77:
Fix several Winsock SPI spec compliance and code quality issues in viosocklib and the viosock kernel driver.

viosock driver:
- VIOSockAcceptInitSocket: do not signal FD_ACCEPT on accepted sockets (semantically wrong); inherit event object and mask from listening socket per WSAEventSelect spec; unconditionally signal FD_WRITE on accepted sockets per spec

viosocklib:
- VIOSockSelect: return 0 on timeout instead of SOCKET_ERROR/WSAETIMEDOUT, per select spec
- Add parameter validation across WSP functions; return WSAEFAULT/WSAEINVAL per spec for invalid arguments
- Fix stub WSP functions returning ERROR_SUCCESS without setting lpErrno; unsupported operations now return WSAEOPNOTSUPP
- native.c: remove dead case STATUS_TIMEOUT in NtStatusToWsaError (I/O timeouts produce STATUS_IO_TIMEOUT, not STATUS_TIMEOUT); add annotations
- Adjust trace levels: VERBOSE for function entry/exit, INFORMATION for validation errors
- Add WspError() helper and WSP_ERROR() macro to unify error returns (last commit, can be dropped independently)
- Enable Win11_SDV static analysis build for viosocklib

### DIFF
--- a/viosock/buildAll.bat
+++ b/viosock/buildAll.bat
@@ -5,6 +5,8 @@ call ..\build\build.bat viosock.sln "Win10 Win11" %*
 if errorlevel 1 goto :eof
 call ..\build\build.bat sys\viosock.vcxproj "Win11_SDV" %*
 if errorlevel 1 goto :eof
+call ..\build\build.bat lib\viosocklib.vcxproj "Win11_SDV" %*
+if errorlevel 1 goto :eof
 call ..\build\build.bat wsk\wsk.vcxproj "Win11_SDV" %*
 if errorlevel 1 goto :eof
 call ..\build\build.bat viosock-wsk-test\viosock-wsk-test.vcxproj "Win11_SDV" %*
--- a/viosock/lib/native.c
+++ b/viosock/lib/native.c
@@ -34,6 +34,15 @@
 #include "native.tmh"
 #endif
 
+/*
+ * We use Native API functions here because they expose parameters that have
+ * no Win32 equivalent: NtCreateFile allows passing extended attributes (EA)
+ * with socket parameters at file creation time, and NtReadFile/NtWriteFile
+ * allow specifying a ByteOffset required for socket I/O on the underlying
+ * device. NtDeviceIoControlFile returns NTSTATUS directly, which we map to
+ * WSA error codes via NtStatusToWsaError.
+ */
+
 NTSTATUS
 NTAPI
 NtWriteFile(_In_ HANDLE FileHandle,
@@ -72,7 +81,7 @@ NtReadFile(_In_ HANDLE FileHandle,
 #define STATUS_CONNECTION_RESET           ((NTSTATUS)0xC000020DL)
 #define STATUS_LOCAL_DISCONNECT           ((NTSTATUS)0xC000013BL)
 #define STATUS_HOST_UNREACHABLE           ((NTSTATUS)0xC000023DL)
-#define STATUS_IO_TIMEOUT                 ((NTSTATUS)0xc00000b5L)
+#define STATUS_IO_TIMEOUT                 ((NTSTATUS)0xC00000B5L)
 
 INT NtStatusToWsaError(_In_ NTSTATUS Status)
 {
@@ -82,7 +91,6 @@ INT NtStatusToWsaError(_In_ NTSTATUS Status)
         case STATUS_INVALID_PARAMETER:
             wsaError = WSAEINVAL;
             break;
-        case STATUS_TIMEOUT:
         case STATUS_IO_TIMEOUT:
             wsaError = WSAETIMEDOUT;
             break;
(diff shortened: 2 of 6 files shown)


viosock/buildAll.bat
```diff
@@ -5,6 +5,8 @@ call ..\build\build.bat viosock.sln "Win10 Win11" %*
 if errorlevel 1 goto :eof
 call ..\build\build.bat sys\viosock.vcxproj "Win11_SDV" %*
 if errorlevel 1 goto :eof
+call ..\build\build.bat lib\viosocklib.vcxproj "Win11_SDV" %*
+if errorlevel 1 goto :eof
```
kostyanf14, line 9: Please move to a separate commit and explain why we need this

kostyanf14:
@irudakov77 Please fix code style and add "Signed-off by" to all commits

viosock/buildAll.bat
irudakov77: While working on viosocklib fixes I ran static analysis and found a few minor issues. Adding viosocklib to buildAll.bat ensures SDV runs on it in CI going forward, so regressions get caught early. I understand it is not required for WHQL tests.

kostyanf14:
ok to test

viosock/lib/native.c
```diff
@@ -82,7 +91,6 @@ INT NtStatusToWsaError(_In_ NTSTATUS Status)
         case STATUS_INVALID_PARAMETER:
             wsaError = WSAEINVAL;
             break;
-        case STATUS_TIMEOUT:
```
kostyanf14, line 85: Can you please explain why this is redundant? Is this an impossible case?

viosock/lib/viosocklib.c
```diff
@@ -42,6 +42,14 @@ WSPUPCALLTABLE g_UpcallTable;
...
 C_ASSERT(sizeof(VIOSOCK_PROTOCOL) / sizeof(VIOSOCK_PROTOCOL[0]) <= 0xFF);
 
+__inline int WspError(_Out_ LPINT lpErrno, _In_ INT Errno)
+{
+    *lpErrno = Errno;
+    return SOCKET_ERROR;
+}
+
+#define WSP_ERROR(err) WspError(lpErrno, (err))
+
```
kostyanf14, line 52: @YanVugenfirer 

We remove the similar logic from viostor where a define triggers setting a value not only to calculate data. So I am afraid to add it there.

viosock/lib/viosocklib.c
```diff
@@ -147,16 +170,21 @@ _Must_inspect_result_ SOCKET WSPAPI VIOSockAccept(_In_ SOCKET s,
         TraceEvents(TRACE_LEVEL_ERROR, DBG_SOCKET, "g_UpcallTable.lpWPUModifyIFSHandle failed: %u\n", *lpErrno);
         CloseHandle(hFile);
     }
-    else
+    else if (addr && addrlen)
```
kostyanf14, line 172: Why do we need this `if`? You already checked `if (addr && addrlen && *addrlen < sizeof(SOCKADDR_VM))` before

kostyanf14 (commented):
Please split your first commit into several (bug fixes, argument validation, code cleanup, log improvements, etc) 
`Adjusted log levels (INFO -> VERBOSE for func entry/exit trace records)` is a good improvement, but it is definitely not related to the commit title and doesn't fix any errors

Also about ASSERT, do you see any improvements in adding it? We compile drivers and test them only in Release, so all ASSERTs will be removed.

viosock/lib/viosocklib.c
irudakov77: These two checks serve different purposes. The spec says:

  ▎ WSAEFAULT is returned if the addrlen parameter is too small

  but also:

  ▎ If addr and/or addrlen are equal to NULL, then no information about the remote address of the accepted socket is returned.

 So checking the size only makes sense when the caller actually wants the address. If addr or addrlen is NULL we skip both the size check  and the GetPeerName call — that's exactly what the two conditions express:

  if (addr && addrlen && *addrlen < sizeof(SOCKADDR_VM))
      → WSAEFAULT  (caller wants address but buffer is too small)

  else if (addr && addrlen)
      → fill address  (caller wants address and buffer is large enough)

  If addr is NULL the first condition passes silently and the second is skipped — correct per spec.

viosock/lib/native.c
irudakov77: STATUS_TIMEOUT (0x00000102) is a success/informational code returned by wait functions like NtWaitForSingleObject when the timeout expires — it is never produced by I/O operations. I/O timeouts come back as STATUS_IO_TIMEOUT (0xC00000B5), which is already handled in the switch. So the case STATUS_TIMEOUT was dead code — it could never be reached through NtDeviceIoControlFile, NtReadFile, or NtWriteFile.

viosock/lib/viosocklib.c
irudakov77: The motivation is two-fold: it eliminates the repetitive two-line pattern (*lpErrno = X; return SOCKET_ERROR;) and, more importantly, it makes it structurally impossible to forget setting lpErrno before returning — the macro does both atomically. That said, I understand the hesitation given the viostor precedent. The macro is isolated in the last commit specifically so it can be dropped without touching any of the functional changes if the consensus is against it.

irudakov77:
> Also about ASSERT, do you see any improvements in adding it? We compile drivers and test them only in Release, so all ASSERTs will be removed.

I do use debug builds and find assertions useful in general, but in this particular case they are redundant — the parameters in question come from a trusted system component, so there is no realistic way for them to be invalid at that point. Removed.

YanVugenfirer (approved):

PR #1557, opened 2026-04-03
RHEL-137735: Fix critical event size validation bug

leidwang:
This error will prevent user from connecting to the VM after migration, and restarting the VM will cause a BSOD.

### DIFF
--- a/viosock/sys/Evt.c
+++ b/viosock/sys/Evt.c
@@ -158,9 +158,13 @@ _IRQL_requires_max_(DISPATCH_LEVEL) VOID VIOSockEvtVqProcess(IN PDEVICE_CONTEXT 
         {
             /* Drop short/long events */
 
-            if (len != sizeof(pEvt))
+            if (len != sizeof(VIRTIO_VSOCK_EVENT))
             {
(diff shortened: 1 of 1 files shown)


viosock/sys/Evt.c
```diff
@@ -158,9 +158,13 @@ _IRQL_requires_max_(DISPATCH_LEVEL) VOID VIOSockEvtVqProcess(IN PDEVICE_CONTEXT
         {
             /* Drop short/long events */
 
-            if (len != sizeof(pEvt))
+            if (len != sizeof(*pEvt))
```
YanVugenfirer: I think for readability it is better to use sizeof(VIRTIO_VSOCK_EVENT)

leidwang: Okay, done!

kostyanf14 (approved):

YanVugenfirer (approved):

leidwang:
Hi @kostyanf14  @YanVugenfirer Would you please merge this commit if there is no any further concern.  It's a small contribution from me!

YanVugenfirer:
@leidwang Sure! Thanks a lot!

PR #1529, opened 2026-02-26
netkvm: Fix incorrect virtio header access in ProcessMergedBuffers

zjmletang:
### Problem

`ProcessMergedBuffers()` reads `num_buffers` from the wrong memory location on modern virtio devices (traditional mode only).

### Root Cause

The code hardcodes `PhysicalPages[0]`:
```cpp
virtio_net_hdr_mrg_rxbuf *pMrgHeader = 
    (virtio_net_hdr_mrg_rxbuf *)pFirstBuffer->PhysicalPages[0].Virtual;
```

On modern devices (`bAnyLayout=true`) in traditional mode, `HeaderPage=1`, meaning:
- `PhysicalPages[0]` = IndirectArea (not virtio header!)
- `PhysicalPages[1]` = actual data buffer with virtio header

Note: Mergeable mode (`HeaderPage=0`) is unaffected since header is in `PhysicalPages[0]`.

### Why This Bug Went Undetected - The 64KB Coincidence

When indirect is enabled, IndirectArea is populated with `vring_desc`:

```
vring_desc (16 bytes)              virtio_net_hdr_mrg_rxbuf (12 bytes)
┌────────────────────────┐         ┌────────────────────────┐
│ addr     [0-7]  8B     │ ──────► │ flags      [0]    1B   │
│                        │         │ gso_type   [1]    1B   │
│                        │         │ hdr_len    [2-3]  2B   │
│                        │         │ gso_size   [4-5]  2B   │
│                        │         │ csum_start [6-7]  2B   │
├────────────────────────┤         ├────────────────────────┤
│ len      [8-11] 4B     │ ──────► │ csum_offset[8-9]  2B   │
│ = 0x00010000 (65536)   │         │ num_buffers[10-11]2B ◄─┼─ reads here!
├────────────────────────┤         └────────────────────────┘
│ flags    [12-13]       │
│ next     [14-15]       │
└────────────────────────┘
```

- `num_buffers` at offset 10-11 maps to **high 16 bits of `vring_desc.len`**
- With contiguous 64KB allocation (common case): `len = 65536 = 0x00010000`
- High 16 bits = `0x0001` = **1** = correct value for single-buffer packets

### When the Bug Manifests

| Condition | Result |
|-----------|--------|
| Indirect disabled | IndirectArea uninitialized → random garbage |
| Indirect enabled, but 64KB contiguous allocation fails (rare) | `len != 65536` → high 16 bits ≠ 1 |

### Fix

```cpp
[...]

PR #1561, opened 2026-04-30
NetKVM: negotiate mergeable buffers based on adapter setting

annie-li:
Only ACK VIRTIO_NET_F_MRG_RXBUF when the "MergeableBuffers" adapter setting is enabled. ACKing the feature while the adapter setting is disabled is confusing and makes this property inconsistent with other NetKVM offload settings.

In the existing implementation, "MergeableBuffers" is disabled by default, but VIRTIO_NET_F_MRG_RXBUF is still ACKed. In that mode the driver posts traditional RX descriptors while accepting the mergeable-buffer protocol. If the device returns a packet with num_buffers > 1, the packet has to be dropped after draining the remaining buffers.

Avoid that grey area by making the adapter setting control feature negotiation directly. When "MergeableBuffers" is disabled, the driver does not ACK VIRTIO_NET_F_MRG_RXBUF and uses the non-mergeable virtio-net header with the traditional RX descriptor layout.

Note: "MergeableBuffers" should ideally default to enabled, but this patch preserves the existing default of disabled and only makes feature negotiation match that setting.

### DIFF
--- a/NetKVM/Common/ParaNdis_Common.cpp
+++ b/NetKVM/Common/ParaNdis_Common.cpp
@@ -404,7 +404,7 @@ static bool ReadNicConfiguration(PARANDIS_ADAPTER *pContext, PUCHAR pNewMACAddre
                                      CheckOSNdisVersion(6, bPollModeTestOnWin11 ? 85 : 89);
 #endif
             // Allow fallback to non-mergeable buffers via registry.
-            // Setting MergeableBuffers=0 disables the feature even if device supports it.
+            // Setting MergeableBuffers=0 prevents negotiating the mergeable RX buffers feature.
             pContext->bMergeableBuffersConfigured = pConfiguration->MergeableBuffers.ulValue != 0;
 
             if (!pContext->bDoSupportPriority)
@@ -898,14 +898,12 @@ NDIS_STATUS ParaNdis_InitializeContext(PARANDIS_ADAPTER *pContext, PNDIS_RESOURC
         InitializeMAC(pContext, CurrentMAC);
         InitializeMaxMTUConfig(pContext);
 
-        // Always ACK VIRTIO_NET_F_MRG_RXBUF if suggested by the device for compatibility.
-        // Registry setting bMergeableBuffersConfigured controls whether we use small 4KB buffers
-        // or traditional large buffers.
-        pContext->bUseMergedBuffers = AckFeature(pContext, VIRTIO_NET_F_MRG_RXBUF);
-        if (!pContext->bMergeableBuffersConfigured && pContext->bUseMergedBuffers)
+        // ACK VIRTIO_NET_F_MRG_RXBUF conditionally to align with VirtIO spec.
+        pContext->bUseMergedBuffers = pContext->bMergeableBuffersConfigured &&
+                                      AckFeature(pContext, VIRTIO_NET_F_MRG_RXBUF);
+        if (!pContext->bMergeableBuffersConfigured)
         {
-            DPrintf(0,
-                    "Mergeable buffers feature ACKed, but disabled by configuration (using traditional large buffers)");
+            DPrintf(5, "Mergeable buffers disabled by configuration");
         }
         pContext->nVirtioHeaderSize = (pContext->bUseMergedBuffers) ? sizeof(virtio_net_hdr_mrg_rxbuf)
                                                                     : sizeof(virtio_net_hdr);
--- a/NetKVM/Common/ParaNdis_RX.cpp
+++ b/NetKVM/Common/ParaNdis_RX.cpp
@@ -289,8 +289,7 @@ error_exit:
 
 pRxNetDescriptor CParaNdisRX::CreateRxDescriptorOnInit()
 {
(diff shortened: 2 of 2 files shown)


ybendito:
@annie-li Annie, does this PR address some real problem with some functional flow or it is rather cosmetic change? This is exactly the same behavior regarding device feature that was earlier, before implementation of merged buffers (intentionally was not changed). With registry setting off we ack the device feature but allocate large buffers so the device will never deliver packets with num_buffers > 1.

zjmletang:
hi @annie-li 
Just some context from the original PR (#1446), in case it helps frame the discussion:

 the current "always ACK VIRTIO_NET_F_MRG_RXBUF regardless of the registry setting" shape was introduced during [#1446](https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1529) review for hypervisor compatibility reasons. Full discussion is in that PR's review thread.

On the side note in your description about defaulting MergeableBuffers to enabled — I share that view. The "default 0" was chosen because throughput comparison data wasn't yet available at the time; this can be revisited once @ybendito and @YanVugenfirer have internal performance numbers.

annie-li:
> @annie-li Annie, does this PR address some real problem with some functional flow or it is rather cosmetic change? This is exactly the same behavior regarding device feature that was earlier,

@ybendito It is not exactly the same. With mergeable buffer implementation, the driver drops packets if device returns (numBuffers>1) in the header. The original implementation didn't check numBuffers at all. I did run into issues with some buggy devices that return random numBuffers value after VIRTIO_NET_F_MRG_RXBUF is ACKed. The device should fix its own issue for sure, but it is safer to not ACK VIRTIO_NET_F_MRG_RXBUF if the driver disables it.
see https://github.com/virtio-win/kvm-guest-drivers-windows/blob/master/NetKVM/Common/ParaNdis_RX.cpp#L976

> before implementation of merged buffers (intentionally was not changed). With registry setting off we ack the device feature but allocate large buffers so the device will never deliver packets with num_buffers > 1.

Suppose all devices behave well, and never deliver packets with num_buffers > 1, then it isn't necessary for the driver to ACK VIRTIO_NET_F_MRG_RXBUF. Is there any concerns in such case to ACK VIRTIO_NET_F_MRG_RXBUF when driver disables it? or I'm missing something that you have run into before?

Last, this also aligns with other adapter settings.

ybendito:
@annie-li I agree in general. We definitely will apply that, probably a little later. Please just update the PR to solve clang-format check, there is probably some missing (or extra) EOL.
Previous version (not checking num_buffers) delivered each part of the splitted packet as a separate packet (which is also not correct).

NetKVM/Common/ParaNdis_Common.cpp
```diff
@@ -898,14 +898,12 @@ NDIS_STATUS ParaNdis_InitializeContext(PARANDIS_ADAPTER *pContext, PNDIS_RESOURC
...
-        if (!pContext->bMergeableBuffersConfigured && pContext->bUseMergedBuffers)
+        // ACK VIRTIO_NET_F_MRG_RXBUF conditionally to align with VirtIO spec.
+        pContext->bUseMergedBuffers = pContext->bMergeableBuffersConfigured &&
+                                      AckFeature(pContext, VIRTIO_NET_F_MRG_RXBUF);
+        if (!pContext->bMergeableBuffersConfigured)
         {
-            DPrintf(0,
-                    "Mergeable buffers feature ACKed, but disabled by configuration (using traditional large buffers)");
+            DPrintf(0, "Mergeable buffers disabled by configuration");
         }
```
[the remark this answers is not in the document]
ybendito: I agree with Gemini

NetKVM/Common/ParaNdis_RX.cpp
```diff
@@ -973,21 +972,6 @@ pRxNetDescriptor CParaNdisRX::ProcessMergedBuffers(pRxNetDescriptor pFirstBuffer
...
-        // retrieve a data-only fragment (which has no VirtIO header) and misinterpret it as
-        // the start of a new packet, causing VirtIO protocol desynchronization
-        DPrintf(0, "ERROR: Received merged packet (%u buffers) in traditional mode. Dropping.", numBuffers);
-
-        CollectRemainingMergeBuffers();
-        ReuseCollectedBuffers();
-
-        return NULL;
-    }
-
```
ybendito, line 990: It looks like now we do not have bMergeableBuffersConfigured in the context, a local variable where it used is enough, correct?

annie-li: `I agree with Gemini`
@ybendito Looks you were looking at V1 code, in the V2 code, I've changed the log level to 5 in V2 version.
About bMergeableBuffersConfigured, a local variable can do the work, but this requests to read the value from its registry key. This is duplicated as what ReadNicConfiguration does and is inconsistent with the structure of the source code.
This may be a little off the topic - looking at the code, bDoSupportPriority should be replaced by a local variable.

kostyanf14:
ok to test

ybendito (approved):

ybendito:
@kostyanf14 could you pls check what happens with HCK-CI?

kostyanf14:
re-run tests

YanVugenfirer (approved):

PR #1568, opened 2026-05-18
CVE-2026-46655: viosock: Validate fd_count in VIOSockSelect against overflow and empt…

irudakov77:
The previous entry check inspected only the sum of fd_count values in 32-bit arithmetic. An attacker-controlled tuple (e.g. 0xFFFFFFFE + 0x40 + 0x02) wraps the sum to a small non-zero value that passes both the zero and FD_SETSIZE bounds. VIOSockSelectCopyFds then iterates up to fd_count times and writes into pPkt->Fds[FD_SETSIZE], producing an out-of-bounds kernel write.

Validate each per-set fd_count against FD_SETSIZE before summing; the sum is then bounded by 3 * FD_SETSIZE and cannot overflow ULONG.

Also reject the empty fd_set tuple (all three counts zero) with STATUS_INVALID_PARAMETER, matching the documented WSAEINVAL for "all three descriptor parameters were null".

Resolves: CVE-2026-46655

### DIFF
--- a/viosock/sys/Device.c
+++ b/viosock/sys/Device.c
@@ -960,6 +960,7 @@ static NTSTATUS VIOSockSelect(IN WDFREQUEST Request, IN OUT size_t *pLength)
     BOOLEAN bIs32BitProcess = FALSE;
     WDF_OBJECT_ATTRIBUTES Attributes;
     PVIOSOCK_SELECT_PKT pPkt = NULL;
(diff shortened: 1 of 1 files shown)


YanVugenfirer (approved):

YanVugenfirer:
ok to test

kostyanf14:
rerun tests

YanVugenfirer (approved):

PR #1572, opened 2026-05-22
netkvm: Fix extra page leak on Enqueue failure causing TX hang

coder014:
### Summary

In `CTXVirtQueue::SubmitPacket`, when `BindToDescriptor` succeeds via the copy path (borrowing DMA pages from `m_ExtraPages` pool) but the subsequent `Enqueue` fails with `SUBMIT_NO_PLACE_IN_QUEUE`, the borrowed pages are never returned. This causes a progressive depletion of the `m_ExtraPages` pool, ultimately leading to an unrecoverable TX hang.

### Root Cause

When `m_SGL->NumberOfElements > m_VirtioSGLSize` (the NB's scatter-gather list exceeds the descriptor's direct capacity), the driver takes the copy path in `FillDescriptorSGList`:

1. `AllocateAndFillCopySGL` allocates a `CExtendedNBStorage`, borrows N pages from `m_ExtraPages`, and copies packet data into them.
2. `MapCopyDataToVirtioSGL` fills the VirtIO SGL with borrowed page physical addresses.
3. `BindToDescriptor` returns `SUCCESS`.

If `TXDescriptor->Enqueue` then fails (e.g., `FreeDescriptors < m_UsedBuffersNum`), the enqueue failure branch in `SubmitPacket` only pushes the descriptor back and reports failure — **without returning borrowed extra pages**.

The NB (with `m_ExtraNBStorage` still holding borrowed pages) is returned to the caller. In `SendMapped`, the `SUBMIT_NO_PLACE_IN_QUEUE` result causes the NB to be pushed back to the mapped queue for retry. On the next attempt, `AllocateAndFillCopySGL` overwrites `m_ExtraNBStorage` with a new allocation, permanently **leaking the old storage and its borrowed pages**.

Once enough pages leak to make `m_ExtraPages.GetCount() < NeedPages`, every subsequent `BorrowPages` call fails, returning `NO_RESOURCE`. This maps to `SUBMIT_NO_PLACE_IN_QUEUE` in the `BindToDescriptor`-failure branch (which does NOT call `KickQueueOnOverflow`), creating a permanent head-of-line blocking loop:

```
retry same NB → BorrowPages fails → NO_RESOURCE → push NB back → Kick (empty notify) → retry → ...
```

### Steps to Reproduce

1. Use a virtio-net device that does **not** negotiate VIRTIO_RING_F_INDIRECT_DESC (i.e., `bUseIndirect = false`) and does negotiate VIRTIO_NET_F_HOST_TSO4/VIRTIO_NET_F_HOST_TSO6.
2. Make sure that `Offload.Tx.LSO` property of the netkvm driver is enabled.
3. Run a TCP stress test with small application writes to trigger LSO aggregation:
   ```
   iperf3 -c <server_ip> -P 4 -i 1 -l 100 -t 60
   ```
4. The TCP stack aggregates small writes into large LSO segments, producing NBs whose `m_SGL->NumberOfElements` exceeds `m_VirtioSGLSize`, forcing the copy path.
5. Under load, `m_FreeHWBuffers` drops below the required `m_CurrVirtioSGLEntry` for a copy-path packet, triggering `Enqueue → SUBMIT_NO_PLACE_IN_QUEUE`.
6. TX hangs within a few seconds. Network connectivity is permanently lost.

### What was expected

TX queue should never hang.

### What actually happens

- `m_ExtraPages` pool progressively depletes.
- All subsequent copy-path packets fail permanently with `NO_RESOURCE`.
- The queue-head NB is retried infinitely, blocking all other NBs behind it.
- Device-side observes continuous notify signals but no descriptors in the ring.

### Fix

Add guard in `CNB::AllocateAndFillCopySGL` to prevent duplicate allocation of `m_ExtraNBStorage`.

### Notes
- The `~CNB()` destructor contains a safety-net `ReturnPages()` call, but it never triggers in this scenario because the NB is pushed back for retry (not destroyed) on `SUBMIT_NO_PLACE_IN_QUEUE`.
- The page leak via `SUBMIT_NO_PLACE_IN_QUEUE` is easily reproducible in a virtqueue using direct descriptor, where `m_UsedBuffersNum = m_CurrVirtioSGLEntry` can be greater than 1. In indirect mode, `m_UsedBuffersNum` is always 1, making the `FreeDescriptors < m_UsedBuffersNum` condition in `Enqueue` extremely unlikely (would require `m_FreeHWBuffers == 0`, which is prevented by the earlier `m_Descriptors.GetCount()` check in `SubmitPacket`), so the bug is much harder to trigger.

### DIFF
--- a/NetKVM/Common/ParaNdis_TX.cpp
+++ b/NetKVM/Common/ParaNdis_TX.cpp
@@ -1496,6 +1496,13 @@ void CNB::ReturnPages()
 
 NBMappingStatus CNB::AllocateAndFillCopySGL(ULONG ParsedHeadersLength)
 {
+    // If we have already borrowed extra pages,
+    // just return SUCCESS otherwise we will leak these pages.
+    if (m_ExtraNBStorage)
+    {
+        return NBMappingStatus::SUCCESS;
+    }
+
     // Calculate the number of pages needed for data, excluding protocol header length.
     ULONG DataLength = GetDataLength() - ParsedHeadersLength;
     ULONG Pages = (DataLength + PAGE_SIZE - 1) / PAGE_SIZE;


zjmletang:
@coder014 Thanks for identifying this issue — the bug analysis is correct.

I'd like to suggest an alternative approach for your consideration: make `AllocateAndFillCopySGL` idempotent by adding an early-return guard:

```cpp
NBMappingStatus CNB::AllocateAndFillCopySGL(ULONG ParsedHeadersLength)
{
    if (m_ExtraNBStorage)
        return NBMappingStatus::SUCCESS;
    // ... rest unchanged
}
```

Advantages:
1. Prevents the leak at its source — no redundant allocation, no pointer overwrite
2. Avoids repeated memcpy on retry (the data in borrowed pages is still valid)
3. Minimal change — only 2 lines added, no modifications needed elsewhere

coder014:
> @coder014 Thanks for identifying this issue — the bug analysis is correct.
> 
> I'd like to suggest an alternative approach for your consideration: make `AllocateAndFillCopySGL` idempotent by adding an early-return guard:
> 
> ```c++
> NBMappingStatus CNB::AllocateAndFillCopySGL(ULONG ParsedHeadersLength)
> {
>     if (m_ExtraNBStorage)
>         return NBMappingStatus::SUCCESS;
>     // ... rest unchanged
> }
> ```
> 
> Advantages:
> 
> 1. Prevents the leak at its source — no redundant allocation, no pointer overwrite
> 2. Avoids repeated memcpy on retry (the data in borrowed pages is still valid)
> 3. Minimal change — only 2 lines added, no modifications needed elsewhere

@zjmletang Thank you for providing another way to fix, it is indeed more elegant.

I have a consideration: with this approach, borrowed pages remain "locked" to the failed NB until it eventually succeeds or is destroyed. If the pool is small relative to per-packet demand (e.g., 17 pages total, 16 consumed by NB_a), a subsequent NB_b requiring the copy path will possibly see BorrowPages fail and be deferred until NB_a completes and its pages are returned. This is not a problem, in-flight `ProcessTXCompletions` will replenish the pool, but it does introduce a brief head-of-line delay under heavy copy-path pressure.

By contrast, the approach that adding `ReturnPages` in the Enqueue failure branch releases pages immediately on failure, keeping the pool available for other NBs at the cost of re-borrowing and re-copying on retry.

Both fixes address the same root cause from different angles, and each has its own shortcomings. I'm not sure which one is better.

zjmletang:
@coder014 
I think the pool pressure scenario may not apply here due to the queue ordering. When SUBMIT_NO_PLACE_IN_QUEUE happens, PushMappedNB uses InsertHeadList (LIFO), and the loop exits immediately. On the next SendMapped, PeekMappedNBL + PopMappedNB (RemoveHeadList) always picks up the same failed NB first. So no other NB should be able to reach BorrowPages before it succeeds.

Also since the entire path runs under DoWithTXLock with per-queue m_ExtraPages, I don't see how concurrent contention could occur.

If the above holds, the idempotent guard would also have a slight performance benefit on retry — skipping the redundant alloc + BorrowPages + memcpy (potentially 64KB for LSO). What do you think?

coder014:
> @coder014 I think the pool pressure scenario may not apply here due to the queue ordering. When SUBMIT_NO_PLACE_IN_QUEUE happens, PushMappedNB uses InsertHeadList (LIFO), and the loop exits immediately. On the next SendMapped, PeekMappedNBL + PopMappedNB (RemoveHeadList) always picks up the same failed NB first. So no other NB should be able to reach BorrowPages before it succeeds.
> 
> Also since the entire path runs under DoWithTXLock with per-queue m_ExtraPages, I don't see how concurrent contention could occur.
> 
> If the above holds, the idempotent guard would also have a slight performance benefit on retry — skipping the redundant alloc + BorrowPages + memcpy (potentially 64KB for LSO). What do you think?

I missed the fact that the ordering of Push-Pop sequence for MappedNB in ​​a single TX queue is LIFO. My consideration is wrong. Let me try fixing it by adding a guard in `CNB::AllocateAndFillCopySGL`.

zjmletang:
@coder014 The two commits could be squashed into one, since the second supersedes the first.

coder014:
> @coder014 The two commits could be squashed into one, since the second supersedes the first.

Thanks ;) Re-committed.

zjmletang (approved):

kostyanf14:
ok to test

zjmletang:
@coder014 The clang-format check failed. Please refer to the instructions in contributing.md to fix it.

ybendito (approved):

YanVugenfirer:
@kostyanf14 same here - HCK-CI failed

kostyanf14:
@YanVugenfirer @ybendito almost all job finished in this PR but Jenkins crashed. Need to run from the scratch. Should I run, or these restuls is enouth?

ybendito:
> @YanVugenfirer @ybendito almost all job finished in this PR but Jenkins crashed. Need to run from the scratch. Should I run, or these restuls is enouth?

I think it's enough

YanVugenfirer:
> @YanVugenfirer @ybendito almost all job finished in this PR but Jenkins crashed. Need to run from the scratch. Should I run, or these restuls is enouth?

Good enough

YanVugenfirer (approved):

YanVugenfirer:
@coder014 Thanks a lot! Merged.

PR #1577, opened 2026-06-01
viosock: Fix select correctness, event notification, and socket handle type

irudakov77:
**Driver fixes:**

- Set the WDF device type to `FILE_DEVICE_NAMED_PIPE` so that `GetFileType()` returns `FILE_TYPE_PIPE` for viosock socket handles, matching real socket providers and making `_open_osfhandle()` succeed.
- Fix `WSPSelect`: clear the output fd sets before processing so that input fd handles don't leak into the output (the IOCTL uses a shared input/output buffer).
- Fix event notification to mark all event bits atomically under a single state lock hold and call `VIOSockEventNotify()` once, so the event object and select workitem are signaled only once per batch (e.g. FD_CONNECT + FD_WRITE on connection).
- Refactoring: `VIOSockSelectGetFds*` macros and `VIOSockSelectEnqueueWorkItem` helper to reduce code duplication.

**Library fixes:**

- Implement `SO_PROTOCOL_INFOW` for `WSPGetSockOpt`.
- Fix protocol registration info: set `iVersion` to 2.

### DIFF
--- a/viosock/lib/install.c
+++ b/viosock/lib/install.c
@@ -50,7 +50,7 @@ static WSAPROTOCOL_INFO g_ProtocolInfo = {.dwServiceFlags1 = XP1_GRACEFUL_CLOSE 
                                           .dwCatalogEntryId = 0,
                                           .ProtocolChain.ChainLen = BASE_PROTOCOL,
                                           .ProtocolChain.ChainEntries = {0},
-                                          .iVersion = 0,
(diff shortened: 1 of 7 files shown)


kostyanf14:
ok to test

kostyanf14:
re-run tests

YanVugenfirer (approved):

PR #1586, opened 2026-07-01
RHEL-183405: [viostor] Validate blk_size to prevent divide-by-zero

harshapa-rh:
…de-by-zero

### DIFF
--- a/viostor/virtio_stor_hw_helper.c
+++ b/viostor/virtio_stor_hw_helper.c
@@ -625,7 +625,7 @@ RhelGetSectors(IN PVOID DeviceExtension, IN PCDB Cdb)
     return (sector.AsULong * (adaptExt->info.blk_size / SECTOR_SIZE));
 }
 
-VOID RhelGetDiskGeometry(IN PVOID DeviceExtension)
+BOOLEAN RhelGetDiskGeometry(IN PVOID DeviceExtension)
 {
     u64 cap;
     u32 v;
@@ -664,6 +664,11 @@ VOID RhelGetDiskGeometry(IN PVOID DeviceExtension)
     if (CHECKBIT(adaptExt->features, VIRTIO_BLK_F_BLK_SIZE))
     {
         virtio_get_config(&adaptExt->vdev, FIELD_OFFSET(blk_config, blk_size), &v, sizeof(v));
+        if (v == 0)
+        {
+            RhelDbgPrint(TRACE_LEVEL_ERROR, "Invalid blk_size %u from device config\n", v);
+            return FALSE;
+        }
         adaptExt->info.blk_size = v;
     }
     else
@@ -737,6 +742,7 @@ VOID RhelGetDiskGeometry(IN PVOID DeviceExtension)
         adaptExt->info.max_discard_seg = min(v, MAX_DISCARD_SEGMENTS);
         RhelDbgPrint(TRACE_LEVEL_INFORMATION, " max_discard_seg = %d\n", adaptExt->info.max_discard_seg);
     }
+    return TRUE;
 }
 
 VOID VioStorVQLock(IN PVOID DeviceExtension, IN ULONG MessageID, IN OUT PSTOR_LOCK_HANDLE LockHandle, IN BOOLEAN isr)
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -395,7 +395,11 @@ VirtIoFindAdapter(IN PVOID DeviceExtension,
         return res;
     }
 
-    RhelGetDiskGeometry(DeviceExtension);
+    if (!RhelGetDiskGeometry(DeviceExtension))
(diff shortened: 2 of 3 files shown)


YanVugenfirer:
Hi @harshapa-rh 

Please squash all the commits 

It is better to run clang-format locally first

YanVugenfirer (changes requested):
Changes requested in the previous comment

YanVugenfirer:
BTW: in PR and commits use the parent Jira key: https://redhat.atlassian.net/browse/RHEL-183405

viostor/virtio_stor_hw_helper.c
```diff
@@ -664,6 +664,11 @@ VOID RhelGetDiskGeometry(IN PVOID DeviceExtension)
     if (CHECKBIT(adaptExt->features, VIRTIO_BLK_F_BLK_SIZE))
     {
         virtio_get_config(&adaptExt->vdev, FIELD_OFFSET(blk_config, blk_size), &v, sizeof(v));
+        if (v < SECTOR_SIZE)
```
YanVugenfirer: 1. Fix the commit message. The squashed commit message should contain only one message
2. We want to validate that v is not zero. If it is zero, propagate error to initialization functions so they will fail. 
3. We probably cannot do anything useful on reconfiguration notification.

YanVugenfirer (changes requested):
1. Fix the commit message. The squashed commit message should contain only one message
2. We want to validate that v is not zero. If it is zero, propagate error to initialization functions so they will fail. 
3. We probably cannot do anything useful on reconfiguration notification.

viostor/virtio_stor_hw_helper.c
```diff
@@ -664,6 +664,11 @@ VOID RhelGetDiskGeometry(IN PVOID DeviceExtension)
     if (CHECKBIT(adaptExt->features, VIRTIO_BLK_F_BLK_SIZE))
     {
         virtio_get_config(&adaptExt->vdev, FIELD_OFFSET(blk_config, blk_size), &v, sizeof(v));
+        if (v == 0)
+        {
+            RhelDbgPrint(TRACE_LEVEL_ERROR, "Invalid blk_size %u from device config, using default\n", v);
```
YanVugenfirer: The debug print is misleading. Please remove the ", using default" part. 
The failure of the function will cause the device not to initialize. 

BTW: Did you test the fix by installing the driver? Also it is worth to return 0 on purpose from the host (use virtio-blk as secondary device in such case)

YanVugenfirer (changes requested):

viostor/virtio_stor_hw_helper.c
harshapa-rh: I fixed the debug print. I was able to install the driver but I couldn't simulate this specific test case.

YanVugenfirer:
run tests

viostor/virtio_stor_hw_helper.c
```diff
@@ -664,6 +664,11 @@ VOID RhelGetDiskGeometry(IN PVOID DeviceExtension)
     if (CHECKBIT(adaptExt->features, VIRTIO_BLK_F_BLK_SIZE))
     {
         virtio_get_config(&adaptExt->vdev, FIELD_OFFSET(blk_config, blk_size), &v, sizeof(v));
+        if (v == 0)
+        {
+            RhelDbgPrint(TRACE_LEVEL_ERROR, "Invalid blk_size %u from device config\n", v);
+            return FALSE;
+        }
```
kevmw, line 671: This early return means you get a partially updated `adaptExt`. I think this is unexpected, you always want a valid new state or the unchanged old state.

It probably doesn't matter when called from `VirtIoFindAdapter()` or `VirtIoHwReinitialize()`, where failure kills the whole device anyway, but `VirtIoConfigUpdated()` ignores the error silently. The driver would then keep partially updated values, but never notify StorPort about the changes, potentially making assumptions inconsistent between the two layers (though of course, both are now inconsistent with the device anyway).

The other, probably better option is making sure that the failure path in `VirtIoConfigUpdated()` actually leads to a disabled device, too.

kevmw:
Please update the commit message with the motivation for the fix. Is this purely theoretical or is there a way to actually end up in this situation? Please also state there why the chosen solution (fail device initialisation on error, but silently ignore the error on runtime updates) is the correct/desirable one.

harshapa-rh:
@kevmw While reproducing I noticed that standard QEMU already prevents this. It will reject these parameters. So this can either be closed as is to reduce redundancy or we can keep this for good measure after I make any suggested fixes.

@YanVugenfirer please review the discussion and decide the best course of action. Thank you.

kevmw:
Hardening can still be a good idea, so I'm not against addressing it even though it can't be reproduced with QEMU. The commit message should just point this out for clarity.

PR #1595, opened 2026-07-13
Support IOMMU in viostor driver viostor

vrozenfe:
Applying the second commit, "Enable IOMMU ...", can be postponed for downstream/WHQL-signed builds until Microsoft enforces DMAR support.


vrozenfe:
fixed gemini reported issues

PR #1596, opened 2026-07-13
RHEL-183407: [viostor] Add NULL check for dpc

harshapa-rh:


YanVugenfirer (approved):

YanVugenfirer:
run tests

kostyanf14:
run tests

vrozenfe (commented):
This code is not only unnecessary but also potentially harmful. viostor is capable of operating without a DPC by completing requests directly from the ISR routine. While the code below is purely theoretical, this failure scenario cannot occur in practice because DPC allocation is performed from a preallocated memory pool. Even if such a failure were possible, we still should not fail the hardware initialization routine because of it. 
Best. Vadim.

harshapa-rh:
@vrozenfe Yes, I was considering that as well. It was part of the reviews so I added the change based on that: [here](https://github.com/virtio-win/kvm-guest-drivers-windows/commit/e6fe0f987295ad99fe40ecaf468186153a74ca38)

What changes would you suggest? This can either be closed as is or I can add a warning instead of returning failure.

YanVugenfirer:
Closing based on the discussion

PR #1597, opened 2026-07-13
Support IOMMU in viostor driver Viostor

vrozenfe:
Applying the second commit, "Enable IOMMU ...", can be postponed for downstream/WHQL-signed builds until Microsoft enforces DMAR support.

### DIFF
--- a/viostor/viostor.inx
+++ b/viostor/viostor.inx
@@ -91,7 +91,7 @@ HKR,,TypesSupported,%REG_DWORD%,7
 [pnpsafe_pci_addreg]
 HKR, "Parameters\PnpInterface", "5", %REG_DWORD%, 0x00000001
 HKR, "Parameters", "BusType", %REG_DWORD%, 0x00000001
-HKR, "Parameters", DmaRemappingCompatible,0x00010001,0
+HKR, "Parameters", DmaRemappingCompatible,0x00010001,3
 
 [pnpsafe_pci_addreg_msix]
 HKR, "Interrupt Management",, 0x00000010
--- a/viostor/virtio_stor_hw_helper.c
+++ b/viostor/virtio_stor_hw_helper.c
@@ -39,9 +39,13 @@
 
 #define SET_VA_PA()                                                                                                    \
     {                                                                                                                  \
-        ULONG len;                                                                                                     \
-        va = adaptExt->indirect ? srbExt->desc : NULL;                                                                 \
-        pa = va ? StorPortGetPhysicalAddress(DeviceExtension, NULL, va, &len).QuadPart : 0;                            \
+        va = NULL;                                                                                                     \
+        pa = 0;                                                                                                        \
+        if (adaptExt->indirect)                                                                                        \
+        {                                                                                                              \
+            va = srbExt->desc;                                                                                         \
+            pa = extPA.QuadPart + FIELD_OFFSET(SRB_EXTENSION, desc);                                                   \
+        }                                                                                                              \
     }
 
 static ULONG GetSrbQueueNumber(IN PVOID DeviceExtension, IN PSRB_TYPE Srb)
@@ -99,8 +103,21 @@ RhelDoFlush(PVOID DeviceExtension, PSRB_TYPE Srb, BOOLEAN resend, BOOLEAN bIsr)
     STOR_LOCK_HANDLE LockHandle = {0};
     struct virtqueue *vq = NULL;
     PREQUEST_LIST element;
+    STOR_PHYSICAL_ADDRESS extPA;
+
+    extPA = StorPortGetPhysicalAddress(DeviceExtension, (PSCSI_REQUEST_BLOCK)Srb, srbExt, &fragLen);
+    if (extPA.QuadPart == 0 || fragLen < sizeof(SRB_EXTENSION))
+    {
+        CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_INSUFFICIENT_RESOURCES);
+        return TRUE;
+    }
 
     SET_VA_PA();
+    if (adaptExt->indirect && pa == 0)
+    {
+        CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_INSUFFICIENT_RESOURCES);
+        return TRUE;
+    }
 
     QueueNumber = resend ? srbExt->queue_number : GetSrbQueueNumber(DeviceExtension, Srb);
     MessageId = QueueToMessageId(DeviceExtension, QueueNumber);
@@ -115,9 +132,9 @@ RhelDoFlush(PVOID DeviceExtension, PSRB_TYPE Srb, BOOLEAN resend, BOOLEAN bIsr)
     srbExt->out = 1;
     srbExt->in = 1;
 
-    srbExt->sg[0].physAddr = StorPortGetPhysicalAddress(DeviceExtension, NULL, &srbExt->vbr.out_hdr, &fragLen);
+    srbExt->sg[0].physAddr.QuadPart = extPA.QuadPart + FIELD_OFFSET(SRB_EXTENSION, vbr.out_hdr);
     srbExt->sg[0].length = sizeof(srbExt->vbr.out_hdr);
-    srbExt->sg[1].physAddr = StorPortGetPhysicalAddress(DeviceExtension, NULL, &srbExt->vbr.status, &fragLen);
+    srbExt->sg[1].physAddr.QuadPart = extPA.QuadPart + FIELD_OFFSET(SRB_EXTENSION, vbr.status);
     srbExt->sg[1].length = sizeof(srbExt->vbr.status);
 
     if (!resend)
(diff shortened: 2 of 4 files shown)


vrozenfe:
run tests

kevmw:
I'm taking a look, but as a first observation, I would really appreciate real commit messages instead of terse subject-line-only messages devoid of any real information. Review would be much easier then.

Commit messages should explain your motivation to make the change, which is often necessary context to understand why you make specific choices without having to ask you (and in two years, you'll probably have forgotten yourself), a high-level technical overview of what the commit does, and potentially also comments on why the series is structured the way it is.

So in this specific pull request, I'd expect things like "we want to enable IOMMU support because X and we envision it to be used like Y" (e.g. only for debugging or also typically in production?), then "this requires the DMA operations in the driver (or whatever) to change in the following way" and finally "actually enabling it is separate in patch 2 because we may want to apply only patch 1 in downstream and having it without actually enabling it makes sense for reason Z".

viostor/viostor.inx
```diff
@@ -91,7 +91,7 @@ HKR,,TypesSupported,%REG_DWORD%,7
 [pnpsafe_pci_addreg]
 HKR, "Parameters\PnpInterface", "5", %REG_DWORD%, 0x00000001
 HKR, "Parameters", "BusType", %REG_DWORD%, 0x00000001
-HKR, "Parameters", DmaRemappingCompatible,0x00010001,0
+HKR, "Parameters", DmaRemappingCompatible,0x00010001,3
```
kevmw, line 94: Why did you pick mode 3 here? Documentation on it is scarce, but it seems to be meant for GPUs and will probably just fall back to 2 for Storport drivers (if not, please document the exact meaning and why we want it - in fact, whatever value we pick, it would be good to document why we want it rather than the other options).

If I understand correctly, 1 would mean that the IOMMU is always used if it's available whereas 2 would enable it only under driver verifier (theoretically it would also apply to external devices, but virtio-blk should never qualify for that).

YanVugenfirer: Note from the documentation: "Support for value 3 was added in Windows 11. The feature is designed to degrade gracefully on Windows 10 where that support isn't present, showing a value of 2 instead of 3 on the Windows 10 device."
https://learn.microsoft.com/en-us/windows-hardware/drivers/pci/enabling-dma-remapping-for-device-drivers

Make sense, as we had some issues on Windows 10 with IOMMU remapping support opting in on older Windows 10 version. But I agree that explanation should be part of the commit message

kevmw: Yes, unfortunately the docs don't really specify what the option means there, apart from being some kind of opt-in (which 1 and 2 are, too). The only thing that comes close to an explanation is from https://learn.microsoft.com/en-us/windows-hardware/drivers/display/iommu-dma-remapping:

> This value informs PnP that the device supports DMA remapping. Dxgkrnl and the HAL then coordinate to determine what type of mapping mode should be used (Remapping, Passthrough, and so forth).

I couldn't find any information that Storport is doing the same thing, or how that coordination would work in our case. This would be the valuable information to have either in the commit message or as a comment here.

The goal seems to be to handle cases where the GPU virtual address space is too small to cover all of the physical memory of the machine. I don't think this is a problem we even have with virtio?

viostor/virtio_stor_hw_helper.c
```diff
@@ -99,8 +103,21 @@ RhelDoFlush(PVOID DeviceExtension, PSRB_TYPE Srb, BOOLEAN resend, BOOLEAN bIsr)
...
+        CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_INSUFFICIENT_RESOURCES);
+        return TRUE;
+    }
 
     SET_VA_PA();
+    if (adaptExt->indirect && pa == 0)
+    {
+        CompleteRequestWithStatus(DeviceExtension, Srb, SRB_STATUS_INSUFFICIENT_RESOURCES);
+        return TRUE;
+    }
```
kevmw, line 120: This is dead code. If `adaptExt->indirect` is true, `SET_VA_PA()` set `pa = extPA.QuadPart + FIELD_OFFSET(SRB_EXTENSION, desc)`, which is never zero. Even if `extPA.QuadPart` could be zero here (it was already caught a few lines above), the offset would make it non-zero.

The same pattern repeats for other request types.

viostor/viostor.inx
YanVugenfirer: @kevmw 

The goal is to support features under VBS umbrella -  Kernel DMA Protection and DMA Guard: https://learn.microsoft.com/en-us/windows/security/hardware-security/kernel-dma-protection-for-thunderbolt
https://learn.microsoft.com/en-us/windows/client-management/mdm/policy-csp-dmaguard#dmaguard-deviceenumerationpolicy

For this we should opt in into being able to function with DMA remapping. Unfortunately Windows choose it as installation level setting and not run time setting for the driver.

kevmw:
So the things that the first commit seems to change are:

- Replace `MmGetPhysicalAddress()` with `StorPortGetPhysicalAddress()` (the former doesn't consider the IOMMU)
- Fix `StorPortGetPhysicalAddress()` calls to pass the SRB where needed and check the return value (this was strictly speaking already incorrect without an IOMMU)
- Don't DMA the serial number into `adaptExt` (which won't have an IOMMU mapping), but into `srbExt` (which does)
- Enable `STOR_ADAPTER_DMA_V3_PREFERRED` (what does this actually change?)

These changes make sense once you identified each, but having a list like this suggests to me that having four separate commits, and each explaining why the change is necessary, would be a better structure for the pull request.

Another thing I thought of is that it would be good to have in the final commit message is an example for QEMU options that set up a VM that actually shows the new feature and enables the IOMMU operation that this change enables. (Doesn't have to be complicated, maybe just copy the config you used yourself for testing.)

viostor/viostor.inx
kevmw: You mean the goal for DMA remapping support in general? Sure, that isn't hard to understand.

I was talking about DmaRemappingCompatible = 3 specifically and what it seems to enable. For the simple case of using the IOMMU/enabling DMA remapping whenever possible, wouldn't 1 be what you want? (And 2 if it's only meant for debugging.)

ybendito: My personal opinion:
- is that this should be 2, not 3
- using 1 is too risky. Even using 2 may create some side-effects during certification of Windows 10 as it already was with Netkvm.

ybendito: For example commit a165bb47cd854007d124f9460439e37b3521dd1d

viostor/virtio_stor_hw_helper.c
vrozenfe: The reason I added this check is that the old code verification tools was to complaining if it wasn't there. Apparently, the current build environment no longer relies on the SDV output, so I'm happy to remove it if the reviewers agree it's unnecessary.

viostor/viostor.inx
vrozenfe: In addition to what Yan already said, if needed, I can post the disassembly of the Storport RaidAreDriversDmarCompatible routine from different Windows versions to demonstrate that parameter 3 is handled differently across platforms.

Personally, I believe that using either parameter value 3 or 0 are the two best options at the moment, depending on whether we decide to follow the WHCP requirements or not.

Another option could be to use 0 for older Windows 10 versions and Windows Server 2016, and 2 for Windows 11 and Windows Server 2025/vNext. However, that would add more complexity to the virtio-win installer and RPM packaging workflow. (In fact, this is why I keep asking from time to time whether we still need to support Windows Server 2016 and earlier Windows 10 versions.)

ybendito: @vrozenfe Could you please elaborate why this (2 for Win11.2025 and 0 for others) somehow involves the installer? We do this in netkvm and AFAIK the installer shouldn't do anything

YanVugenfirer: @ybendito I also think 3 should be used. 2 was good for preliminary testing. We want DMA remapping to work in production, and 3 takes case of Windows 10 (it will be 2 in Windows 10 according to MS docs)

vrozenfe: @ybendito We'll probably need to keep the value set to 0 for Windows Server 2016, since any other setting doesn't work well there, if I recall correctly. That would mean we'd need different INF files for different Windows 10 versions.
Alternatively, we could detect the OS version at runtime and enable or disable DMA v3 accordingly.
If we decide to go with different INF files for different Windows 10/Windows Server variants, we'll need to make some adjustments to our packaging-related scripts. That's definitely doable, but it might turn out to be unnecessary once Windows Server 2016 support is dropped. Best, Vadim

vrozenfe:
@kevmw Thank you for your feedback, Kevin.

IMO, the changes are quite small, all related to the same issue, and should be easy to read and understand for most experienced Windows device driver developers.

I do agree that adding a placeholder for the serial number to the SRB extension may look a bit awkward, but I was simply reusing the existing ugly approach of adding a discard structure placeholder to the SRB extension, introduced earlier.

Both of these issues - removing discard and serial number placeholders,  will be addressed very soon when we increase the maximum number of supported CPUs/MSI-X vectors from 256 to 2048.

All the best,

Vadim

ybendito:
IMHO we do not need to enable IOMMU for Win10 at all. DMA v3 is not available before 1803, so in the INF we can do the same that we do with NetKVM. But this is completely IMHO.

kevmw:
@vrozenfe I agree that the individual changes are very simple, and you don't even have to be an experienced Windows developer to understand them. Still, as I pointed out above, they are four separate logical changes, and as such should be four separate commits so that the reader doesn't have to mentally disentangle them first before being able to understand if the change is complete and correct and if there are any dependencies between the changes that are mixed together. Even if it's possible or maybe even easy to understand now, anything that helps moving further along from "possible" to "easy" to "trivial" is good. And non-empty commit messages telling the motivation for the change should be a given anyway. I'm not requesting anything extraordinary here, just pointing out basic good engineering practices.

In other well-maintained projects, if I want to know why some code looks like it looks, I can use git blame and the commit that added the code will more often than not just tell me that. In virtio-win, I often instead end up with a commit that only has a terse subject line and is essentially an unexplained code dump that mixes multiple changes. Your commits stand out in the repository for this, and I think that really should change to bring the virtio-win repository to the next level of quality. If almost everyone else manages to do this, I'm sure you can do it, too. Even if you are so good a developer that you'll never need to read it again and you'll immediately remember the reasons for your changes two years down the line, other developers are part of the project, too, and they won't. Just be respectful to your future readers, it's not hard. (And certainly takes less time to just do than arguing about how it shouldn't be necessary because it's obvious anyway.)

As for having the serial number in the SRB extension, I didn't complain about the code change, it makes sense. I just said that moving it there is a logically separate step from the other three steps.

PR #1605, opened 2026-07-16
RHEL-192599: netkvm: fix iSCSI connectivity problem by MDL length adjustment

ybendito:
Jira link: https://redhat.atlassian.net/browse/RHEL-192599

Some of built-in Windows components responsible for iSCSI initiator on Server 2025 does not process properly the case when the NET_BUFFER's MDL chain is larger than the data length of NET_BUFFER. Regular TCP transfer works without problems but the iSCSI disk fails to format and probably also some other operations.

### DIFF
--- a/NetKVM/Common/ParaNdis_RX.cpp
+++ b/NetKVM/Common/ParaNdis_RX.cpp
@@ -15,9 +15,87 @@ static FORCEINLINE VOID ParaNdis_ReceiveQueueAddBuffer(PPARANDIS_RECEIVE_QUEUE p
     NdisInterlockedInsertTailList(&pQueue->BuffersList, &pBuffer->ReceiveQueueListEntry, &pQueue->Lock);
 }
 
+// Capacity of Holder MDL for PhysicalPages[pageIndex], as created by Bind.
+// First data page: VA starts at DataStartOffset, length is size - DataStartOffset.
+// Later pages: full page mapping.
+static FORCEINLINE ULONG ParaNdis_RxDataMdlCapacity(pRxNetDescriptor p, ULONG pageIndex)
+{
+    ULONG capacity = p->PhysicalPages[pageIndex].size;
+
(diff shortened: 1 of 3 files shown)


YanVugenfirer:
run tests

YanVugenfirer:
@YanVugenfirer @zjmletang 

1. Is there performance hit for all this recalculation?
2. Maybe it is better to revert for now until performance effects are understood better?

ybendito:
Win22 and Win10 - 2c_mini6rsssendrecv failed with well known error after 25th round
Probably due to expired filter.
Win19 - the setup still does not work
All the rest is ok

zjmletang:
> @YanVugenfirer @zjmletang
> 
> 1. Is there performance hit for all this recalculation?
> 2. Maybe it is better to revert for now until performance effects are understood better?

@YanVugenfirer 
On performance, I think the cost is bounded by construction:
1. NdisAdjustMdlLength is just a ByteCount write — no allocation or locking involved. The loop runs once per Holder MDL, typically 2 (one contiguous data block + the small tail). So the per-packet cost is a few field writes, negligible compared to the rest of the RX path, though I haven't benchmarked it.
2. For mergeable descriptors the function is a two-pointer-check early exit; and if mergeable eventually becomes the default, this adjustment ends up entirely on the non-default path.

So IMO there shouldn't be a real performance concern here.

zjmletang (commented):
The approach looks good to me overall.

One question: the commit message says the problem is "MDL chain larger than the data length", but that's literally true for mergeable as well (e.g. 1514 bytes in a single ~4K MDL), which the patch skips. Is the actual trigger narrower than described? If so, might be worth reflecting that in the commit message.

YanVugenfirer (approved):

ybendito:
@zjmletang the problem happens when there is an MDL which is completely out of network buffer. In case of mergeable buffers the chain contains only MDLs that are really participate in the network buffers..

PR #1609, opened 2026-07-22
viosock: tests: Add vsock_test / vsock_perf Windows ports and interop harness [SKIP-HCK-CI]

irudakov77:
This PR ports two Linux tools from `linux/tools/testing/vsock/` to Windows, groups them together with the existing `viosock-test` / `viosock-wsk-test` / `viosocklib-test` under a common `viosock/tests/` directory, and adds a Bash-scripted harness that drives them between a Linux host and a Windows guest running the viosock driver.

## Guest prerequisites

 Windows 10/11 guest with an SSH server (Administrator login). Administrator login passwordless-authorised by the SSH key you will point the harness at (installed into `%USERPROFILE%\.ssh\authorized_keys`; the matching private-key path lands in the config's `ssh_key` field, or is passed as `--ssh-key <path>` to `prepare-ci.sh` / `resolve-guest.sh`).

`prepare-guest.sh` does the `VirtIOTestCert` cert import + testsigning + firewall rule in one shot; `install-driver.sh` then does the `pnputil /add-driver /install`.

## Running the tests

The runner scripts live in the repo under `viosock/tests/run/`; the host runs them directly from a checkout - they aren't packaged or deployed anywhere. Only the signed artifact drop (`--package <dir>`) is copied to the guest, and that drop expects, in one flat directory:

* `viosock.inf`, `viosock.sys`, `viosock.cat`, `viosocklib_x64.dll`, `viosocklib_x86.dll`, `viosockwspsvc.exe` - signed driver and WSP LSP.
* `vsock_test.exe` - test binary (`vsock_test_x86.exe` next to it is optional; used with `--bits x86`).
* `VirtIOTestCert.cer` - publisher cert to import into `Root` + `TrustedPublisher` on the guest, unless `--whql` is passed (which assumes the driver is WHQL-signed and already trusted).
* `vsock_perf.exe` - the throughput benchmark. Only needed if you plan to run the perf sweep; `prepare-ci` doesn't install it -
  `prepare-perf.sh --from <pkg>` will pick it up in the same directory.

From the Linux host:

```sh
# One-shot CI-style setup (single signed artifact drop):
viosock/tests/run/prepare-ci.sh \
    --guest <libvirt-domain> --ssh-key ~/.ssh/id_xxx \
    --package /path/to/pkg

# Interop sweep (forward + reverse + loopback):
viosock/tests/run/run-all.sh

# Throughput sweep — vsock_test setup above must be done first; the perf
# host script reuses the kernel sources prepare-host already extracted
# and the driver already installed on the guest:
viosock/tests/run/prepare-perf.sh \
    --config <cfg> --from /path/to/pkg
PERF_NO_POLL=1 viosock/tests/run/run-perf.sh --logdir /tmp/perf
```
Everything else - dev-bench flow, per-direction runners, --pick, --variant, .list grammar, etc - is documented in viosock/tests/run/README.md.

Currently disabled tests

Every disabled row in viosock/tests/run/*.list carries a grep-able disabled: <reason> note. High-level breakdown:

- Not implemented in the viosock driver
  - SOCK_SEQPACKET across the board - IDs 6, 7, 8, 9, 11, 13, 15, 19, 25, 35.
  - WSAPoll on accept()-ed vsock sockets - surfaces on 3 forward ("WSAPoll: No error"), 10 forward/reverse (SO_RCVLOWAT path); also the reason vsock_perf --no-poll exists.
  - SIOCOUTQ / SIOCINQ (unread/unsent-bytes ioctls). The driver doesn't expose these - trips 24, 32 forward, 34.
- Not supported by the platform (Winsock / Windows)
  - SIGPIPE - 16, 17 (SHUT_WR / SHUT_RD).
  - MSG_ZEROCOPY / MSG_ERRQUEUE / sock_extended_err notification path - 18, 19, 20, 27, 28, 37, plus the accepted-fd SO_ZEROCOPY regression (36).
  - kmemleak - 26.
  - /proc/kallsyms + SOCK_NONBLOCK + multiple transports - 29.
  - pthread + kill(SIGUSR1) - 33.
- Driver bugs (would flip green when the underlying issue lands)
  - Acceptor credit-update path - "send failed: No error" on 22, 23.
  - Acceptor sends RST on retry connect() - 30 reverse.
  - Acceptor doesn't honor SO_LINGER timeout - 32 reverse.
  - Acceptor accepts TX beyond announced buffer - "TX credit too large: queued 2 MB (expected ≤ 64 KB)" - 38.
  - Loopback Tx-ownership-transfer hangs - 2, 3, 10, 12, 14, 32, 34, 38 loopback.

Notes

- --variant is reserved in the runners as posix|wsa|overlapped; only posix is exercised today. A future native-Winsock port can slot in without harness changes.
- Skipped tests in test_cases[] occupy their --pick IDs but the bodies are removed (run_client/run_server = NULL), so IDs stay stable against the Linux vsock_test binary - no --skip  bookkeeping needed on either side.


kostyanf14:
@irudakov77 Do you want to format /viosock/tests/vsock_test/vsock_test.c with all rules, or file copied from some other place? 

@YanVugenfirer

irudakov77:
> @irudakov77 Do you want to format /viosock/tests/vsock_test/vsock_test.c with all rules, or file copied from some other place?
> 
> @YanVugenfirer

This needs to be formatted, but the precommit hook is missing. Fixed.

kostyanf14:
ok to test

YanVugenfirer (approved):

YanVugenfirer:
@irudakov77 Some AI comments (Can be addressed later):

```
  PR #1609 Review: viosock tests — vsock_test / vsock_perf Windows ports + interop harness

  High Severity

  1. SOCKET-to-int truncation on 64-bit Windows — compat.h:187,228

  compat_socket() and compat_accept() return (int)s where s is a SOCKET (UINT_PTR, 8 bytes on x64). All subsequent compat wrappers cast back
  (SOCKET)fd, which sign-extends. If a handle has bit 31 set, the reconstructed value is wrong (0xFFFFFFFF80000000+), and every operation on it
  returns WSAENOTSOCK. In practice kernel handles are small today, but Microsoft explicitly warns against truncating SOCKET to int — this is
  borrowed from the Linux int fd convention and doesn't transfer safely.

  2. send()/recv() size_t→int truncation — compat.h:244,276

  Winsock send/recv take int for length, but the compat wrappers accept size_t. If len > INT_MAX, the cast wraps negative. This is reachable:
  vsock_perf.c:239 passes buf_size_bytes (from --buf-size, an unsigned long long) directly to read() → compat_recv(). A user passing --buf-size 
  3G triggers undefined behavior.

  3. Default LOCAL_BIN path mismatch — run-forward.sh:16, run-reverse.sh:29 vs prepare-host.sh:114

  The runners default to LOCAL_BIN="/ssd/vsock_test", but prepare-host.sh installs to /opt/vsock-test/vsock_test. Tests will fail out of the box
  unless the user overrides --bin. run-perf.sh:42 correctly defaults to /opt/vsock-test/vsock_perf, so this is inconsistent.

  Medium Severity

  4. control_writeln doesn't handle partial sends — control.c:102-112

  On a TCP stream, send() can return fewer bytes than requested without error. The code only retries on EINTR and treats any short write as fatal
  (exit(EXIT_FAILURE)). Under Winsock memory pressure, this produces a spurious "send: No error" abort.

  6. No EXIT trap in run-perf.sh / perf-one.sh — orphaned background processes

  These scripts launch background receiver processes but have no trap ... EXIT to clean them up on Ctrl-C. Compare with run-forward.sh:108 and
  run-reverse.sh:105 which have proper traps.

  7. No port-readiness check for local server in forward runner — run-forward.sh:128

  The forward runner does sleep 1 after spawning the local server. No readiness check (unlike run-reverse.sh / run-loopback.sh which call
  wait_guest_port). On a loaded host, the client can connect before the server is ready, causing spurious failures.

  Low Severity

  8. compat.h duplicated verbatim (671 lines) — vsock_test/compat.h = vsock_perf/compat.h

  Confirmed byte-identical. Bug fixes must be applied to both copies. Extract to a shared location (e.g., viosock/tests/common/).

  9. vsock_perf missing from cleanAll.bat — viosock/cleanAll.bat:3

  _cln_subdirs_ lists tests\vsock_test but not tests\vsock_perf. Build artifacts for vsock_perf won't be cleaned.

  10. ViosockPackage.vcxproj hardcodes x64 for vsock_test.exe — line 139

  <LibFiles Include="..\tests\vsock_test\x64\$(ConfigurationName)\vsock_test\vsock_test.exe">

  Uses literal x64 instead of $(PlatformTarget). ARM64 package builds will try to copy the x64 binary.

  11. pkill -f pattern too broad — run-forward.sh:103, run-reverse.sh:103

  pkill -f "$LOCAL_BIN" matches any process whose command line contains the path, including editors or monitors. Use kill "$srv_pid" for the
  known background process.

  12. No numeric validation on --timeout, --port — multiple runner scripts

  These values are used directly in timeout commands and --control-port= without checking they're numeric. Non-numeric input gives confusing
  errors.

  13. compat_fcntl(F_GETFL) always returns 0 — compat.h:418

  F_GETFL unconditionally returns 0, so O_NONBLOCK status is invisible. No code currently queries it, but the stub silently lies — a footgun for
  future callers.

  14. MSG_DONTWAIT emulation is not thread-safe — compat.h:238-254

  The FIONBIO toggle around send() is a TOCTOU race if another thread touches the same socket. Currently single-threaded, so safe — but the
  compat layer presents a general API.

  ---

  Not bugs (verified correct)

  - compat_munmap using MEM_DECOMMIT — deliberately creates access-violating pages for "invalid buffer" tests
  - wsa_set_errno default case — preserves raw Winsock error code as fallback
  - WSACleanup on error paths — properly called in both binaries
  - free_test_iovec offset arithmetic — correctly reverses allocation offsets
  - Include paths after directory restructuring — all properly updated to ..\..\
```

irudakov77:
These are all tests I have for now. Let's run them and check how many fail for this vsock.

PR #1612, opened 2026-07-24
RHEL-183402: [viostor] fail device init on feature negotiation failure

rachael-george:
`RhelSetGuestFeatures` was a void function, if `virtio_set_features` failed, there was no way to tell the caller. The driver would continue with `adaptExt->features` still holding the host-offered features, it could try to use features like FLUSH or DISCARD that the device never actually agreed to.

**Fix:** 
The fix changes `RhelSetGuestFeatures` to return BOOLEAN. On failure, it zeroes out `adaptExt->features` and marks the device as FAILED per the VirtIO spec. The two callers `VirtIoFindAdapter` and `VirtIoHwReinitialize` now check the return value and stop initialization instead of continuing with it.

In practice this is very unlikely with QEMU since the guest only selects features the host already offered. It could trigger with a buggy device implementation.

### DIFF
--- a/viostor/virtio_stor.c
+++ b/viostor/virtio_stor.c
@@ -100,7 +100,7 @@ RhelScsiModeSelect(PVOID DeviceExtension, PSRB_TYPE Srb);
 UCHAR
 RhelScsiGetCapacity(IN PVOID DeviceExtension, IN OUT PSRB_TYPE Srb);
 
-VOID RhelSetGuestFeatures(IN PVOID DeviceExtension);
+BOOLEAN RhelSetGuestFeatures(IN PVOID DeviceExtension);
 
 UCHAR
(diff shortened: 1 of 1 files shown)


YanVugenfirer:
The practical case of device failing feature negotiations is when driver not ACKing VIRTIO_F_ACCESS_PLATFORM:
https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html
 
"A device SHOULD offer VIRTIO_F_ACCESS_PLATFORM if its access to memory is through bus addresses distinct from and translated by the platform to physical addresses used by the driver, and/or if it can only access certain memory addresses with said access specified and/or granted by the platform. A device MAY fail to operate further if VIRTIO_F_ACCESS_PLATFORM is not accepted."

YanVugenfirer:
ok to test

YanVugenfirer (approved):

PR #1613, opened 2026-07-25
Stdvga v2

zjmletang:
## Summary

This is the V2 update of the stdvga display-only driver.

The initial version had two ways to request a resolution change: the standard
Windows display stack path, such as `Set-DisplayResolution` or equivalent
user-mode display API calls, and a private registry-driven path in the driver
that simulated monitor reconnects to trigger VidPn renegotiation.

V2 removes the private registry/hot-plug path and keeps resolution changes on
the standard Windows display stack. It also adds `stdvgares.exe` as a small
helper to make that standard path easier to use from unattended environments,
including callers running from Session 0, while still requiring an active
interactive user session as required by Windows.

This version also cleans up the driver implementation by removing obsolete
registry worker code, unregistered WDDM DDI stubs, stale comments, dead debug
state, and noisy wrapper-level tracing. Driver logging is moved to WPP/ETW with
function-prefixed messages so TraceView output remains useful without logging
hot paths at the default level.

## Key Changes

- Add `stdvgares.exe`, a command-line helper for changing stdvga resolution through the standard Windows display stack.
- Remove the private registry-based resolution control path and its hot-plug worker.
- Use `800x600` as the conservative default mode; larger modes remain available through Windows display settings, `Set-DisplayResolution`, or `stdvgares.exe`.
- Move driver logging to WPP/ETW, keep pure DDI forwarding wrappers quiet, and include function names in trace messages.
- Remove unregistered full WDDM DDI stubs and other stale implementation details.


kostyanf14:
ok to test

stdvga/device.cpp
```diff
@@ -153,77 +153,6 @@ static NTSTATUS SetCurrentMode(_Inout_ PSTDVGA_DEVICE_CONTEXT DevCtx, _In_ UINT
 // ---- DDI Implementations ----
 //
 
-// System thread to trigger hot-plug after StartDevice completes.
```
kostyanf14, line 156: this patch up to you, but technically, custom resolution in viogpu get/set registry
no problem to drop if you want

zjmletang: Thanks for pointing at viogpu - I went and read that path, so please correct me if I got
it wrong.

If I read `BuildModeList()` correctly, viogpu refills its custom mode slot from the host
scanout on every enumeration, so the registry is what keeps a user-chosen arbitrary
resolution from being lost across reboots. stdvga has no custom resolution support today
and only advertises a fixed mode list that is the same on every boot, so I'm not sure
there is anything for it to persist yet.

One thing I also ran into: the registry path I had conflicted with the Virtual Display
Driver in my testing. That may well be solvable, I just wasn't sure the added complexity
was worth it at this stage. If custom resolution support lands in stdvga later, it would
probably make sense to revisit this together with the viogpu approach.

kostyanf14 (approved):

Issue #977, opened 2023-08-29
About HWCursor and BZ#1919947


gang929:
The VirtioGPU driver disabled the HWCursor by default.  It's maybe caused by  BZ#1919947.

After install VirtioGPU driver,  and enable HWCursor by setting reg value [HWCursor](https://github.com/virtio-win/kvm-guest-drivers-windows/blob/eff7b43073b95c563d2e5c6b0d7b6dd954f1232a/viogpu/viogpudo/viogpudo.cpp#L1895),  it would show incorrect cursor.

The root reason maybe is that m_CtrlQueue and m_CursorQueue is not a same queue.  When [update cursor](https://github.com/virtio-win/kvm-guest-drivers-windows/blob/eff7b43073b95c563d2e5c6b0d7b6dd954f1232a/viogpu/viogpudo/viogpudo.cpp#L2614) and [QueueCursor](https://github.com/virtio-win/kvm-guest-drivers-windows/blob/eff7b43073b95c563d2e5c6b0d7b6dd954f1232a/viogpu/viogpudo/viogpudo.cpp#L2628),  the cursor pic data is processed after cursor update cmd by virtiogpu device sometimes,  so,  the cursor is still show old cursor pic data.

By adding some delay between `UpdateCursor` and `QueueCursor`, the cursor will show correctlly.
```
@@   @@NTSTATUS VioGpuAdapter::SetPointerShape(_In_ CONST DXGKARG_SETPOINTERSHAPE* pSetPointerShape, _In_ CONST CURRENT_MODE* pModeCur)
    if (UpdateCursor(pSetPointerShape, pModeCur))
    {
        PGPU_UPDATE_CURSOR crsr;
        PGPU_VBUFFER vbuf;
        UINT ret = 0;
        crsr = (PGPU_UPDATE_CURSOR)m_CursorQueue.AllocCursor(&vbuf);
        RtlZeroMemory(crsr, sizeof(*crsr));

+     m_VioDev.system->vdev_sleep(&m_VioDev, 1);

        crsr->hdr.type = VIRTIO_GPU_CMD_UPDATE_CURSOR;
        crsr->resource_id = m_pCursorBuf->GetId();

```

But,  the solution is too LOW.  
Has any better solution? 
Maybe it's better to moving `SetPointerShape QueueCursor` cmd to m_CtrlQueue,  but, it is need qemu supporting.

vrozenfe:
hi @gang929 ,

[...]

PR #1615, opened 2026-07-27
RHEL-216681: viogpu: serialize the pointer DDIs and honor pointer visibility

ameleo:
Two small patches around the hardware cursor, following the discussion in #977
(RHELMISC-34288). This is my first contribution here, so please tell me if the
split, the style, or anything about the process should be different -- I am
happy to redo them.

`HWCursor` stays opt-in and its default is unchanged, so nothing moves for users
who do not set it. The two patches are independent and can be taken separately.

### Context

With the hardware cursor disabled, Windows composes the pointer into the primary
surface itself, so a backend that captures or streams that surface gets the
pointer baked into every frame and cannot separate it from the content. That is
what led me to look at the two issues below.

### 1. Serialize the pointer DDIs

As noted in #977, SetPointerShape and SetPointerPosition are not synchronised
and are called from different threads. Updating the shape is not atomic -- the
new image is copied into the shared cursor buffer and queued for transfer, then
UPDATE_CURSOR is queued on the cursor queue -- so a second SetPointerShape can
overwrite that buffer while the first is still queueing, and a MOVE_CURSOR
issued concurrently by SetPointerPosition can land in the middle of the pair.
The cursor commands then reach the device in an order Windows never asked for.

This patch makes the two DDIs mutually exclusive with a guarded mutex. Both run
at PASSIVE_LEVEL, so they are allowed to block, and the mutex is only held
across the image copy and the queueing of the commands (`TransferToHost2D` does
not wait for the device), so it does not add a blocking point on the pointer
path.

You mentioned a queue and a kernel thread for this. I started with the mutex
because it was the smallest change I could find that removes the interleaving,
but I have no attachment to it: if you prefer the queue and thread design, or if
there is a reason the DDIs should stay non-blocking that I am missing, I will
gladly rework it.

Both functions now have a single exit so the mutex is always released; nothing
else changes here.

### 2. Hide the sprite instead of parking it at (0, 0)

This one is independent of the above.

`SetPointerPosition` treats `Flags.Visible = 0` as "move the sprite to the top
left corner": it keeps sending MOVE_CURSOR with the resource still attached and
only forces the coordinates to (0, 0), so the sprite stays visible, parked over
the top left corner of the display.

Windows clears `Flags.Visible` when it draws the pointer itself instead of
letting the adapter do it, typically when DWM composites it into the primary
during a window move or resize loop. The guest then paints its own pointer while
the parked sprite is still shown, which appears as a second, stationary cursor.
It is most visible for a remote viewer, where the sprite is drawn by the host
while the composed pointer arrives in the video stream.

The patch sends a real hide (UPDATE_CURSOR with `resource_id = 0`) and tracks
the hidden state so repeated hides are not resent. Coming back from a hide needs
a full UPDATE_CURSOR, since MOVE_CURSOR does not re-associate a detached
resource; the hot spot of the current shape is kept for that. The image itself
stays uploaded, so no new transfer is involved.

### Not addressed here

The cursor image transfer goes on the control queue, while UPDATE_CURSOR goes on
the cursor queue. Because those are independent, the device can apply
UPDATE_CURSOR before the transfer has landed and briefly show the previous
image. That is separate from the serialization above and is not fixed in this
PR: it means waiting for the transfer to complete before queueing UPDATE_CURSOR,
which touches the synchronous-wait path in `viogpu_queue.cpp` and seems better
as its own patch. I would rather send it separately, unless you prefer it here.

Sorry for not calling that out from the start: the commit message of the first
patch mentions a cursor keeping its previous image, which really belongs to this
queue asynchrony rather than to the missing serialization. I will tighten that
wording the next time I update the branch, together with anything that comes out
of review.

### Testing

Validated on a Windows 11 guest running virtio-gpu, in a downstream build
carrying these same two changes, with `HWCursor` enabled: dragging a window no
longer leaves a second stationary cursor, and cursor tracking, shape changes
(arrow, I-beam, resize arrows) and mode changes behave as before. DDI traces
showed the expected sequence -- one `Visible = 0` when the move loop starts,
silence while DWM composes, `Visible = 1` on release.

The branch as posted builds clean, but I could not deploy it to a guest as-is,
since it does not carry the rest of that downstream tree.

* Win11 EWDK (x64, `Win11 Release`): 0 errors, 0 warnings.
* `clang-format` clean on the changed lines, per `Tools/clang-format-helper.sh`.

Thanks for looking, and for the notes in #977 -- they are what made it clear
where to start.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cursor visibility handling so hiding the pointer no longer moves it to the top-left corner.
  * Restored hidden cursors at their correct position and hotspot when shown again.
  * Improved reliability when changing cursor shape and position simultaneously.
  * Cursor updates now report their operation status more consistently.

### DIFF
--- a/viogpu/viogpudo/viogpudo.cpp
+++ b/viogpu/viogpudo/viogpudo.cpp
@@ -2315,6 +2315,7 @@ VioGpuAdapter::VioGpuAdapter(_In_ VioGpuDod *pVioGpuDod)
     m_Id = g_InstanceId++;
     m_pFrameBuf = NULL;
     m_pCursorBuf = NULL;
+    KeInitializeGuardedMutex(&m_CursorMutex);
     m_PendingWorks = 0;
     m_bStopWorkThread = FALSE;
     m_pWorkThread = NULL;
@@ -2973,6 +2974,11 @@ NTSTATUS VioGpuAdapter::SetPointerShape(_In_ CONST DXGKARG_SETPOINTERSHAPE *pSet
         return STATUS_UNSUCCESSFUL;
     }
 
+    // The image upload and the UPDATE_CURSOR below must not interleave with a concurrent MOVE_CURSOR
+    // issued by SetPointerPosition on another thread (see m_CursorMutex).
+    NTSTATUS status = STATUS_UNSUCCESSFUL;
+    KeAcquireGuardedMutex(&m_CursorMutex);
+
     if (UpdateCursor(pSetPointerShape, pModeCur))
     {
         PGPU_UPDATE_CURSOR crsr;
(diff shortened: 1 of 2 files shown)


kostyanf14:
ok to test

vrozenfe:
@ameleo Nice job, Mickael. Thank you very much!

@kostyanf14 Could we ask someone from our team to test it manually, both locally and over VNC, to confirm that the problem has been fixed?

Best,
Vadim

YanVugenfirer:
@coderabbitai review

YanVugenfirer:
> @ameleo Nice job, Mickael. Thank you very much!
> 
> @kostyanf14 Could we ask someone from our team to test it manually, both locally and over VNC, to confirm that the problem has been fixed?
> 
> Best, Vadim

@vrozenfe Tested by https://redhat.atlassian.net/browse/RHEL-220527

YanVugenfirer (approved):

PR #1616, opened 2026-07-27
RHEL-216649: viofs: add optional polling-based file change notifications

alexfs:
This PR adds partial change-notification support without requiring changes to virtiofsd or the virtio-fs protocol.
When enabled, viofs runs a shared periodic worker that queries attributes for open file and directory handles. If their timestamps change, viofs invalidates cached directory data where applicable and notifies WinFsp.
This allows:

- Directories currently displayed in Explorer to refresh when directory attributes change.
- Applications with open files to detect host-side content changes.

All watched handles to be polled by a single worker thread.
The feature is disabled by default. It can be enabled with:
-n <refresh_seconds> (Default is 3 sec when zero)

Known limitations
- only open handles are monitored. Changes to files that are not open will not be reported individually, so their updated metadata may not appear in Explorer until the containing directory is refreshed by the user or otherwise re-enumerated.

### DIFF
--- a/viofs/svc/virtiofs.cpp
+++ b/viofs/svc/virtiofs.cpp
@@ -50,6 +50,7 @@
 
 #include <map>
 #include <string>
+#include <utility>
 
 #include "virtiofs.h"
 #include "fusereq.h"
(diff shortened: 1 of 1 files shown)


YanVugenfirer:
@alexfs Please also update wiki with the new parameter: https://github.com/virtio-win/virtio-win.github.io/blob/main/Knowledge-Base/Virtiofs-qs.md

YanVugenfirer (approved):

PR #1618, opened 2026-08-01
netkvm: fix TX RSS queue mapping for unmatched CPUs

hypervvv:
## Problem

`ParaNdis_SetupRSSQueueMap()` used a temporary CPU-to-bundle table that
was initialized with zero. Zero is also the valid index of bundle 0.
Consequently, an RSS CPU without a matching path-bundle affinity was
treated as if it belonged to bundle 0.

This can concentrate TX traffic on the first virtqueue when the number
of RSS CPUs is greater than the number of path bundles.

## Change

The patch reuses the existing `ResolveQueue()` helper when constructing
`RSS2QueueMap`.

`ResolveQueue()`:

- Preserves exact CPU-affinity matches.
- Assigns unmatched RSS entries to available bundles in round-robin
  order.
- Uses the same mapping logic as the device RSS configuration path.

The map is built from `ActiveRSSScalingSettings`, which is also used by
the send path.

## Mapping Result

Test configuration:

- 16 RSS CPUs
- 4 path bundles
- 128 RSS indirection entries

Before the patch:

| Queue | RSS entries |
|------:|------------:|
| 0 | 104 |
| 1 | 8 |
| 2 | 8 |
| 3 | 8 |

After the patch:

| Queue | RSS entries |
|------:|------------:|
| 0 | 32 |
| 1 | 32 |
| 2 | 32 |
| 3 | 32 |

## Test Environment

- Base commit: `abb4382405c9b195ace627fab91bff5ce19ba007`
- Guest OS: Windows Server 2022
- QEMU: 6.2.0
- Build: `Win10 Release|x64`
- 16 vCPUs
- 4 hardware queues/path bundles
- iperf3 server running on the host
- The patched driver was based on the same commit and contained only
  this RSS patch.

## Test Commands

Client:

```text
iperf3.exe -c 192.168.122.1 -P 8 -p 5201 -t 10
```

Server:

```text
iperf3 -s
```

Each driver version was tested in the guest after rebooting, followed
by eight consecutive runs.

## Throughput Results

The values below use the iperf3 `sender` summary from each run.

| Run | Original | Patched |
|----:|---------:|--------:|
| 1 | 37.9 Gbit/s | 30.8 Gbit/s |
| 2 | 18.4 Gbit/s | 30.8 Gbit/s |
| 3 | 34.5 Gbit/s | 30.4 Gbit/s |
| 4 | 21.9 Gbit/s | 30.7 Gbit/s |
| 5 | 32.6 Gbit/s | 31.5 Gbit/s |
| 6 | 35.4 Gbit/s | 29.6 Gbit/s |
| 7 | 35.5 Gbit/s | 31.4 Gbit/s |
| 8 | 30.1 Gbit/s | 30.0 Gbit/s |

| Driver | Average | Minimum | Maximum | Range |
|--------|--------:|--------:|--------:|------:|
| Original | 30.79 Gbit/s | 18.4 Gbit/s | 37.9 Gbit/s | 19.5 Gbit/s |
| Patched | 30.65 Gbit/s | 29.6 Gbit/s | 31.5 Gbit/s | 1.9 Gbit/s |

The average patched throughput is calculated as:

```text
(30.8 + 30.8 + 30.4 + 30.7 + 31.5 + 29.6 + 31.4 + 30.0) / 8
= 30.65 Gbit/s
```

`Range` means `maximum - minimum`; it is not the standard deviation.

[benchmark.zip](https://github.com/user-attachments/files/30611106/benchmark.zip)

## Conclusion

The test confirms that the patch changes the RSS queue mapping from a
bundle-0-heavy distribution to an even distribution across the four
path bundles.

In this test, the patch did not increase average throughput:

- Original average: `30.79 Gbit/s`
- Patched average: `30.65 Gbit/s`

However, the observed result range decreased from `19.5 Gbit/s` to
`1.9 Gbit/s`. Therefore, the data supports the claim that the patch
corrects the RSS queue mapping and makes the observed throughput more
consistent in this configuration. It does not support a claim of higher
average or peak throughput.

The current data does not establish behavior for other stream counts,
packet sizes, traffic directions, UDP traffic, NUMA layouts, or other
CPU and queue configurations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RSS queue mapping to better reflect active scaling settings.
  * Added fallback handling to maintain queue distribution when preferred processor assignments are unavailable.
  * Removed outdated processor validation and lookup behavior that could lead to incorrect queue assignments.


kostyanf14:
ok to test

YanVugenfirer:
@kostyanf14 will RSS test run?

kostyanf14:
> @kostyanf14 will RSS test run?

yes, but need to check results because it under filter

zjmletang (approved):

ybendito (approved):
Seems OK to me

YanVugenfirer:
@coderabbitai review

ybendito:
22H2x64: LibProtocolDriver: Install protocol driver failed.
2022x64: LibProtocolDriver: Install protocol driver failed. 
This failure should be filtered, in general

YanVugenfirer (approved):

ybendito:
@YanVugenfirer do we have a Jira issue for that?

YanVugenfirer (approved):

ybendito:
https://redhat.atlassian.net/browse/RHEL-255444 - issue for focused testing

PR #1621, opened 2026-08-08
netkvm: make the TX fragmentation limit configurable and lower the default to 63

### LINKED ISSUE #1530: netkvm：modify MAX_FRAGMENTS_IN_ONE_NB
Based on https://github.com/virtio-win/kvm-guest-drivers-windows/pull/1001/changes, netkvm has added the MAX_FRAGMENTS_IN_ONE_NB configuration to address the issue where continuous transmission of small TCP packets under certain specific hardware conditions causes page overflow, which in turn leads to network card packet loss.

As can be seen in the Conversation, there was a registry setting named MaxFragmentsInOneNB in the original design, which was presumably intended to allow modifying the value of MAX_FRAGMENTS_IN_ONE_NB via the registry.

However, in mm286, the final code implementation defines MAX_FRAGMENTS_IN_ONE_NB as a fixed value of 256（NetKVM/Common/ndis56common.h
，line87）. In that case, for network cards with a maximum fragment count limit below 256, will the aforementioned packet loss issue theoretically still occur when encountering scenarios of continuous small TCP packet transmission?

Therefore, I want  to consult the following questions:

(1) What is the reason for limiting MAX_FRAGMENTS_IN_ONE_NB to 256?

(2) Currently, besides modifying the source code, are there any other methods to modify MAX_FRAGMENTS_IN_ONE_NB, such as via the registry?

(3) If the answer to question (2) is negative, is there a repair plan for this?

zjmletang:
Fixes #1530

Implements the approach @ybendito proposed in that issue: expose the limit as a
visible configuration parameter with limits 20..256, and lower the default to 63.

- `netkvm: make the TX fragmentation limit configurable` — replaces the
  compile-time `MAX_FRAGMENTS_IN_ONE_NB` with `Init.TxFragmentationLimit`,
  keeping the default at 256, so there is no functional change
- `netkvm: lower the default TX fragmentation limit to 63`

Ordered this way so the mechanism change is behaviour-preserving and the default
change is a two-line commit that can be reverted on its own.

The mechanism change is a faithful 1:1 replacement: the macro bounded
`m_SGTableCapacity`, i.e. the total number of SG entries including the 1..4
header entries, which is exactly what the parameter now bounds. At the default of
256 the clamp is in fact inert on the indirect path, since
`virtio_get_indirect_page_capacity()` is itself 256.

Verified on Windows Server 2022 with WPP tracing that the driver really consumes
the parameter, not just that Windows stores it. Writing values straight into the
adapter key, which bypasses the INF range check:

| raw registry value | driver verdict | value applied |
|---|---|---|
| 999 | out of range | 0x3f (63, default kept) |
| 5   | out of range | 0x3f (63, default kept) |
| 20  | accepted     | 0x14 |
| 256 | accepted     | 0x100 |
| 63  | accepted     | 0x3f |

So the driver-side range check holds independently of the INF/UI validation and
falls back to the default rather than to a garbage value. Small-packet TCP load
at the default 63 (40k x 200-byte sends plus HTTPS traffic) shows
`OutboundPacketErrors = 0` with the link staying up.

Note on removing the macro: I added `MAX_FRAGMENTS_IN_ONE_NB` myself and I think
it now conflicts with the parameter rather than complementing it — the INF
materialises the parameter's default into the adapter key at install, so a vendor
build overriding the macro silently has no effect, leaving two sources of truth
for the same default. Happy to restore it and drive the INF default from a
substitution token instead (the way `SeparateTail` uses
`INX_NETKVM_SEPARATE_TAIL`) if you'd rather keep the compile-time override
authoritative.

## Summary by CodeRabbit

* **New Features**
  * Added the **TX Fragmentation Limit** network adapter setting.
  * The setting defaults to 63 and supports values from 20 to 256.
  * Configuring this value adjusts how transmit data is coalesced and can affect effective queue capacity.

### DIFF
--- a/NetKVM/Common/ParaNdis_Common.cpp
+++ b/NetKVM/Common/ParaNdis_Common.cpp
@@ -155,7 +155,7 @@ static const tConfigurationEntries defaultConfiguration =
     { "MinRxBufferPercent", PARANDIS_MIN_RX_BUFFER_PERCENT_DEFAULT, 0, 100},
     { "*NdisPoll", 0, 0, 1},
     { "MergeableBuffers", 0, 0, 1},
-    { "TxFragmentationLimit", 256, 20, 256},
+    { "TxFragmentationLimit", 63, 20, 256},
 };
 
 static void ParaNdis_ResetVirtIONetDevice(PARANDIS_ADAPTER *pContext)
--- a/NetKVM/Common/ParaNdis_Common.cpp
+++ b/NetKVM/Common/ParaNdis_Common.cpp
@@ -111,6 +111,7 @@ typedef struct _tagConfigurationEntries
     tConfigurationEntry MinRxBufferPercent;
     tConfigurationEntry PollMode;
     tConfigurationEntry MergeableBuffers;
+    tConfigurationEntry TxFragmentationLimit;
 } tConfigurationEntries;
 
 // clang-format off
@@ -154,6 +155,7 @@ static const tConfigurationEntries defaultConfiguration =
     { "MinRxBufferPercent", PARANDIS_MIN_RX_BUFFER_PERCENT_DEFAULT, 0, 100},
     { "*NdisPoll", 0, 0, 1},
     { "MergeableBuffers", 0, 0, 1},
+    { "TxFragmentationLimit", 256, 20, 256},
 };
 
 static void ParaNdis_ResetVirtIONetDevice(PARANDIS_ADAPTER *pContext)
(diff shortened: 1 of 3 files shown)


YanVugenfirer:
ok to test

YanVugenfirer:
@coderabbitai review

zjmletang:
The CI seems to be having issues.

ybendito (approved):

YanVugenfirer (approved):

YanVugenfirer:
ok to test

YanVugenfirer:
re-run tests

zjmletang:
The CI jobs for Win2022x64 and Win2019x64 stopped unexpectedly.

PR #1647, opened 2026-09-16
virtio: expose the negotiated virtqueue size to drivers

zjmletang:
The negotiated virtqueue size is only known inside the library: the transport layer reads it from the device (and possibly scales it down on ring allocation failure), but it never reaches the drivers, which only receive an opaque queue pointer and have no way to learn the actual size of their queues.

Add a vring_size field to the public struct virtqueue, filled in by the split and packed ring constructors, and a virtqueue_get_vring_size() inline accessor, the counterpart of the Linux virtqueue_get_vring_size().

The upcoming balloon free page reporting support will use this to size its reporting requests; other drivers may need it as well to batch requests up to the negotiated size.

## Summary by CodeRabbit

* **New Features**
  * Added access to the negotiated VirtIO queue size.
  * VirtIO queues now retain their descriptor count for both split and packed ring configurations.
  * The reported queue size remains consistent after queue initialization, regardless of ring format.

* **Style**
  * Reformatted diagnostic logging for improved readability without changing its behavior.


kostyanf14:
ok to test

zjmletang:
BTW, I also fixed a clang-format violation introduced by master commit 233a5895 (vioserial): the TraceEvents call was manually wrapped into three lines, although it fits on a single 105-column line (within the 120-column limit). This makes the clang-format check fail for every PR based on the current master. I included the fix in this PR only because it blocks this PR's style check and is a pure formatting change — if you prefer it not to be part of this PR, just let me know and I'll drop the commit and rebase once it is fixed on master.

kostyanf14:
rerun tests

kostyanf14:
> BTW, I also fixed a clang-format violation introduced by master commit [233a589](https://github.com/virtio-win/kvm-guest-drivers-windows/commit/233a5895364a90278a69710f3ce0f2a21b1caa4b) (vioserial): the TraceEvents call was manually wrapped into three lines, although it fits on a single 105-column line (within the 120-column limit). This makes the clang-format check fail for every PR based on the current master. I included the fix in this PR only because it blocks this PR's style check and is a pure formatting change — if you prefer it not to be part of this PR, just let me know and I'll drop the commit and rebase once it is fixed on master.

Yes, our mistake. Please ignore

kostyanf14:
rerun tests

zjmletang:
Just realized virtio_get_queue_size already exists (missed it earlier), so this PR is redundant. Closing

PR #1649, opened 2026-09-20
viosock: read the vq size from the set up virtqueue

zjmletang:
## Problem

viosock fails to start on legacy vsock devices (`PCI\VEN_1AF4&DEV_1012`): the device shows Code 10, and the WPP trace shows `virtio_query_queue_allocation(VIOSOCK_VQ_RX)` returning `STATUS_NOT_FOUND` in `VIOSockRxVqInit`. The same symptom was previously reported in #887.

## Root cause

`VIOSockQueuesInit` first sets up and activates all the virtqueues via `VirtIOWdfInitQueues`, and only afterwards do `VIOSockRxVqInit`/`VIOSockTxVqInit` call `virtio_query_queue_allocation` to learn the queue depth. On a legacy device, activating a queue programs `QUEUE_PFN` (the register doubles as both the ring address and the enable switch), and `vio_legacy_query_vq_alloc` deliberately returns `STATUS_NOT_FOUND` for queues with a non-zero PFN — this guard against double activation must stay. Any query issued after setup therefore fails. The modern path has no such refusal, which is why the problem never showed up on modern devices.

## Fix

The queue depth is already known once the queues have been set up — it is cached in `vdev->info[index].num` — so read it with `virtio_get_queue_size()`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved initialization of network socket transmit and receive queues.
  - Queue sizing is now handled more consistently, supporting more reliable packet processing during socket setup.


kostyanf14:
ok to test

YanVugenfirer:
ok to test

ybendito (approved):
looks or for me

ybendito:
@irudakov77 Could you ack?

YanVugenfirer (approved):

irudakov77:
Fine by me

